PR #2, opened 2019-12-17
Map Update

Razgriz1032:
Uses proper Southern Cross maps. Deletes corporate reg books as Vorestation has those commented out.


Razgriz1032:
Forget it. Screw you, Travis, it's FINE.

PR #12, opened 2019-12-24
Fixes

TheFurryFeline:
Changelog Notes:

- Removes wikilinks for species.

- Comments out loneliness trait for Teshari.


Razgriz1032:
I would prefer to keep SEPARATE THINGS in SEPARATE PR's. Because I WOULD merge this one, however I can't because you're missing a metric fuckton of other stuff that comes with the event changes and it would leave events unfinished.

Razgriz1032:
Wouldn't it be better to comment out the huge swatch of teshari loneliness instead of cutting it out? Maybe we could bring it back in in a future update.

PR #69, opened 2020-01-02
Fat Taur Fix

MrMacint0sh:
This update should fix the fat taurs being invisible. Also adding in the YW fat taurs, too

Happy New Years!


Razgriz1032:
This PR is currently unacceptable. Please note our new policy that any **additions** that we can easily make to .dm files must instead go into new _ch files. Also note that we already have ported taur fat body sprites to taurs_ch.dmi. Make changes to your code to reflect this and remove your edits to taurs_vr.dmi.

Darlantanis:
As we are downstream of YW, the YW taurs you "ported" are already implemented in our code; readding them is thus redundant. It is recommended you fix this bug upstream on YW's codebase.

If for whatever reason this is not feasible, you may use a _ch to set the taurs' icons to use the proper .dmi file. It's pointless to add the same code and icons twice.

Razgriz1032:
Fixed in #71

PR #89, opened 2020-01-05
Squish a Runtime Bug

TheFurryFeline:
Title.

Changelog Notes:

- Removes a file call from the .dme that was causing runtime spam for weather on a planet that this server doesn't use.


Razgriz1032:
Good find!

PR #99, opened 2020-01-07
Blep

TheFurryFeline:
Attempt to eliminate vacuum area check ZAS issue that pops up.

Changelog Notes:

- Update travis.yml to current BYOND 512.

- Reverts a YW change for a ZAS check. Doesn't give a "vacuum out of bounds" anymore, but currently gives a "no zones checked" when travis is run.


Razgriz1032:
Actually travis.yml was already updated. I don't believe merging this would harm it.

Razgriz1032:
Is this good to merge or is this WIP?

TheFurryFeline:
Only change, then, is the reversal of a ZAS check that YW did for mine/explored turfs that were vacuum-related. Otherwise, not much else worth noting other than the slight changes in the map file that can be freely reverted since it was purely experimental while I was trying to figure out how to work with ZAS unit checks.

Razgriz1032:
Lets undo those map changes and leave it alone

TheFurryFeline:
Ohhhhkayyyyyy. It doesn't look any different other than how the assigned letters or whatever that stuff's called still has the turfs' items and the like jiggled around. I did revert the changes to the 2 POIs in the centre as well as the southwest. As for the northeast, icon_states had a space in each for whatever reason and editing the map file fixed it automatically.

Razgriz1032:
Bah, nvm, just gonna expedite this process and slot your work in with #107 
We just need the one file changed.

PR #111, opened 2020-01-10
Drinks with Nutriment Factors

Somememeguy:
Gives nutriment values to sodas, (most) drinks that have soda in their recipe, as well as gives high nutriment values to drinks that use honey, peanut butter, or corn oil. Also makes it easier for people to add nutrient values to any new drink they wish to add.

+Adds in nutriment calculation for drinks (base of 0)
+Most sodas have gained a nutriment factor of around 1.5-4
+Most drinks that contain honey, corn oil, or peanut oil have a nutriment factor now as well (ranging between 4-20)


Razgriz1032:
Have you physically tested this in a test environment?

Somememeguy:
> Have you physically tested this in a test environment?

Yes I did! The sodas work, and so does the alcohol. I didn't have to add stuff for the alcohol portion because its already in the code for ethanol (which is what all alcoholic drinks are listed under).

Razgriz1032:
Ok, sounds good.

PR #116, opened 2020-01-10
advanced microwave, orca head fix, new fat naga tail 

Shadowfire117:
New stuff:
 Advanced microwave, doesnt break from some materials, has much larger capacity, fancy bluetech stuff
New chunky naga tail!
Changes:
 made item_levels for microwaves so they can handle different items
Fixed Orca head using succubus wing sprite due to wrong path to icon


Ketrai:
Don't merge this just yet, adding my half.

Razgriz1032:
This PR is currently unacceptable. It does not follow our standards as pinned in the staff development channel (which I should also probably pin in the regular dev channel.)

Anyway.
1. The addition of the object `/obj/item/weapon/circuitboard/microwave/advanced` must be removed in frame.dm and put in a new file titled frame_ch
2. Any and all new tweaks to code in base files, such as the ones found in microwave.dm, must come with `CHOMP` on the comment line as this allows us to determine what exactly WE have edited and makes it easeir to find.
3. seed_datums also need `//CHOMP` comments
4. The new circuit in circuits.dm must also be moved to a new circuits_ch.dm.
5. The chomp fat naga must move from sprite_accessories_taur_vr.dm to a new sprite_accessories_taur_ch.dm file.

PR #117, opened 2020-01-10
Chomp station2 patch 1

Shadowfire117:


Ketrai:
Will add my chunk of content to this pr immediately once all files are in place.

PR #137, opened 2020-01-13
Readding Traits, plus 1 new

Vulpias:
Moved the new Traits into a new Chompspecific file. Hopefully its now correct. 

incase of questions ask away. and if I need to


code/modules/mob/living/carbon/human/species/station/station_special_abilities_ch.dm
```diff
@@ -0,0 +1,160 @@
+//succuby bite is back baby
```
TheFurryFeline: Enable this file in the .dme. Add a define for it so they actually get called in neutral traits

Vulpias: updated

Razgriz1032:
Looks good to me

PR #147, opened 2020-01-14
Letting micros ride Paper airplanes, again

Sharkmare:
PUT MICROS ON PAPERLANES

### DIFF
--- a/code/modules/paperwork/paperplane_ch.dm
+++ b/code/modules/paperwork/paperplane_ch.dm
@@ -1,0 +1,132 @@
+// Ported from TG. Known issue: Throw hit can possibly double-proc. Seems to be throw code.
+/obj/item/weapon/paperplane
+	name = "paper plane"
+	desc = "Paper folded into the shape of a plane."
+	icon = 'icons/obj/bureaucracy.dmi'
+	icon_state = "paperplane"
+	throw_range = 7
+	throw_speed = 1
+	throwforce = 0
+	w_class = ITEMSIZE_TINY
+
+	var/obj/item/weapon/paper/internalPaper
+
+/obj/item/weapon/paperplane/New(loc, obj/item/weapon/paper/newPaper)
+	. = ..()
+	pixel_y = rand(-8, 8)
+	pixel_x = rand(-9, 9)
+	if(newPaper)
+		internalPaper = newPaper
+		flags = newPaper.flags
+		color = newPaper.color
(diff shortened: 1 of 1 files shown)


Razgriz1032:
Can you make a blank commit to kick travis?

Razgriz1032:
I just remembered that I can do that myself.
Also I will need to add this file to the dme. I will do that after travis says this is ok

Razgriz1032:
Nope. Indentation issue. Will close due to inactivity.

Razgriz1032:
This entire file can be shortened just to:

`//CHOMPEDIT This should let micros ride paper airplanes

/obj/item/weapon/paperplane/attackby(var/obj/item/I, var/mob/user)
	if(istype(I,/obj/item/weapon/holder/micro))
		var/full = 0
		for(var/mob/M in src)
			full++
		if(full >= 2)
			to_chat(user,"<span class='warning'>You can't fit anyone else into \the [src]!</span>")
		else
			var/obj/item/weapon/holder/micro/holder = I
			if(holder.held_mob && holder.held_mob in holder)
				to_chat(holder.held_mob,"<span class='warning'>[user] stuffs you into \the [src]!</span>")
				holder.held_mob.forceMove(src)
				to_chat(user,"<span class='notice'>You stuff \the [holder.held_mob] into \the [src]!</span>")
	else
		..()

/obj/item/weapon/paperplane/proc/clear_mobs(var/mob/user)
	for(var/mob/M in src)
		M.forceMove(get_turf(user))
		to_chat(M,"<span class='warning'>[user] shakes you out of \the [src]!</span>")
		to_chat(user,"<span class='notice'>You shake [M] out of \the [src]!</span>")

/obj/item/weapon/paperplane/attack_self(mob/user)
	clear_mobs(user)
	to_chat(user, "<span class='notice'>You unfold [src].</span>")
	var/atom/movable/internal_paper_tmp = internalPaper
	internal_paper_tmp.forceMove(loc)
	internalPaper = null
	qdel(src)
	user.put_in_hands(internal_paper_tmp)
//END CHOMPEDIT`

Because everything else that's in the file is already here: 
https://github.com/CHOMPStation2/CHOMPStation2/blob/master/code/modules/paperwork/paperplane.dm

Razgriz1032:
In fact it needs to be cut down.
image

Sharkmare:
It needs to fully replace the old paperplane file because it modifies the core paperplane files as well so it cant just be added onto.

PR #150, opened 2020-01-14
Revert Changes

TheFurryFeline:
Title.

Changelog Notes:

- Revert changes to Rouge Drone event output regarding what's where.


TheFurryFeline:
As a notice, YW is presently making their own changes to the rouge drone event so the next update we take from them likely will conflict with this PR.

PR #152, opened 2020-01-14
Improved Request Spice

Kazzc:
Committed out the old "request Spice" from Virgo and made a new command for "request spice"

It should send a ding noise, comes with commands that allows an admin to "jump, give cookie, Look at VV, look at Player panel, DM them"

### DIFF
--- a/code/modules/admin/verbs/adminhelp_vr.dm
+++ b/code/modules/admin/verbs/adminhelp_vr.dm
@@ -15,6 +15,8 @@
 		query_string += "&admin_number_afk=[afkmins.len]"
 		world.Export("[config.chat_webhook_url]?[query_string]")
 
+
+/* //CHOMPStation Addition 14/1/20 Ksc - Committing this out to be replaced with a improved request spice.
 /client/verb/adminspice()
 	set category = "Admin"
 	set name = "Request Spice"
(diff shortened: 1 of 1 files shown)


code/modules/admin/verbs/adminhelp_vr.dm
```diff
@@ -36,3 +38,37 @@
...
 		usr.verbs += /client/verb/adminspice	// 10 minute cool-down for spice request
+*/
+
+/mob/verb/Spice(msg as text) //CHOMPStation Addition 14/1/20 Ksc - The improved request spice command.
+	set category = "Admin"
+	set name = "Request Spice"
+	set desc = "Request for admins to challange your shift"
+
+	if(say_disabled)	//This is here to try to identify lag problems
+		usr << "<font color='red'>Speech is currently admin-disabled.</font>"
```
TheFurryFeline: Standard lately is `to_chat(usr,[insert rest of code])` in place of `usr << [insert rest of code]`

Kazzc: ok

code/modules/admin/verbs/adminhelp_vr.dm
```diff
@@ -15,6 +15,8 @@
 		query_string += "&admin_number_afk=[afkmins.len]"
 		world.Export("[config.chat_webhook_url]?[query_string]")
 
+
+/* //CHOMPStation Addition 14/1/20 Ksc - Committing this out to be replaced with a improved request spice.
```
TheFurryFeline: Technically, this would be a removal to be overridden by the new verb below.

code/modules/admin/verbs/adminhelp_vr.dm
```diff
@@ -36,3 +38,37 @@
 	usr.verbs -= /client/verb/adminspice
 	spawn(6000)
 		usr.verbs += /client/verb/adminspice	// 10 minute cool-down for spice request
+*/
+
+/mob/verb/Spice(msg as text) //CHOMPStation Addition 14/1/20 Ksc - The improved request spice command.
```
TheFurryFeline: Also would have this be like Chomp Addition Start/End for the whole snippet but since it's the end of the file, it'll be fine for now. :>

PR #177, opened 2020-01-19
Add an Examine to Bonfires

lbnesquik:
This just let you check out how much fuel is in there.

### DIFF
--- a/code/game/objects/structures/bonfire.dm
+++ b/code/game/objects/structures/bonfire.dm
@@ -34,13 +34,15 @@
 /obj/structure/bonfire/permanent/sifwood/New(newloc, material_name)
 	..(newloc, MAT_SIFWOOD)
 
+//CHOMPStation Addition Start
 /obj/structure/bonfire/examine(mob/user)
 	var/X = get_fuel_amount()
-	user << "The fire has [X] burning logs in it."
+	user << "The fire has [X] logs in it."
 	if(grill)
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/bonfire.dm
```diff
@@ -34,6 +34,14 @@
 /obj/structure/bonfire/permanent/sifwood/New(newloc, material_name)
 	..(newloc, MAT_SIFWOOD)
 
+/obj/structure/bonfire/examine(mob/user)
+	var/X = get_fuel_amount()
+	user << "The fire has [X] burning logs in it."
+	if(grill)
+		user << "The [src] has a crude grill plate over it."
+	if(stake)
+		user << "The [src] has a makeshift stake plate over it, perfect for witches and space templars."
```
TheFurryFeline: Undefined var for stake here. Also, please use to_chat per the new standard.

TheFurryFeline: Also, this is editing a core file. If you can, ya should make a _ch file addition or add `//CHOMPStation Addition Start/End` at either end of the new code block.

ExabyteDiscord: Seconding what @TheFurryFeline said. Definitely going to need these changes before the merge could be accepted.

lbnesquik:
@ExabyteDiscord REquested changes have been made

PR #185, opened 2020-01-21
Add a craftable, two-handed sledgehammer.

lbnesquik:
- Can be made out of plasteel, rods and metal.
- Relatively high damage but less than the fireaxe.
- Special sound.
- Inhand sprite
- Mjolnir version for admin abuse.


Ketrai:
Feels rather weak. Does it have any special functionality?

TheFurryFeline:
Hmm, maybe rename the sound file to smash1_ch.ogg.

lbnesquik:
It is intentionally way under the fireaxe because it is craft able. I could make it closer however.

Razgriz1032:
How do you manage to make the hammer head to start construction?

Also, would you consider slowing down the rate of attack? Just curious.

lbnesquik:
@Razgriz1032 You use plasteel to make the head. Expensive to make it not too common, but not a requirement. 
As for attack speed, sure, if you tell me what variable that is and if it doesn't make the weapon completely underpowered.

Razgriz1032:
I think it'll be fine for now. I'll go ahead and merge.

PR #186, opened 2020-01-21
Add a buckler, craftable with 20 wood.

lbnesquik:
This is simply re-using a sprite i've found in the code.
25 block chance instead of 50, given it is smaller and easier to get than most other shields.


code/modules/materials/material_recipes_vr.dm
```diff
@@ -12,3 +12,7 @@
 /material/durasteel/generate_recipes()
 	. = ..()
 	recipes += new/datum/stack_recipe("durasteel fishing rod", /obj/item/weapon/material/fishing_rod/modern/strong, 2)
+
+/material/wood/generate_recipes()
+	. = ..()
+	recipes += new/datum/stack_recipe("wooden buckler", /obj/item/weapon/shield/riot/buckler, 20, time = 40)
```
TheFurryFeline: This sort of code looks like it'll work in a _ch file.

Darlantanis:
After some thought, I've decided I'm against this PR. Wood is exceptionally easy to obtain on this map meaning the only sacrifice for using a buckler is an empty hand slot. Outside of wielded weapons (e.g. spears) there is no reason not to have/use a buckler. That said, it could have a place in the code as an uncraftable item perhaps for use with events or exploration POI's.

This PR will become the new meta if accepted in its current state.

lbnesquik:
It could be made into rare cargo contraband or perhaps buyable for points at certain vending machines. Maybe with a slowdown on top of it to make it less meta.

Razgriz1032:
I am against this as well. As Nadyr said, if it was uncraftable and/or set up to be a POI object or admin spawn only, sure, but since there's been no progress on this, I will close.

PR #190, opened 2020-01-22
Tweak stethoscope  lewd code.

lbnesquik:
Make it so the stethoscope doesn't pick up vore & stuffing by listening to people feet or something.


ExabyteDiscord:
This could also probably be ported upstream if it hasn't already

PR #197, opened 2020-01-25
Medical Doctor and Roboticist alt titles

Razgriz1032:
Adding Dentist to Medical Doctor alt titles.

- [ ] Needs review by medical players. Does this need a different icon? Does this need default loadout changes? Etc

Adding MechTech to Roboticist alt titles (because MechWarrior, mang. Also this was old CHOMPStation content.)

### DIFF
--- a/code/game/jobs/job/medical.dm
+++ b/code/game/jobs/job/medical.dm
@@ -43,7 +43,7 @@
 		"Emergency Physician" = /decl/hierarchy/outfit/job/medical/doctor/emergency_physician,
 		"Nurse" = /decl/hierarchy/outfit/job/medical/doctor/nurse,
 		"Virologist" = /decl/hierarchy/outfit/job/medical/doctor/virologist,
-		"Dentist" = /decl/hierarchy/outfit/job/medical/doctor/nurse)
+		"Dentist" = /decl/hierarchy/outfit/job/medical/doctor/nurse) //CHOMPStation addition: dentist
 
 //Chemist is a medical job damnit	//YEAH FUCK YOU SCIENCE	-Pete	//Guys, behave -Erro
 /datum/job/chemist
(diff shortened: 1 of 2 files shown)


D-H-A:
There is currently nothing dentist related in the game or involved with medical doctors.

Darlantanis:
What's the purpose of a dentist? There are no existing mechanics for a dentist and as a medical alt they'd have all the duties of a medical doctor (they'll count toward medical randoms, won't they?). 

If it's a roleplay-focused job, sure, but then I'd question why they're not a civ job because meta means they'll just be normal doctors if part of medical.

Razgriz1032:
There are a couple jobs with no mechanics. Chaplain has barely any. There are no mechanics for psychiatrist either. Maybe we can think of making it a new thing mirroring the psychiatrist?

Darlantanis:
Psych is the only rp-job I can think of which isn't a civ job (unless you want to count command secretary despite that one just being intern with extra access) though meta often leads to it handling low-level medical treatment and making chems. Dentist could work similarly if we want to continue that with risk of additional meta-responsibilities (ideally they shouldn't count for med events if officially they only have rp-responsibilities). This should probably come with a small office for roleplay similar to the psych one).

Alternatively it could be a civ job if we don't want the risk of meta-responsibility. Here though it will probably definitely need a new/repurposed room as a dental office.

Rykka-Stormheart:
> Psych is the only rp-job I can think of which isn't a civ job (unless you want to count command secretary despite that one just being intern with extra access) though meta often leads to it handling low-level medical treatment and making chems. Dentist could work similarly if we want to continue that with risk of additional meta-responsibilities (ideally they shouldn't count for med events if officially they only have rp-responsibilities). This should probably come with a small office for roleplay similar to the psych one).
> 
> Alternatively it could be a civ job if we don't want the risk of meta-responsibility. Here though it will probably definitely need a new/repurposed room as a dental office.

Semi-med player here, I agree with the above.
An alt-title of Psych or a Civ job/alt-title would fit far better, because it'd be an RP job. Perhaps we can poke @TheFurryFeline if they have room to make a dentist office in their medical overhaul on the map?

Darlantanis:
After thought and no further points raised, I think I'm settled on dentist being a civ role instead of a medical one.

PR #209, opened 2020-02-02
Stops shuttle 1 & 2 from spamming common every time they dock or undock

TalkingCactus:
makeitstop

### DIFF
--- a/maps/southern_cross/shuttles/crew_shuttles.dm
+++ b/maps/southern_cross/shuttles/crew_shuttles.dm
@@ -98,13 +98,13 @@
 	routes_to_make = list(
 		/datum/shuttle_destination/shuttle1/outside_SC = 0,
 	)
-/*
+/*//Chompstation Removal
 /datum/shuttle_destination/shuttle1/root/get_arrival_message()
 	return "Attention, [master.my_shuttle.visible_name] has arrived to Hangar One."
 
 /datum/shuttle_destination/shuttle1/root/get_departure_message()
 	return "Attention, [master.my_shuttle.visible_name] has departed Hangar One."
-*/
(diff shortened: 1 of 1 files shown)


Darlantanis:
Biggest problem with just commenting this out is these shuttles are the main (often only) method of travel between the station and planet. This fixes the spam but introduces a new problem that now the only way for crew to determine the location of a shuttle is to check manually. Which is hardly ideal when Hangars 1 and 2 are on almost opposite ends of the station.

An ideal fix would remove the spam while still making it easy for any crew member to remotely determine where shuttles 1 and 2 are.

TalkingCactus:
I agree that there should be another way to see where the shuttles are, but I also don't think it's so bad having to wait 5 minutes, maximum, for a shuttle. At least until we have another method of knowing where they are.

A simple communicator or PDA app could track their position easily enough, too. Should be a pretty quick project.

Zeldazackman:
There was a slight edit from for a private server I'd seen a little bit ago that had a status display and a renamed newscaster to display the arrival and departure timings for the shuttles

Darlantanis:
Oh, an additional issue now that I remember it. Pathfinders/pilots can manually fly shuttles 1/2 elsewhere (and sometimes still do despite the addition of shuttle 3). There's currently no easy method of determining if one of those shuttles is out of commission unless the messages stop.

ExabyteDiscord:
I personally think we should have at least some way of checking if the shuttles are there, but it doesn't seem like it'd be as simple a fix as I thought to replace it in a convenient way.

Razgriz1032:
I am going to close this PR since this is decidedly not the solution at this time. Zeldazackman's comment seems like a good place to restart.

PR #210, opened 2020-02-02
Sif Snow Fix

Darlantanis:
You know how walking through planet snow is slow as balls? Turns out it's not supposed to be like that, YW changes to snow turfs broke the map a bit. This PR fixes planet-side snow to perform as designed on Polaris. Changes include:
-No extreme slow down. Snow will still slow you but it's not nearly as bad as before.
-Stepping in slow leaves footprints now.
-Crunchy snow walking sound effects.
-You can dig up snow by hand (or with a shovel if you want to destroy the tile) to make snowballs and snow bricks (i.e. you can build igloos now).
-Did I mention you can build igloos now?
-Seriously go build a snow village and have a snowball fight.

### DIFF
--- a/maps/southern_cross/turfs/outdoors.dm
+++ b/maps/southern_cross/turfs/outdoors.dm
@@ -164,7 +164,7 @@
 /obj/effect/step_trigger/teleporter/planetary_fall/sif/find_planet()
 	planet = planet_sif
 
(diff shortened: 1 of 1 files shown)


Razgriz1032:
lgtm

PR #213, opened 2020-02-06
Gateway Missions

Darlantanis:
Implements gateway missions for exploration accessible from the exploration carrier. Currently two maps exist for exist for extended rounds, should be possible to admin-load other maps for events and such.
-Adds code to defines and submaps for gateway exploration.
-Adds carpfarm and snowfield as gateway maps.
-Adds gateway to the exploration carrier.
-Also adds a teleporter and changes maintenance areas around the exploration carrier.


maps/southern_cross/southern_cross_defines.dm
```diff
@@ -92,12 +93,9 @@
 			)
 
 	lateload_single_pick = list(
-		//list("Snow Outpost"),		// Unplayable mapgen,
-		//list("Zoo"),				// Too big. way, way too big
 		list("Carp Farm"),
-		list("Snow Field"),
-		list("Listening Post")
```
TheFurryFeline: Why not keep the listening post in, or the other two?

Darlantanis: The Listen Post has less content than a small POI and mostly serves as a repository of Virgo events/Virgo admin history. That doesn't have a place here.

Snow outpost is a reskin of Southern Cross wilderness with the same POI's which already spawn in the wilderness. There's not much point to using it while we're on Southern Cross.

The zoo looked interesting but the comments gave me the impression it has a negative impact on server performance. This one might be viable but I'd like to see how it impacts server performance (probably by admin-loading the z-level in) before adding it to the gateway list. zoo.dm will need cleaned up a bit if added, it makes reference to a character which I suspect is somebody's character on Virgo.

TheFurryFeline: Fair nuff on these points.

PR #216, opened 2020-02-08
Modified "control_freak" to full access.

Kazzc:
Players and Admins now have access to the full list of "client side options". So that means they can use; Command, Macros, Gamepad Mapping, Gamepad setup, Edit Skin & Preference.


TheFurryFeline:
For reference, these are the options we are restoring, macros and gamepad having been removed by YW.
Macros_Restored

PR #220, opened 2020-02-10
minor dog-borg patch

JesusChristItsJasonBourne:
re-adds V-3 security borg and V-3 medihound so that everyone can enjoy vale's sprites again - additionally re-adds medical grabber so medihounds can once more do chems


Razgriz1032:
Lgtm

PR #222, opened 2020-02-12
Pursuit Hardsuit

Razgriz1032:
Re-adds https://github.com/CHOMPstation/CHOMPstation/pull/310

There have been a couple changes since the original PR on old chomp over time. Thrusters have been added and it is restricted to HOS access only.

I am testing this on a private test server now, but I am opening this now so that I can get travis out of the way if this works first go.


Razgriz1032:
Testing is successful. Suit is functional as well as the modules.

PR #224, opened 2020-02-12
New uniform, rolled states, fixed star trek uniforms

Razgriz1032:
-Adds a new security uniform. It's just green like a flight suit (just a copy of the Tactical Jumpsuit really), but it is restricted to security use as it comes with the armor of standard security jumpsuits.

-Adds a few rolled down suit states from old chomp code to icon files. Some of these sprites might not be necessary, but they won't hurt to have. These additions won't be seen in-game.

-Unfortunately rolled sleeves states cannot be easily separated into _ch files, so they were added to the base files.

-Adds a rolled down and rolled sleeves sprite of tactical jumpsuit (which applies to the new security jumpsuit as well.)

-Ports https://github.com/CHOMPstation/CHOMPstation/pull/709

-Fixes star trek uniforms. Engineering and security are now separate uniforms so that they can have different stats.

-Nerfed armor values on a padded star trek coat as no other jackets/coats for standard use have those values.


Razgriz1032:
So it was worse than I thought. All these fedcoats got different armor values. So I have nixed them so they're just fluff items now.

code/modules/vore/fluffstuff/custom_clothes_vr.dm
```diff
@@ -262,21 +262,21 @@
 	desc = "A uniform jacket from the United Federation. Starfleet still uses this uniform and there are variations of it. Wearing this may make you feel all scientific."
 	icon_state = "fedblue"
 	item_state = "fedblue"
-	armor = list(melee = 0, bullet = 0, laser = 0,energy = 0, bomb = 0, bio = 50, rad = 0)
+	armor = list(melee = 0, bullet = 0, laser = 0, energy = 0, bomb = 0, bio = 10, rad = 0)
```
TheFurryFeline: I'd suggest adding the CHOMPEdit comments to all instances here.

PR #225, opened 2020-02-12
Stair Decals and a Boxing Ring

D-H-A:
Title self explanatory. 

- Adds a few new floor decals -these being stairs. They're strictly for visual purposes.
- Adds a boxing ring string, plus two more gym floors.


Razgriz1032:
Lgtm

PR #227, opened 2020-02-14
Tiny Armadillo

Kazzc:
Sprited and coded a simple little critter, I plan to work on it more in the future to wear them as a hat. For now is a simple mob.

*It can be pet
*Runs fast if you hurt it
*Taste like sand
*Can handle higher temperatures but will die in the cold
*Can walk over it
*Makes cute armadillo noises
*Is a chubby critter that rolls around.


JesusChristItsJasonBourne:
I'd totally be up for these replacing an older department pet

Razgriz1032:
lgtm

PR #233, opened 2020-02-16
[Submap] I Got No Wood

TheFurryFeline:
This changes the Snow Field Holodeck thingy, so... Submap change, I guess? Prevent anyone from getting wood from it as an exploit.

Changelog Notes:

- Prevents wood generation exploit for the holodeck that lets you get wood from holodeck generated dead/pine trees.

- Thus adding holodeck variation.


Razgriz1032:
This is a smart idea. lgtm

PR #235, opened 2020-02-16
Liquid vore bellies!

Shadowfire117:
After a week of non-stop work on coding this its finally here! Churn your friends into different liquids in your vore bellies, make some over time using nutrition, empty into containers, into other vore bellies or stomach, even into other people as well! You can write custom messages for every belly for people to see if you want them to, so this vore addition works well for not only vore, but also for inflation and other 'kinky' things, bottoms can take advantage of this as well to show off 'donation's from your tops!

It works for anything that has a vorepanel, so synths and borgs can also join in on the fun! (Not an expert on those folks, so sorry if there are some bugs when I made it to include them :c)

I plan on making on the second phase of this poject (wont start now, I need break and feedback x.x):

Allow people to set transfer verbs and custom names for liquids when they are transfered, sorry, there wont be actual custom fluids, thats not something that can really be done easily if at all!
Toggle for belly noises when walking around, so the more full people are the more often and clearly sloshing and such is heard
More reagents! Do give tell what you would like, of course meds and absurd chems will be a no!
Certain races/synths/borgs will be able to select additional liquids they can already produce in non-vore way!
Potential round persistency of liquids! This will only be the fullness of belly, not the content, since this can be abused to transfer powergaming reagents across shifts! So selected reagent of belly will be what the belly contains
Adjust how quick and slow the rate of liquid production can be
Possible custom messages to replace normal vore messages when at certain intervals of fullness, like prey splashing in fluids, folks seeing struggle messages with bulges appearing on liquid filled belly
Puddles! Puddles which name and color are based on fluid spilled on floor! Will also mean transfer will get a choice to dump onto floor

Finally I wanna thank Jon for a lot of code help, and Aronai to get the few final kinks worked out!


Dinnel:
I've tested it. Seems to be working fairly well but the only problem is transferring it to a container like a beaker. It does transfer but it's only 1 at a time regardless of how much I say. Another note, I ate a test dummy that I filled with milk and their contents went into my blood (the milk). This caused toxin damage over time but I think this might be from the vore code in general.

code/modules/vore/eating/living_ch.dm
```diff
@@ -0,0 +1,138 @@
...
+					user.visible_message("<span class='notice'>[user] fills [TR]'s stomach with [RTB.reagent_name] from [TG]'s [RTB].</span>")
+
+				add_attack_logs(user,TR,"Transfered reagents from [TG]'s [RTB] to [TR]'s Stomach")	//Bonus for staff so they can see if people have abused transfer and done pref breaks
+
+		if("Containers")
+			var/list/choices = list()
+			for(var/obj/item/weapon/reagent_containers/rc in view(user.loc,1))
+				choices += rc
+			var/obj/item/weapon/reagent_containers/T = input(user,"Choose what to transfer to","Select Target") as null|anything in choices
+			RTB.reagents.vore_trans_to_con(T, 1, 1, 0)
```
Dinnel: I think this might be the offending line. There is no 'transfer_amount' on it and instead there are 1s. I was only able to pour 1 unit into a beaker, maybe this might be it? I can try modifying it a bit.

Shadowfire117: yeah youre right, it appears the code on that proc went back to the earlier version where I didnt have the transfer option set up yet. will fix now

Razgriz1032:
Still pending comments in some areas

Razgriz1032:
Approved

PR #218, opened 2020-02-08
Casino port!

Shadowfire117:
1571908233 shadowfire117_sergnoodle

This is first phase of casino port where I have made my casino spaceship compatible with new chomp codebase. There isnt a connecting port at station yet due to changes being made to main map, but with second phase of casino port there should be one for casino shuttle to move back and forth to transfer visitors.

Non-casino changes
Port of:
purple assistant cap, wah!
wiggler hat, dont know if the sprite works fully yet.
The monocoole, every self respecting technomancer needs this! Also need this for casino rewards.
Snaps, can recommend for lunch, I prefer the licorice flavoured ones!
Space cleaner dispenser, yep any janitors dream, so far its only on casino spaceship.
Ported old modular sofa for station usage, is already on casino, future plans could involve making the modular sofa that can be built from steel be a little more pretty and made from wood as well and also get another overlay/color when added cloth for cushion.

also deleting a .txt file that I didnt get to delete before someone really wanted me to upload wip files so they could work off them, sorry about that but its fixed now!
deleted alienwhitelist and custom_items.txt since github doesnt seem to have those blacklisted

casino port:
Ported the map with the golden goose spacecasino
ported the deluxe edition bar and kitchen gear! Say hello to the good old deluxe dispenser, deluxe food vendor and deluxe drink vendor, perhaps one day they will be constructible by crew!
ported the casino stripper pole, casino roulette, slotmachine, poker/blackjack table, ported the casino wheel of fortune
casino wallet~
casino jukebox! And re implemented the system so the casino songs no longer are on the normal jukebox, sorry folks, hope you didnt like them too much :c
casino visitor and staff manual, might need a bit of revision depending on what event managers wants
Casino chips, gotta gamble with something!
fixed runtime errors from deluxe dispenser trying to add cartridges with reagents that have been renamed on new chomp

Be aware, there will be more casino changes in near future along with some other non-casino changes, including porting of pint metamorphic glass.

PR #237, opened 2020-02-19
Revert deletion of alien and custom item whitelists

Razgriz1032:
Reverts the deletion of the alien and custom item whitelists from casino port
https://github.com/CHOMPStation2/CHOMPStation2/pull/218

Because that just causes issues I don't care to immediately fix because updating of those files isn't even handled on github anyway. We're handling them manually on the server.
image


Razgriz1032:
Pardon me for ramming this through. This is actually holding up the server update.

PR #241, opened 2020-02-20
Gauss projectile icon fix

Razgriz1032:
Removed '_yw' from the path to use regular projectiles .dmi. It already has a gauss icon in it. The _yw file had 4 dir icon but projectiles only work with 1 static dir.

Waiting to hear back from upstream to fix it for them. If I can, this PR may be subject to change or closure.


Razgriz1032:
This is the upstream: https://github.com/Yawn-Wider/YWPolarisVore/pull/601

PR #243, opened 2020-02-22
Drink reagent port

Shadowfire117:
Ported drinks, added recipes, changed "Sit on my face" (please do) to use the peanutbutter already in codebase.
Made a system so reagents can have their metaglass sprite file be custom and the metaglass will change its own icon pathway dynamically.
Made the old commented out nutbutter cartridge be replaced with a peanutbutter cartridge

Planning on porting more reagents I can find for 'raw' drinking chems, medical chems and other chems.


Razgriz1032:
I will have to compare it with this PR tomorrow.
#8

Shadowfire117:
> 
> 
> I will have to compare it with this PR tomorrow.
> #8

No point, I made this with code taken from chompstation2 few hours ago, so its built on that metaglass code.

Razgriz1032:
lgtm

PR #251, opened 2020-03-08
Egg

TheFurryFeline:
:egg:

### DIFF
--- a/code/modules/client/preference_setup/loadout/loadout_fluffitems_ch.dm
+++ b/code/modules/client/preference_setup/loadout/loadout_fluffitems_ch.dm
@@ -7,7 +7,7 @@
 	character_name = list("Ash Williams")
 	description = "A bronze medal awarded for distinguished conduct. Whilst a great honor, this is most basic award on offer. It is often awarded by a captain to a member of their crew."
 	path = /obj/item/clothing/accessory/medal/conduct
-	display_name = "Ash Williams's Medal"
+	display_name = "Ash Williams' Medal"
 
 /datum/gear/fluff/durgitmedal
 	ckeywhitelist = list("blackmajor")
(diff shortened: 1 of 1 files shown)


Razgriz1032:
The pathing convention for the medals is bad. People with multiple medals would be a mess.
Instead of
/datum/gear/fluff/medalofsilvervalornadyr

Why not
/datum/gear/fluff/nadyr/medal1
/datum/gear/fluff/nadyr/medal2

And I'm serious about just calling it medal1 and medal2. You don't need the name of the medal in the path.

Razgriz1032:
1. You did the opposite of what I said and by leading with the name of the medal instead of leading with a character name.
2. Your file comment says alphabetized by character name when you alphabetized by medal.
3. You should instead alphabetize by ckey because the ckey represents the individual and one individual may have multiple characters. That's  how it was organized before, and that's  just basic database common sense.

TheFurryFeline:
> 1. You did the opposite of what I said and by leading with the name of the medal instead of leading with a character name.
> 2. Your file comment says alphabetized by character name when you alphabetized by medal.
> 3. You should instead alphabetize by ckey because the ckey represents the individual and one individual may have multiple characters. That's  how it was organized before, and that's  just basic database common sense.

I like this naming scheme, though. As for the alphabetisation, it **is** by character name with the medals being from bronze, silver, and to gold.

TheFurryFeline:
In fact, this way, you can have multiple medals for the same char by just adding the same name again with a number 2 and so on.

TheFurryFeline:
It'd also be easier for others to see how to add their own in the future. With the name of the medal in the path as well as the character name after it, it's a simple matter of adding their own listing if they want to make the PR themselves after getting awarded it.

Darlantanis:
Completed by #260.

Please avoid rejecting requested changes twice next time. Getting a PR up 30 minutes after it's requested isn't helpful when 5 days later it still lacks the requested revisions.

PR #257, opened 2020-03-10
Finish suit cycler stats inheritance 

Razgriz1032:
Finished off suit cycler inheritance. Now suits will forget their old stats and inherit new stats so that you can't stack stats and create a super suit, or something.


Shadowfire117:
Code seems pretty fine, cant tell theres any issue

PR #264, opened 2020-03-16
Re-enable custom_items.txt

Razgriz1032:
Reinstating commented out code that will let custom_items.txt do its job. This makes it easier for us to do casino items without having to do a PR, and also increases anonymity.


Razgriz1032:
I will look into that at a later date. It works for now tho.

PR #265, opened 2020-03-18
Remove Neko character. Also egg

Razgriz1032:
Removing Neko character that is no longer in code.


Shadowfire117:
Code looks fine

PR #266, opened 2020-03-19
Completely New Outpost

D-H-A:
New outpost built to help settle the issues of the old one. It was also built with the intentions of making planet-side more appealing to players to visit, and for future planet-side events. A bunch of issues of the old one were resolved in the process of making this, such as how the cold environment was unfriendly to newbies, or anyone coming without the gear for their species.

**Things to note:**
- The xenoarch outpost is untouched for now.
- The extra umbrellas, coats, exploration gear rooms, wildlife defense guns, and so on spread out over everywhere will be reduced and removed some time after this, due to said objects being accessible in localized areas before the planet’s surface.
- The power of the SMEs used was scaled with the original SMEs of the old outpost, considering they were overkill iirc.
- A few things to look out for are the power usage, and the ability of the air alarm’s heating between airlock usages. They should be good, in normal situations, but you never know. I’ll keep an eye out and see what it does live whenever somebody gets creative or hyper active with them.
- The un-used areas from the old outpost are still left in the code. I’ll get to them during the purging of excessive objects and areas discussed in the second bulletin after this map settles.
- The entire outpost is in a new style of sorts, when compared to the station with its attention to details. Most of it is similar to what the station already does with its decals, but there are a few glaring differences such as the windows. I may try re-vamping the design of the station in the future to match it.
- Those aren't full tile decals under the vending machines. The full tiles are currently broken for some reason. Will get to them eventually if they're not fixed.
- Egg.

outpost done


TheFurryFeline:
Why the advanced roller beds at the medical area? Those are upgraded gear for medical that's not normally available without science's help as well as mining.

Darlantanis:
This is a go from me as you've implemented my previous feedback and I trust you not to sneak anything in.

That said, I'd like this PR to be up for at least 2 days before acceptance. Given the scale of this re-design it wouldn't hurt to allow time for extra eyes to peer over it.

D-H-A:
> 
> 
> Why the advanced roller beds at the medical area? Those are upgraded gear for medical that's not normally available without science's help as well as mining.

They came with the southern cross map, but were removed before the transfer over due to the same reason you mentioned. After messing around with the job, they were implemented back, along with a rack of EMT belts to save the trouble of traveling to station medical for them, in order to free up space for both medical and explorer related items. If it becomes an issue of players constantly grabbing them for medical on station, I’ll either remove them entirely, or look into the locker’s access.

TheFurryFeline:
Alright, that was my only concern I had from an off-paw glance at the
difference rendering from MapDiffBot.

Razgriz1032:
I've tested this on a test server and it runs ok. Code looks good to me. Merging.

PR #274, opened 2020-03-22
Minor Outpost Changes

D-H-A:
- Removed the shutters from the ATV garage windows.
- Replaced the plating under the outside catwalks with planet-side plating.
- Replaced plating in ATV garage with normal plating.
- Flooring outside of mining outpost replaced.
- Slightly better lighting around three places that had shadows (Even though they were hardly visible shadows.)


TheFurryFeline:
There a chance this could be looked at soon? Would like to be able to separate all the main station z-levels soon onto each of their own files so the MapDiffBot works for it.

PR #285, opened 2020-03-30
[QoL] New Tile Sound Effects

Darlantanis:
So Sif won't be made out of metal anymore. 
-Adds 21 new footstep sound effects.
-Adds sound effects to dirt, sand, water, deep water, rock, and sif grass using new sound effects and some unused ones lurking in the code.
Shout-out to Kassc for sending me the sound files. 

Good sound design is most noticeable when it's missing.


Razgriz1032:
The amount of new sound files pleases me

PR #302, opened 2020-04-16
Bar's Safe, Folks. Drink On!

TheFurryFeline:
Changelog Notes:

- Changes rad storm announcement to include info about how the bar's shielded in inclusion to the maints and dorms as also stated.

### DIFF
--- a/code/modules/events/radiation_storm.dm
+++ b/code/modules/events/radiation_storm.dm
@@ -9,7 +9,7 @@
 	var/postStartTicks 		= 0
 
 /datum/event/radiation_storm/announce()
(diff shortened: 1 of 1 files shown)


Razgriz1032:
Can't you just say "shielded maintenance tunnels, bar, or dorms"?

TheFurryFeline:
Bar's specifically mapped in now to be the area of the desk and behind the shutter compared to Northern Star's old widespread dual area thing. The main area itself is mapped as a cafeteria with the bar at the north and the kitchen at the west. Therefore, it's "area of the cafeteria."

I'd rather not run the risk of it being inaccurately said over comms what's safe seeing that even the coffee shop is also mapped as a crew_quarters area, but will change it to be what you suggested there.

Razgriz1032:
Ahh, labeled as cafeteria. Interesting. Sounds good.

PR #304, opened 2020-04-23
Liquid belly refactor fix

Shadowfire117:
Re-added code injection to vorepanel_vr, though rate of ticks is different, so the rate of production and the amounts have changed, to properly fix bellies rate and amount, re-select liquid  and also production rate.


Dinnel:
Neat, but I'm worried about the cost of fluids. It seems to have went from 1.5 per whole unit of liquid to 15 per unit. Isn't that a bit too much?

TheFurryFeline:
Seems to be compiling fine, other than the runtimes caused by the sofas in the casino's map.

PR #321, opened 2020-05-02
Fix runtime in solargrub evolution

Razgriz1032:
I think it was angry because it was looking for multiple evolution forms. Now it's just one.

### DIFF
--- a/code/modules/mob/living/simple_mob/subtypes/vore/solargrub.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/vore/solargrub.dm
@@ -109,8 +109,13 @@ var/global/list/moth_amount = 0 // Chompstation Addition, Rykka waz here. *pawst
 
 /mob/living/simple_mob/vore/solargrub/proc/death_star()
 	visible_message("<span class='warning'>\The [src]'s shell rips open and evolves!</span>")
+
+/*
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_mob/subtypes/vore/solargrub.dm
```diff
@@ -109,8 +109,13 @@ var/global/list/moth_amount = 0 // Chompstation Addition, Rykka waz here. *pawst
 
 /mob/living/simple_mob/vore/solargrub/proc/death_star()
 	visible_message("<span class='warning'>\The [src]'s shell rips open and evolves!</span>")
+
+/*
+//Commenting this bit out. It's unncecessary, especially since we only use one form.
```
TheFurryFeline, line 114: If it's unneeded, then maybe remove it? No need for bloated code that's not even used.

Razgriz1032: Just in case we ever get around to balancing lunar moths and I don't want to lose the method of doing that.

TheFurryFeline: Fair nuff

PR #325, opened 2020-05-04
[Map] Sofa Runtime Squasher

TheFurryFeline:
Basically, on the tin. Fixes the cause of runtimes that makes Travis fails by temporarily changing sif_ora sofas with oldsofa.


TheFurryFeline:
Behold. No more runtimes.

Razgriz1032:
:HYPERS:

PR #328, opened 2020-05-05
No more solargrubs in docks

Razgriz1032:
Solargrubs don't seem to end up in any interesting places. Rather, they keep ending up in the docks. I think I'll just exclude the docks from the list and monitor where they spawn.

Also excluded them from spawning in dorms at the outpost.


Razgriz1032:
Needs testing. I will do that when I have time. Work week is busy

PR #330, opened 2020-05-06
Makes the teleporter stop killing you all the time.

BlackMajor:
That prob(5) is the defining reason why nobody ever wants to use the teleporter.
It's supposed to be called when the teleporter is inaccurate, such as back when teleporter calibration was a thing. But the teleporter doesn't do calibration anymore, so that accuracy var is always 0.

Stop throwing command staff into space plz.

### DIFF
--- a/code/game/machinery/teleporter.dm
+++ b/code/game/machinery/teleporter.dm
@@ -217,7 +217,9 @@
 				for(var/rider in L.buckled_mobs)
 					R.force_dismount(rider)
 		//VOREStation Addition End: Prevent taurriding abuse
+		//CHOMPStation start: Removes prob(5) of being yeeted
 		do_teleport(M, com.locked) //dead-on precision
+		//CHOMPStation end: Removes prob(5) of being yeeted
 
 		if(com.one_time_use) //Make one-time-use cards only usable one time!
 			com.one_time_use = 0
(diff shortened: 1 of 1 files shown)


code/game/machinery/teleporter.dm
```diff
@@ -217,10 +217,7 @@
 				for(var/rider in L.buckled_mobs)
 					R.force_dismount(rider)
 		//VOREStation Addition End: Prevent taurriding abuse
-		if(prob(5) && !accurate) //oh dear a problem, put em in deep space
-			do_teleport(M, locate(rand((2*TRANSITIONEDGE), world.maxx - (2*TRANSITIONEDGE)), rand((2*TRANSITIONEDGE), world.maxy - (2*TRANSITIONEDGE)), 3), 2)
-		else
-			do_teleport(M, com.locked) //dead-on precision
+		do_teleport(M, com.locked) //dead-on precision
```
TheFurryFeline, line 220: Add a Chompstation Edit comment per our standards. Otherwise, seems fine, but will leave the decision up to the headmins.

BlackMajor: Fek, I keep forgetting downstream modular thing.

Razgriz1032:
This PR is completely untested.

Calibration works. Right click the teleporter activation unit and hit test fire to calibrate. It works for 5 minutes or until shutdown, whichever comes first.

Denied.

BlackMajor:
Lmao shit there's a right click function?

TheFurryFeline:
There is, indeed. And that would be appreciated. Maybe add an option to cancel selecting a target. This would need to be ported upstream to Polaris, however.

PR #331, opened 2020-05-06
Fix Mappath

TheFurryFeline:
We had 2 UFO POIs, one never spawned because of a duplicated mappath. This fixes that. Beware the beasties within. This is an earlyport of https://github.com/VOREStation/VOREStation/commit/7066d04df5143b09ba17f39c319d97ab6fce7eca#diff-efe275666b5ad7aed46e76b55f32e6c5


Razgriz1032:
LOOOOL, good find

PR #383, opened 2020-06-16
Dorms-MK2

Kazzc:
Some minor fixes on the dorms and some minor map changes.

Second-Floor: 
-Added ladders on aft and starboard maintenance by the civilian area to grant a backdoor access to the third floor dorms.

Third-Floor:
-Added ladders on aft and starboard maintenance of the dorms to access second floor civilian area.
-Fixed missing cables so the missing dorms have power.
-Fixed a single simulated tile in space. (Active Edges: none)
-Fixed the breaker box by adding a missing cable.
-Added a light switch on the second floor bar
-Minor layout changes on maintenance
-Minor decal changes on the dorms
-Minor changes on random item drops.


Kazzc:
I left a single item that seems to be very destructive. lets see if I can fix it real quick.

PR #413, opened 2020-07-20
Borgs who drop modules with Q now refreshes the inventory

Shadow-Quill:
title, was annoying to deal with
porting up to polaris


CHOMPStation2:
Sounds like this could use porting upstream too. Seems good.

PR #416, opened 2020-07-22
Bandaid fix for transcore resource consumption

BlackMajor:
NOTE: I'd recommend test merging this for a while first to see if it alleviates the issue. Don't just full merge it off the bat.

So funny thing, the transcore saves the character slot and prefs of every mob that has a backup implant installed approximately every minute, per number of implants.

A body with 100 implants in it will be saved 100 times in one tick.

Yeah probably not a good idea.

-Disables the transcore implant mind backup
-Moves the implant mind backup to the post_implant proc on a backup implant, so you backup when you backup. (note, this means you need to backup again if you get an NIF to save it)
-Also adds some save preference stuff so you can't spam save/load to hitch the server

This is just treating a symptom however, the cause is of course how transcore is coded.
But uhh, figured it'd be a good idea to deal with the current issue earlier since stripping transcore is a big project.

EDIT: After test merging this for a full round, transcore is not responsible for the servers recent persistent OOM'ing.
Though this PR does still dramatically reduce transcore's cpu usage.

### DIFF
--- a/code/__defines/misc.dm
+++ b/code/__defines/misc.dm
@@ -349,3 +349,6 @@ var/global/list/##LIST_NAME = list();\
 #define JOB_SILICON_ROBOT	0x2
 #define JOB_SILICON_AI		0x4
 #define JOB_SILICON			0x6 // 2|4, probably don't set jobs to this, but good for checking
+
+//Preference save/load cooldown. This is in deciseconds.
+#define PREF_SAVELOAD_COOLDOWN 2 //Should be sufficiently hard to achieve without a broken mouse or autoclicker while still fulfilling its intended goal.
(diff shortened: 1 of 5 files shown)


Razgriz1032:
Wonder if there could be something added to NIFs that it procs the save. Something like,
`nif install
     if user has had a backup this round
          save()`

BlackMajor:
> Wonder if there could be something added to NIFs that it procs the save. Something like,
> nif install
> if user has had a backup this round
> save()

Issues for another day. The NIF user just needs to take literally any of the backup methods available to them post install to save it (or stay alive to round end).

KotetsuRedwood:
> Wonder if there could be something added to NIFs that it procs the save. Something like,
> nif install
> if user has had a backup this round
> save()

To be fair, there |is| a NIF program that allows for backing up on a buttonpress. It's not automatic, and it's one you'd have t o get from a vendor, but it does exist!

BlackMajor:
Oh right, I forgot to close this after we tested it and found it wasn't related.

PR #418, opened 2020-07-23
Misc memory improvements

BlackMajor:
This PR reduces memory usage by... rough estimate, about 200mb. Enough to let the server survive a 6hr round without OOM'ing.

Moves the associative ore_types list into a global list to reduce the memory usage of every single mineral turf having an associative list. Lists are expensive in byond.

Changes how airlocks generate their access lists, to only generate the list if they need them.

And some misc GLOB list improvements in cases where they were needed.


BlackMajor:
Bro you were supposed to test merge this first incase I broke anything important lmao.

Ah well, I test my code so chances of something breaking are smol..

PR #425, opened 2020-07-25
Disable Vchat

Razgriz1032:
Test merge for disabling vchat.


Razgriz1032:
VChat not a problem. Also verified that radio messages cause lag still. Might be something wrong with recent updates to communications instead.

PR #429, opened 2020-07-26
Genetics and goia parts

Shadowfire117:
Not that they have to do with each other!

Changes:
Made genetics lab in the dedicated room for it
gotten rid of cloning room, it wasnt needed anymore
shifted morgue to cloning area
made a comfy corner in medical patient wing to space
moved equipment storage into small room to use the space
shifted bathroom a little
Geneticist is now select able again

Goia parts! Theres a tail, ears, and two taur bodies, one slim, one chonk


vorestation.dme
```diff
@@ -1732,10 +1732,10 @@
 #include "code\modules\ai\ai_holder_targeting_vr.dm"
 #include "code\modules\ai\interfaces.dm"
 #include "code\modules\ai\say_list.dm"
-#include "code\modules\ai\ai_holder_subtypes\simple_mob_ai.dm"
```
Razgriz1032, line 1735: Whatever this is doing, don't let it do this.

Razgriz1032:
I don't like what is happening to vorestation.dme. Grab updated dme from master.

vorestation.dme
Razgriz1032: Just grab the .dme from master

PR #439, opened 2020-07-29
Solar Moth nerfs

LunarFleet:
Reduces the maximum temperature to 450 C, instead of being 10k C, which is twice the temperature of the sun. Also reduces some other stats, which made very little sense, such as solar moths being immune to lasers.

### DIFF
--- a/code/modules/mob/living/simple_mob/subtypes/vore/solarmoth_ch.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/vore/solarmoth_ch.dm
@@ -17,19 +17,15 @@
 
 	var/mycolour = COLOR_BLUE //Variable Lighting colours
 	var/original_temp = null //Value to remember temp
-	var/set_temperature = T0C + 10000	//Sets the target point of 10k degrees celsius
-	var/heating_power = 100000		//This controls the strength at which it heats the environment.
-	var/emp_heavy = 2
-	var/emp_med = 4
-	var/emp_light = 7
-	var/emp_long = 10
+	var/set_temperature = T0C + 450	//Sets the target point of 450 degrees celsius
+	var/heating_power = 100000	//This controls the strength at which it heats the environment. // The number seems ridiculous but this is actually pretty reasonable - Lunar
 
 	faction = "grubs"
-	maxHealth = 400 // Tanky fuckers.
-	health = 400 // Tanky fuckers.
+	maxHealth = 75 // Tanky fuckers.
+	health = 75 // Tanky fuckers.
 
-	melee_damage_lower = 1
-	melee_damage_upper = 5
+	melee_damage_lower = 5
+	melee_damage_upper = 10
 
 	movement_cooldown = 5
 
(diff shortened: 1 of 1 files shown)


Razgriz1032:
I suggest nix all changes except for the temperature. Temperature change is good.

Change health to 200 and change melee armor to -50, change bullet armor to 0.

Leave EMP alone, it's only a localized blast. Leave lunarmoth alone, it's not even in any spawn rotation - that needs review.

Commenting code out in the future might be preferable for removals and such.

CHOMPStation2:
Just to log here, I think the final changes are good. The only thing I am strongly against is removal of EMP. Over our discord discussion, we clarified that the EMP only procs on death.

Kazzc:
One of the things I like to consider in regards of this mob is this.
-Can spawn when not having engineers
-5K can seriously grief the entire station, even more if a player wants to be cheeky
-Should be a simple mob with a gimmick and plays in to it nicely.

Why the nerfs?
-We have no mobs or barely any with actual armor values, if they have armor they is no more than 20% reduction. Having a mob with glaring 100% resistance feels very unbalance and needless punishes a player, already wasting their time and resources dealing with them in a less efficient way should be punishing enough.
-450 is good enough temp to harm a player and not do massive glaring damage by ignorant or reckless players to the station. One will still need a suit, so it fits the bill
-EMP can be more dangerous than you think, it can electrify doors, cause the room to emergency vent and break computers and machines to a point they need to get fixed again. (I am okay with having EMP on but is something to note)
-Increasing its base damage should be good as anything above 15 allows a mob to break reinforced glass, should compensate for how it use to heat up a room. 20 damage should be good for them to break glass and breach out to players that decide to kill it with laser guns, while the moth itself has 50% energy resistance to tank the energy damage long enough for it to escape.
-Unsure of explosion resistance?, why does it have explosion resistance?, the chances of it being hurt by explosions?... I guess 25% is fine but explosion was nerfed greatly on mobs now as hallows does not turn in to damage anymore.

These are my toughs on the matter.

Admiralkosmov:
I'm just gonna restate what I said on the discord.
I really don't like this. The whole point of solar moths being tough to kill was to incentivise not killing them. They aren't aggressive, you can literally just walk up to them and drag them out into space. The nerfs only make it easier to kill them, which is not what the intended strategy to handling them was when they were created. As for the temperature, I really don't care much about whether it's 10000, 10000000 or 450 degrees, as the solution in each case is the same - your end goal is still reducing the temperature to a manageable level and releasing them through the nearest opening into space. Adjusting their health and armor doesn't adress the issue of "nobody knows how to handle them properly", it only enforces the "oh yeah, I'm just gonna shoot it until it dies" mindset.

Darlantanis:
Have we considered a weaker version for the solargrubs event with the stronger version for mapping/staff events?

LunarFleet:
Nearly every thing I have changed has been a variable, meaning any staff can easily buff or nerf the solar moth back to whatever level they want. That being said, I will happily add a buffed variant if someone like Kassc or another event manager wants one, because I am happy to give more event material.

PR #474, opened 2020-08-12
Adding bluespace to our universe.

Kazzc:
So this is a test-bench in hopes to fix our OOM issues.
I reduced all map sizes from 255x255 to 185x185
Some things to note in this changes...
-Station is still the same, just has a lot less space around it
-All shuttles now dock closer on the station
-Transition map. Shuttles have very small transition areas, one can see other shuttles or see world borders in some shuttles (Will fix later)
-Plains, Wilderness and Caves are smaller so have smaller amount POI generation, I still made it more generous so finding POI's will be easier but still less numerous than the 255x255 map.

I been testing this for hours, all transitions between vertical z-levels on SouthernCross function, transition between caves, wilderness and plains function, transition between shuttles work... About everything is exactly the same just smaller.

Please do note that this is a test, do report any bugs or inconsistent behaviors of any map.


Razgriz1032:
Mapdiff can suck a cock

Razgriz1032:
According to Travis, your POI's are failing

PR #490, opened 2020-08-14
Disable casino map temporarily

Razgriz1032:


Razgriz1032:
Runtime be damned, server will live.

PR #492, opened 2020-08-14
Kassc's payday masks

LunarFleet:
The title says it all


Darlantanis:
I've noticed you put the sprites into _vr .dmi files. Have you sent this to VOREstation? _vr typically denotes a VOREstation-specific file, if these are just for us (and the code doesn't disallow it for some reason) the proper naming would be _ch. E.g., mask_ch.dm.

If these must be added to miscellaneous.dm you will need to include "CHOMPedit" in the comments (preferably with "CHOMPedit end" wherever you leave off. However if it's possible without making the code a mess it'd be better to put these in miscellaneous_ch.dm (which probably doesn't exist yet).

I'm not sure if anybody's explained the _ch files or our commenting policy so I'll elaborate a bit. When we do upstream PR's, any file we pull will conflict if it's been modified here which will need to be manually resolved. When upstream files must be modified for a feature, including "CHOMPedit" makes it super easy to find edits with CTRL+F and thus avoid overwriting any of our changes. _ch files meanwhile are preferable where possible as they will never conflict as they should not exist upstream.

Darlantanis:
I should probably also add that anything submitted (and accepted) upstream will not need "CHOMPedit" comments since it is now an upstream feature. Calling it an "early port" cues us in that it's been submitted upstream.

LunarFleet:
See, I did that before with other PRs, but I worked on this PR with two hours of sleep, so I completely forgot. I'll fix it when I wake up in a few hours. Same with the flechette merc nerfs.

LunarFleet:
Well, we have run into an issue.
Due to how the species specific sprite sheets are set up, you cannot have two files fit in for one spritesheet. IE, every sprite for a certain kind of item; masks, space suits, etc, has to be in its OWN file, for EACH species. As a result, I cannot set a _ch file for my additions. I have left _ch .DMIs which only contain the CH additions, but they are merely there for ease of porting in the future. I don't really see any other way to go about this, sadly.

Darlantanis:
If it's unavoidable (because code) it's acceptable to not use _ch files. I would recommend explaining the need for this (via comments) in miscellaneous_ch.dm so we have documentation for this if it causes conflicts a year or two from now after everybody has forgotten about this PR.

Razgriz1032:
I still see change to Tajaran VR file?
image

ExabyteDiscord:
Assigning WIP label as looking through this seems there was still stuff needing to happen but no changes have been made since Raz's comment.

LunarFleet:
Actually this isn't really WIP unless y'all are 100% sure you want the CH files removed. Other files seem to have that same separation, so its easier to tell what icons are from what server, as you can't really mark DMI's with // CH edits

Darlantanis:
Yeah, this is a weird case where it's not possible to make the code work with _ch files without weird code edits. I'll ask Raz to be certain but this seems like the sort of situation we'd make an exception for.

Darlantanis:
Guess I'm wrong. Raz's answer is to check hardsuit code to see how to make the masks point at _ch files instead of _vr files.

Razgriz1032:
https://github.com/CHOMPStation2/CHOMPStation2/blob/master/code/modules/clothing/spacesuits/rig/suits/station_ch.dm

It's a sprite sheet somewhere. It needs to be made to point at the ch files.
Notice in the sprite sheet how a new object was created and everything is being pointed to /obj/item/clothing/head/helmet/space/rig/**ch**. 

There should be a sprite sheet for clothing that is a sprite sheet and **/ch** can be added to the end of it and the sprite sheet defined to point at the _ch icon files.



I am not going to accept changes to _vr icon files since those are being constantly updated.

LunarFleet:
Ah, when I looked over the files, all the sprite sheets were listed further down the typepath, I'll give it a look.

Razgriz1032:
And now there are lots of conflicts. This is why having them in _ch files is optimal

LunarFleet:
Wait
This PR is not closed?
This was the last PR I  made before falling into the tired dimension. I'll just close it, I thought it was closed weeks ago ngl.

PR #496, opened 2020-08-16
Unify Fabricator and some POI fix

Kazzc:
-Modified Lonehome and their respective files to make travis happy abot this POI
Changed 'abandonedhome.dmm' to 'lonehome.dmm', changed all its filepath names and area name to 'lonehome'.
-Added Unified drone spawners on the second floor, engineering. turned on it's DM file.


Darlantanis:
You say this fix will make Travis happy but it has not.
image

PR #500, opened 2020-08-17
Flechette merc nerfs

LunarFleet:
Decreases the damage, armor pen, accuracy, and firerate.


Darlantanis:
You will need "CHOMPedit" comments in magnetic.dm or will need to move the projectile type to a _ch file.

PR #512, opened 2020-08-23
Give HoS and CMO Teleport Access

cadyn:
I can understand why RD and CE have teleport access, but I feel like if we're gonna give it to two heads we might as well just give it to all of them. Would make a lot of people's jobs easier as well, especially CMO for surface rescues.

### DIFF
--- a/code/game/jobs/job/medical.dm
+++ b/code/game/jobs/job/medical.dm
@@ -16,10 +16,10 @@
 	req_admin_notify = 1
 	economic_modifier = 10
 	access = list(access_medical, access_medical_equip, access_morgue, access_genetics, access_heads,
-			access_chemistry, access_virology, access_cmo, access_surgery, access_RC_announce,
+			access_chemistry, access_virology, access_cmo, access_surgery, access_RC_announce, access_teleporter,
 			access_keycard_auth, access_sec_doors, access_psychiatrist, access_eva, access_external_airlocks, access_maint_tunnels)
 	minimal_access = list(access_medical, access_medical_equip, access_morgue, access_genetics, access_heads,
-			access_chemistry, access_virology, access_cmo, access_surgery, access_RC_announce,
+			access_chemistry, access_virology, access_cmo, access_surgery, access_RC_announce, access_teleporter,
 			access_keycard_auth, access_sec_doors, access_psychiatrist, access_eva, access_external_airlocks, access_maint_tunnels)
 
 	minimum_character_age = 25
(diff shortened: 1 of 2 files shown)


Holdelta:
As something I've noticed, might want to make sure the turrets inside the teleporter stop targeting heads in general if this goes through.

cadyn:
> As something I've noticed, might want to make sure the turrets inside the teleporter stop targeting heads in general if this goes through.

Giving them access to the teleporter usually makes the teleporter not target them

Razgriz1032:
But have you tested it

cadyn:
> But have you tested it

Yea, ofcourse I did. When I used it, the turrets didn't fire at the CMO or HoS, and both could open the door.

Darlantanis:
Personally not in favor of this. CE/RD are granted teleporter access as part of telecomms access. CMO/HoS/HoP lack access as it is not granted for being Command.

cadyn:
> Personally not in favor of this. CE/RD are granted teleporter access as part of telecomms access. CMO/HoS/HoP lack access as it is not granted for being Command.

I don't really see what the teleporter has to do with telecomms other than having the prefix "-tele". And yeah, I kinda feel like It should be granted for being command, especially CMO would get some really good use out of it.

LunarFleet:
I wondered why this isn't merged yet, so, I'm here now. I want to mention that CMO is **easily** the head of staff who needs teleporter access the most. Rapid responses to the surface are very very useful when dealing with IB, organ failure, etc. CE/RD may have it because of telecomms access, but CMO has just as much reason to have it _due to it shaving a nice threeish minutes off a medical response time_

I make no argument for HoS

Razgriz1032:
I want to say yes

LunarFleet:
Have you considered making your dream something more than a meme, and saying yes? Do not live in the realm of fantasies, do not let reality stop you, you can do it, you can say yes.

Razgriz1032:
I just realized that the turret in the teleporter might aggro if the HoS and CMO aren't on the turret's IFF list. IDK where that list is.

cadyn:
> I just realized that the turret in the teleporter might aggro if the HoS and CMO aren't on the turret's IFF list. IDK where that list is. It might be embedded in the map?

The turrets have a list of access requirements(which is currently just set to access_teleporter) which they check against humans to see if they should or shouldn't be there. Since we gave the CMO and HoS teleporter access with this PR, it'll also make the turrets not aggro them

Razgriz1032:
I'm merging this and we'll see how it goes :P

PR #517, opened 2020-08-24
Optimizing machines controller.

cadyn:
Machines controller seems to take a lot of processing power, so I figured I'd give it a shot and try to optimize it. I got a solid start but around 50ms of time spent in the controller isn't logged so 🤷 Gonna try to work on this some more and absolutely taking suggestions


cadyn:
Gonna just close this since if I ever do get around to doing optimization work, I'll probably do it upstream.

PR #520, opened 2020-08-25
Making sif uninhabitable for most, as it should be.

### LINKED ISSUE #423: Cold weather on Sif is not harmful.
Under normal circumstances on Sif, such as going to the planet with no cold protection, does not result in any temperature based harm, regardless of the weather pattern outside and time spent within it.

To sit in a blizzard, naked, and freeze to death. In all seriousness, I expect to have SOME repercussions for not wearing a winter coat, much less anything at all.

#### What actually happened
Sat in a blizzard, naked, and got nothing but goosebumps.

#### Steps to reproduce
- (Step 1) Got to Sif.
- (Step 2) Take off all your clothing, with just base human traits.
- (Step 3) Wait for a blizzard, storm, or other such super cold weather pattern.
- (Step 4) Bask in the winter chill without worry.

#### Code Revision
 
mM6F4AU
Not me, but I can get pictures myself if you'd like. I experienced much the same.

cadyn:
So, -30C in casual clothes is a good way to get fucked up in real life, why shouldn't it be the same ingame? Even with this change, the temperature without requiring a winter coat is honestly still unreasonable, but I feel like we should at least get it up this much.

Closes #423


Somememeguy:
I do not like this. I really enjoy Sif being habitable. More then that, the thing you just adjusted applies to all temperature changes, not just ones on Sif. If you want to make Sif uninhabitable, you should make Sif colder, not just change a variable that effects everything. Otherwise, I feel like this is a really really bad change.

Razgriz1032:
I think it is a map-specific issue because Yawn Wider's cryogaia is chilly, so is Virgo, but not Sif for some reason despite what the readouts say. Also what happened was Polaris broke it somehow and they use nearly the same map. Double also, Polaris is considering reworking the temperature system because right now temperature is a binary system: you're either cold or you're not. So they're gonna be reworking it anyway.

cadyn:
Honestly, I still feel like this change should be made because at the moment, you can literally stand naked in a -30C environment for as long as you want, which is totally fucked. Either way, gonna send a hypothermia PR down the chain once I finish tweaking it.

cadyn:
> I do not like this. I really enjoy Sif being habitable. More then that, the thing you just adjusted applies to all temperature changes, not just ones on Sif. If you want to make Sif uninhabitable, you should make Sif colder, not just change a variable that effects everything. Otherwise, I feel like this is a really really bad change. More then that, I do not believe that variable actually makes it so that you take cold damage at an earlier stage? By the looks of it, it only makes it so you get cold faster.

Also, if you look at the actual code this is the way it works:

When you go into a cold environment there's two rates affecting your body temperature.
Your metabolism is working to keep your body warm on one hand
On the other hand, thermodynamics is working to bring your body down to the temperature of the environment.

Eventually both of these forces will reach an equilibrium and your body temperature will stabilize somewhere

And the variable I changed affects the second rate, which means that the two forces will reach equilibrium at a colder temperature, which means that going into cold environments brings you below your species maximum cold level more often, and also applies for Sif. It's not a terrible a change, it's just a gentle nudge in the direction of reality.

PR #524, opened 2020-08-28
Fixing Xenomorph life cycle + numerous changes and balances.

MisterGrimm:
So this has been a very annoying project of mine in a attempt to fix the xenomorph life cycle, and breath some actual life into them. The main point of this is to make it where xenomorphs can reproduce on their own, so event admins don't have to watch every little thing they do to keep their numbers up.

These changes include : 

- Changing aliens_allowed config to 1 so xenomorphs can lay eggs, the ability to turn this to 0 ingame exists via admin permissions.
- Adds new Xenomorph eggs with new Xenomorph egg code.
- Removes broken facehuggers, embryos, and old xenomorph eggs since they don't work.
- Xenomorph eggs now produce a larva, with the egg acting in a similar manner to a ghost trap. Simply click as a ghost when it is fully grown, and you will be prompted to become a larva if you are not banned from it. Setting aliens_allowed to 0 prevents eggs from being laid.
- Extends Larva growth to 325, from the original 200. Attempted balancing act since xenomorphs don't need hosts now.
- Larva can see their growth stat now, thank god.
- Changed larva max health from 25 to 50, in a attempt to balance the extended larva growth.
- Removed nutrition gain from Radiation for xenomorphs, and incidentally diona. It would cause immense obesity, and potentially rupture your appendix. All other radiation effects to xenomorphs/diona are unchanged.
- Attempted to balance xenomorphs by making them a little more vulnerable to brute damage, while lowering their burn vulnerability some. This may require a metric ton of testing to get it right.
- Added very good darksight for xenomorphs, so they can pretty much see anything in the dark. They still have to adjust to it however.
- Tried to give xenomorphs thermal-like vision, not sure if it actually worked.
- **NEW 64x64 XENOMORPH SPRITES**, looks kind of like a eldritch dog/fox now. They can easily be reversed if people don't like them, caste sprite differences included.
Capture
- Had to make a completely new pixel_x type called pixel_offset_x for datum/species, because larva were being a bitch at messing with 64x64 subspecies for xenomorphs. Will be very useful in the future if more 64x64 species get added.
- Simplemob xenomorphs **do not have changed sprites**, they still look the same.
- Changed some visual messages to line up to the new xenomorph look.
- Gave queen a little more health, since they are slow as hell. 250 to 300.


MisterGrimm:
Planning to get some help with the POI's, as they were compatible when I originally started this project. I guess someone has been adding stuff to the POI's.

Darlantanis:
You'll need to add "CHOMPedit" comments to any non _ch files you've edited, this is how we keep track of our content should an upstream PR conflict with anything.

We've got at least one POI with xenomorphs: how will said POI be affected? I'm a little concerned if it could cause a xenomorph outbreak should it spawn on Sif since the design is based on xenomorphs which cannot infect/spread. Same concern goes for xenomorphs in the deadly predator crate.

MisterGrimm:
POI with xenomorphs won't spread as long as there isn't any eggs around, as simplemobs do not act the same way as player xenomorphs. Any xenomorph that is designated a simplemob cannot do a single thing player xenomorphs do, so as long as there are no xenomorph eggs around it won't change anything. If there is any eggs, however, it will. From what Kassc was telling me there was only three POI's that have eggs, so they will have to be removed. I can get Kassc to help me on that.

Ill also get onto adding CHOMPedit to my edited stuff, although I personally doubt there will be any updates upstream for xenomorphs, considering how left in the dust they are.

MisterGrimm:
Was there anything else that needs to be discussed, or were those the only points of discussion you were worried about?

Current discussions :

- [x] CHOMPedit needed on edited code.
- [x] POI xenomorph eggs needing changed to xeno container eggs to prevent player larva infestations.
- [x] Explanation that simplemob xenomorphs are completely different from player xenomorphs, e.g. POI xenomorphs and predator crates only have simplemobs.

Issue #532, opened 2020-09-06
Giving pilots more gear and access


123MrMango321:
Pilots lack access to several exploration doors (The prep rooms in front of each hangar on the station to be exact), lockers (The gun lockers on the carrier's hangar are locked to pilots, but not the one in the shuttle), and do not have any catalogers, forcing them to shoot open explorer's lockers in order to get any.

I'd like for Pilots to have as much access as explorers, unlocking doors and gun lockers that they can't get to. And give them catalogers in their own lockers so they don't need to take from explorer's

PR #529, opened 2020-09-05
Clear Skies Ahead!

ExabyteDiscord:
Clears up a hell of a lot of clutter with all of the pre-defined areas of shuttles to properly use the landmarks system. Now there's a ton of free room to build and/or allow larger shuttles to dock with the station, as well as allowing overmap shuttles to dock at the docks.

PR #539, opened 2020-09-08
Pilot/Field medics given access to exploration doors

cadyn:
Pilots and field medics given access to exploration doors where access was previously missing, and catalogers added to lockers.


cadyn:
Closes #532

Darlantanis:
Holding off on map changes until #529 is acceptable as it'll be much more painful to fix conflicts there than to small tweaks like this.

cadyn:
Waiting for #529 to merge before reopening

PR #552, opened 2020-09-18
Machines.dm Optimization + Bots improvements

cadyn:
Ok, this is like one of the few things I'm not sending upstream because I'm honestly feeling a bit grumpy and lazy about upstreaming shite rn, so I just slapped some Chompedit tags on there. Upstream can come and get it if they really want it 3:<

Aaanyways here's a list of things done:
Machines processing should be slightly faster now by switching a branch to a bitwise operation.
Cleanbots don't fucking spaz out and spam pathfind to space anymore, instead if an objective is repetitively not able to pathed too, they'll wait longer and longer before trying again. This also applies to all bots.
Bots patrolling will now go around obstacles that weren't in the initial path if given long enough


Darlantanis:
I think this is good but I'd prefer Raz to look over it, subsystem tweaks are a fair bit beyond my comfort zone with code review.

PR #562, opened 2020-09-30
non-euclidean portals

Kazzc:
-SC2
*Added a missing table on Chemistry
-SC10, SC6, SC5.
*Added non-euclidean portals to make transition between outpost, caverns and wilderness smooth if not seamless.
-POI Spatial_anomaly.dmm
*Removed blocks that prevent entry to this POI for testing, apparently there is a chance to crash a players client if they ghost over the POI or access the POI with masons/material scanner/x-ray.
I can't seem to crash my own client from this so I will like to get some crash reports from players.


Kazzc:
hrrmmfff... nvm, do not merge. I may have messed something up.

PR #568, opened 2020-10-06
Direct pull from virgo

cadyn:
Did a direct pull from virgo because I noticed there were a lot of bugs we were having that were already fixed up there and not on YW either, so I decided to just yoink this straight from virgo. Also includes my own fixes for CD stuff with chomp files. <3


cadyn:
Ree, fuck you too Travis

PR #571, opened 2020-10-07
Fix Web Shuttle Sensors

ExabyteDiscord:
Web shuttle sensors didn't display correctly. Now they do! Yay.


ExabyteDiscord:
Suppose I'll just merge this myself, literally just a 1 line fix. tgui.bundle.js is just the compiled version of tgui so

PR #574, opened 2020-10-08
Give Quick NIF to the boys

cadyn:
Give the boys quick NIF


Razgriz1032:
We already have that??
https://github.com/CHOMPStation2/CHOMPStation2/pull/395

PR #577, opened 2020-10-09
Removes loss POI, gnome, pufferfish and ack

ExabyteDiscord:
Begone demons


cadyn (changes requested):
Removing these sounds would destroy the integrity of chompstation13. We need to keep them. Please uncomment these.

Razgriz1032:
The only one I want kept is ack :pepehands:

PR #579, opened 2020-10-11
Xenomorph Balancing Act 1.1

MisterGrimm:
Simple balances and fixes due to a recent event.

- Changed plasma cost for laying eggs, from 75 to 200.

- Changed Agony damage of neurotoxin, from 80 to 60.

- Finally fixed the fact that xenomorphs cannot see in the dark. Xenomorphs can now see turfs and mobs, but lose some range when in the dark.

- Changed ability name Gut to Slaughter, to more accurately describe what it does.


MisterGrimm:
Travis why do you hate me. e3e

PR #581, opened 2020-10-13
Fix resleeving

cadyn:
Currently, there is an issue which causes people's colors to get messed up by resleeving. This PR fixes it. Pretty darn simple :p


cadyn:
Travis why

PR #582, opened 2020-10-14
TGUI fix

cadyn:
Fix for TGUI stolen from someone's branch from presumably an in-progress PR for virgo. Shouldn't need chompedit comments since it will almost definitely end up on the upstream branch as well.

Edit: Was actually just merged upstream lol https://github.com/VOREStation/VOREStation/pull/9218


cadyn:
Closes #583

PR #585, opened 2020-10-14
Added TGstation scream sound effects, and tweaked scream sound effect code.

MisterGrimm:
- Added four default scream sounds for Male.

- Added four default scream sounds for Female.

- Tweaked emote.dm scream sound code so it can pick from a list, instead of being limited to a single scream sound effect.

- Tweaked emote.dm scream sound code so it can be turned off in emote preferences.


code/modules/mob/living/carbon/human/species/species.dm
```diff
@@ -75,8 +75,8 @@
 
 	//Soundy emotey things.
 	var/scream_verb = "screams"
-	var/male_scream_sound		//= 'sound/goonstation/voice/male_scream.ogg' Removed due to licensing, replace!
-	var/female_scream_sound		//= 'sound/goonstation/voice/female_scream.ogg' Removed due to licensing, replace!
+	var/male_scream_sound = list('sound/effects/mob_effects/m_scream_1.ogg','sound/effects/mob_effects/m_scream_2.ogg','sound/effects/mob_effects/m_scream_3.ogg','sound/effects/mob_effects/m_scream_4.ogg','sound/effects/mob_effects/wilhelm_scream.ogg') //CHOMpedit start : Added tgstation screams
```
Rykka-Stormheart, line 78: Wilhelm Scream is already used by deathgasp2, btw.

MisterGrimm: We have a death gasp? I have not once seen or heard of it.

MisterGrimm: I think it should be fine if they both exist, considering deathgasp2 only triggers if you manually trigger it yourself, of which I have never seen actually done.

Rykka-Stormheart: > I think it should be fine if they both exist, considering deathgasp2 only triggers if you manually trigger it yourself, of which I have never seen actually done.

It's used plenty on Virgo, and your rand list doesn't include any sorting by rarity, so the wilhem has the same 1/5 chance of being picked as the rest. It's going to quickly go from 'haha easter egg' to really overdone.

Darlantanis:
I have one significant issue with this PR. All of these screams sound very human which is somewhat troublesome in a server where human players are a minority. I could see this working if these applied only to humans (or human-based species) but it's somewhat immersion breaking for, say, vulp or unathi screaming like this when taking too much damage.

They also seem a bit lrp-ish but that's mostly personal opinion. 

Side thought, will this apply to bone-breaking during digestion? Will it lead to random screaming in a crowded bar due to vore scenes?

MisterGrimm:
> 
> 
> I have one significant issue with this PR. All of these screams sound very human which is somewhat troublesome in a server where human players are a minority. I could see this working if these applied only to humans (or human-based species) but it's somewhat immersion breaking for, say, vulp or unathi screaming like this when taking too much damage.
> 
> They also seem a bit lrp-ish but that's mostly personal opinion.
> 
> Side thought, will this apply to bone-breaking during digestion? Will it lead to random screaming in a crowded bar due to vore scenes?

1. All species have a way to change their scream sound effects in the species datum. Some species already have these sound effects replaced, some don't. The scream sound effects I have added are the default sound effects, and only pull those sounds if there is not a species replacing them. So these would basically only be for humans, and species that do not already have a custom scream sound effect. Custom scream sound effects can easily be added by adding **male_scream_sound =** and **female_scream_sound =** to a species, with the = having the pathway to the sound. If this PR goes through I plan to add at least one scream sound effect for every species, one for each gender.

2. Digestion screams were fixed years ago by virgo, they should not make any noise from it due to bellies preventing the scream emote. Unless you have a counter argument to prove this false.

3. Since digestion screams are being prevented (_See 2._) the only way one could scream in a belly would be to manually do the emote. Obviously doing that manually in a bar full of people would already be borderline pref breaking without the sounds, so I would assume the sounds would only bring the bwoinking faster.

Darlantanis:
I'm aware you can set the scream sound per species, the concern is only vox and tesharii have different scream sounds currently so the default will apply to most species. Perhaps manually setting to null for now would be a better work around as I kinda doubt we have anybody with a bunch of different scream sounds ready for implementation. It also makes me wonder how to make sounds like this work for custom species, iirc they all use the same species datum.

I'm admittedly surprised bellies mask the scream sound (rather than just text), that's a little more thorough than I expect of upstream code.

MisterGrimm:
> 
> 
> I'm aware you can set the scream sound per species, the concern is only vox and tesharii have different scream sounds currently so the default will apply to most species. Perhaps manually setting to null for now would be a better work around as I kinda doubt we have anybody with a bunch of different scream sounds ready for implementation. It also makes me wonder how to make sounds like this work for custom species, iirc they all use the same species datum.
> 
> I'm admittedly surprised bellies mask the scream sound (rather than just text), that's a little more thorough than I expect of upstream code.

Yeah, I can totally do that. It would probably be best anyway. The custom species will be a tricky one, and I may need outside help to give them a list to pick from in character creator. Otherwise ill just set them to null too since custom species is going to be the problem child with this project.

And yeah, years back on virgo everyone got irritated of the screams in bellies, so they had to find a way to prevent the scream emote from happening. Breaking bones, damage in general, and agony damage have if clauses in them to prevent you from using the scream emote automatically if you are in a belly. If you search **!isbelly** in VSC you will quickly find all of the if clauses preventing someone from screaming in a belly.

MisterGrimm:
There, this should make it where most species can't use scream sounds now. I didn't add it to species that are already very human-like, like the replicants and vatgrown as examples. This should add a decent framework to begin adding scream sounds to just about any species, which I plan to eventually begin working on later down the line.

PR #586, opened 2020-10-14
Blitz rods

cadyn:
Finally made this PR

- Blitz rods are made by first attacking a fuel rod compressor with a stack of fewer than 25 sheets of supermatter, after which one sheet will be consumed, then adding 25 sheets of phoron. Finally, you must add a lead shell to the unshielded blitz rod to give it some shadow of stability.
- Blitz rods can penetrate through one wall or mob before they will impact with their final target and explode, hijacking code from the SM crystal to either qdel the impacted target or "dust" them if they're living.
- Throwing blitz rods will cause a smaller explosion and the blitz rod to shatter.
- Trying to fire an unshielded blitz rod will cause you to explode.
- Firing a shielded blitz rod will cause a blinding flash and deafening boom which wall knock close players over as well as destroy the fuel rod cannon it's being fired from
- Trying to use a blitz rod in a fuel injector will result in an alert asking if you're sure you want to continue before the fuel injector shakes and then explodes.
- Trying to hold an unshielded blitz rod without gauntlets or thick gloves will result in burns.
- Unshielded blitz rods also give off passive radiation.

Are blitz rods OP? A little bit, but it's balanced by the fact that they are stupidly difficult to get a hold of. Supermatter crystals are notorious for being difficult to obtain, I would reckon it would take someone experienced at least two or three hours to make one. It's also balanced by how dangerous it is, with simple mistakes easily leading to the creators death or worse, and even avoiding all mistakes, once you fire it, you will be knocked over and likely stunned for a bit. 

What do they add to the game? The way I see it, this adds mostly to the science experience. Currently the only real insane difficulty goal I can think of is making a janus mech, which is yards beyond the difficulty of other goals a scientist can strive to achieve. This adds something between the insane difficulty of the janus mech and the other possible science goals which should open up players to trying to learn how to use the particle accelerator to create things like supermatter. It also is a fun little thing for admins to potentially spawn or to even be put in POIs if you're bold enough. The satisfaction of creating one of these is seeing your product and being able to fire it and see the stupid power which you wield, even if only once.


cadyn:
Edit: Blitz rods no longer qdel anything and only dust you if you're a mob with a maxHealth less than 200

Darlantanis:
Because of the strength but extreme difficulty of obtaining these it's difficult to predict balance/how abuseable this is just from code. We've decided to tentatively accept this with the caveat it may be nerfed or reverted in the future depending on how it's used in-game.

PR #602, opened 2020-10-16
Only eggs can sustain me

LunarFleet:
Fluid bellies cover milk/honey/etc, which was previously covered by reagent implants. But what were egg implants replaced with? A trait, only available to  custom species. 0/10. 

So here we are. Most of this is based on virgo's reagent implant, and their old fluff reagent implants, though I have polished it up so it is less clunky and barebones. **These implants are accessible under the utility tab of the loadout menu.** All of the variants are in one selectable list, because, according to virgo fluff implant comments, having two implants at the same time will break it, so you may only have one at a time. I made a few different, simple variants based on people's tastes. For reference, based on comments on the original egg laying implants, a maximum volume of 3000 takes roughly thirty minutes to fill, meaning 100 takes a minute. 

There are three sets of variations, though none of them have a transfer amount lower than 50. The transfer amount, ultimately, effects how long it takes to produce an egg.

First variant is just slower, one version taking thirty minutes to fill, second taking an hour. 
Second variant just effects the capacity, not the recharge rate. First has a maximum capacity of 20, second has one of 40. 
There is no mechanical benefit to this, I just figured some people would have liked more  e g g. 
Third and final variant is slow and lower capacity (Only one egg at a time), once again, thirty minutes for the first verison, hour for the second. 

These take nutrients to create, judging by the /obj/item/weapon/implant/reagent_generator which it is based on. So, it should be balanced. Honestly, I haven't run the math, but looking at the original reagent generator implant, then at the egg one, I feel like this is gonna drain your nutrition a lot, and testing seems to support that.

With the big explanation (mostly for non dev dorks who wanna know how the implants work) over, I will say that i have tested this with no noticable issues, and have gotten no compiler issues. So, y'know

https://www.youtube.com/watch?v=xYDpz4nBdNg

### DIFF
--- a/code/game/objects/items/weapons/implants/implantreagent_ch.dm
+++ b/code/game/objects/items/weapons/implants/implantreagent_ch.dm
@@ -36,19 +36,26 @@
 			if(istype(I, /obj/item/weapon/implant/reagent_generator))
 				rimplant = I
 				break
+
 	rimplant.empty_message = list("Your lower belly feels smooth and empty, clearly there are no eggs left to be had!", "The reduced pressure in your lower belly tells you there are no eggs left, for now...")
-	rimplant.full_message = list("Your lower belly is a bit bloated, possessing a mildly bumpy texture if pressed agaisnt...", "Your lower abdomen feels really heavy, making it a bit hard to walk.")
-	rimplant.emote_descriptor = list("an egg right out of [src]'s lower belly!", "into [src]'s belly firmly, forcing him to lay an egg!", "[src] really tight, who promptly lays an egg!")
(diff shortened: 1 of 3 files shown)


cadyn (approved):
So, this PR is very interesting. In terms of code quality, there's definitely a lot to be sought to still, but that's not too surprising as it's generally the case for this spaghettibase. There is a lot of clean code here, too, and the important part is going to be the functionality. Lunar, as I can see, touched up a lot of this to make it make functional(and cleaner even in places) and I think this is genuinely worth adding. She did quite a bit of testing, and after reading over it myself a couple times, I can't see much going wrong here aside from the unconvential verbing scheme. Overall approve this.

Razgriz1032:
EGG

PR #606, opened 2020-10-16
Reduce Volume size for vials.

MisterGrimm:
Vials are a mostly science/exploration, and sometimes medical beaker used to collect samples or hold small amounts of rare chemicals. This is just a simple fix that highlights that use, by reducing a 'balance' PR years back that changed all beakers to multiples of 30.

For all of the other beakers this makes perfect sense, as they are often used for mixing on chemistry machines. The vial however, is a different story. It is not intended to be used for mixing reasons, but rather small storage or sample collection. 

This PR changes the following :

- Reduces Volume size for Vials from 30 to 10, since it makes no sense to give them half the capacity of a beaker.

- Edit default transfer values, and transfer value options for vials. This is just so transfer works well with the new vial volume.

### DIFF
--- a/code/modules/reagents/reagent_containers/glass.dm
+++ b/code/modules/reagents/reagent_containers/glass.dm
@@ -240,10 +240,10 @@
 	icon_state = "vial"
 	center_of_mass = list("x" = 15,"y" = 9)
 	matter = list("glass" = 250)
-	volume = 30
+	volume = 10 //CHOMPedit reduced volume size, because its stupid to give a tiny little vial 30 volume
 	w_class = ITEMSIZE_TINY
-	amount_per_transfer_from_this = 10
-	possible_transfer_amounts = list(5,10,15,30)
+	amount_per_transfer_from_this = 5 //CHOMPedit edited default transfer amount so you don't dump everything in one go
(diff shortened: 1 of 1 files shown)


LunarFleet:
I have no authority on this, but, as a CMO main, I'd say aaaaaabsolutely not. Vials are not abused, or even used much to begin with, first off. There is no reasons in nerfing something just because 'realism'.

Vials are only consistently used for virology, because the antibody isolator does not accept beakers/bottles. Some people like to carry vials on them instead of bottles, usually if they're trying to use the hypospray mk2, which brings me to my next point.

Nerfing the vials like this will make the hypospray literally useless, literally. It is already fairly useless due to not being able to draw chems like a syringe can, but now, what you're suggesting, would make it unable to even administer a full 15u injection. 

I'm not taking part in, or going to reference, any of the discussion that happened with muffindrake in the ss13-development channel of the discord, because I believe everyone is allowed their own way of medical, as long as it is not outright detrimental to other players. 

No, simply making the vials smaller for no reason isn't really helpful, nor is it a needed change. It will simply make the hypospray completely useless. Keep in mind, droppers hold five units, _droppers_. What you're suggesting is that the vials only hold two droppers worth; there are many sizes of vials in the world, 30u is a good number for SS13.

MisterGrimm:
Closing due to significant controversial feedback.

PR #710, opened 2020-10-19
[MIRROR] Minor Update to ERT Gear

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/9205
--------------------
Corrects a terrible oversight in the equipment for the ERT ship in that I somehow *completely* forgot to issue the engineers with any insulated gloves as part of their 'basic' kit. You were fine if you used the hardsuit, but not if you use the mark 7 voidsuits. Oops.

Fixed by issuing six pairs to go with the six voidsuits. There were technically four pairs in the Tool Maker already, but those are easy to miss.

Not sure anyone's actually had time to notice since I don't think the ERT's been called yet, but better to fix it ahead of time.


Darlantanis:
Needs fixing because github HEAD thing. 

Submap we don't use, low priority.

PR #717, opened 2020-10-19
[MIRROR] [MIRROR] adds VR thunderdome to CentComm area

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/9229
--------------------
Original PR: https://github.com/PolarisSS13/Polaris/pull/7715
--------------------
Area currently not accessible, outside of adminery, since there's no VR Sleepers on the map anywhere.
There are plans to extend this area more to more stuff that isn't just deathmatch, but for now, this will do.
Also includes vendors to be used in said thunderdome.


Darlantanis:
One of these changes is southern_cross-6.dmm which is not something we merge from upstream (like it's not even the same map as ours). Is it possible to remove the changes just to that file?

Razgriz1032:
Mmh. I tried to just delete the file from the changed files tab, but it literally deleted all of it. I can just upload current map instead.

Darlantanis:
github/10

PR #728, opened 2020-10-19
[MIRROR] Adds Numbing Fangs trait

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/9173
--------------------
Adds the numbing fangs from "Sharp Melee & Numbing Fangs" as a separate trait.  Cannot be taken with "Sharp Melee" or "Sharp Melee & Numbing Fangs", but you can still take the latter if you want both.  Positive trait, costs 1 point.  (Didn't need to explicitly add them as excluded traits of each other, as there's code to automatically detect that they modify the same variable.  Tested to confirm.)


Shadow-Quill (approved):
Description is worded a bit odd, but that's an upstream issue.

Darlantanis:
yeah

PR #730, opened 2020-10-19
[MIRROR] Adds Virgo-Erigone, Virgo 3b and Virgo 4 as home/citizenship options

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/9195
--------------------
None of the citizenship/etc stuff had stuff that was specific to our setting in it, all of them stating home systems like "Sol" which are by definition a good distance away to travel for work. They're left in so you can choose 'em, but I added Virgo as a home system option for those of us that decide we live in-system where we work.


Darlantanis:
These are lore tweaks which are not suitable for our server. Given this is an upstream change, is it safe to just close the PR?

Darlantanis:
This is lore tweaks which does not fit our server.

PR #742, opened 2020-10-21
Blacklist Cold-Adapted & Hot-Adapted for Synths

r4d6:
Because of the way they work, Cold-Adapted and Heat-Adapted make it so that any synth that take it start receiving burn damage as soon as they spawn.

I don't know how to fix that, so I prefered preventing synths from using them with the already-built synth-blacklist. Anyway, synths can't be cold while they can overheat, so it even make sense IC.

The other bit of code commented out actually allowed people to take synth-blacklisted traits as a synth, for exemple, trying (and failing) to take the Cold-Adapted trait would allow you to take the Haemophilia trait as a synth.


cadyn (approved):
I helped them a bit with this and discussed it in the development channel. prefs.dirty_synth is a stupid variable which is set to 1 by a user.isSynthetic check and then set back to 0 after the user tries to take a trait not meant for synths until refreshed for some unholy reason. This is just commenting out the code that sets it to 0, and adding the not_for_synthetic flag to cold adapted and heat adapted, since synths handle temperatures differently and it's been shown to break shit.

r4d6:
> I helped them a bit with this and discussed it in the development channel. prefs.dirty_synth is a stupid variable which is set to 1 by a user.isSynthetic check and then set back to 0 after the user tries to take a trait not meant for synths until refreshed for some unholy reason. This is just commenting out the code that sets it to 0, and adding the not_for_synthetic flag to cold adapted and heat adapted, since synths handle temperatures differently and it's been shown to break shit.

So, I looked more closely at the code :
When it first run, it does a check and use it to see if you're a synth or not, upon which it change the value.
Right before the 'can synth take this trait', it does _another_ check, but this one doesn't change the variable at all.
Once the check is done the variable is reset to 0.

So basically it check if you're a synth and remember it, then it check again, then it refuse you, then believe you're not a synth and remember that and the second check does nothing.

Darlantanis:
I'd recommend sending this upstream but I'll merge it if you'd rather not to.

r4d6:
> I'd recommend sending this upstream but I'll merge it if you'd rather not to.

I just check, and Virgo doesn't seem to have any issue with it, but is also got [this pr](https://github.com/VOREStation/VOREStation/pull/8553) which apparently make the Cold-Adapted & Heat-Adapted Traits also change body temperature, which fix organics burning the instant they get on the station or in a space suit.

PR #743, opened 2020-10-22
Dehusking surgery

BlackMajor:
Webedit PR. Sue me.

This just makes it so that if you're doing burn repair on someone, it'll dehusk them.

Because having zero mechanical way to revive someone sans sleeving who's been husked is kinda lame.

### DIFF
--- a/code/modules/surgery/external_repair.dm
+++ b/code/modules/surgery/external_repair.dm
@@ -129,6 +129,12 @@
 	"<span class='notice'>You finish taping up [target]'s [affected] with \the [tool].</span>")
 		affected.createwound(BRUISE, 10)
 	affected.heal_damage(0, 25, 0, 0)
+	//CHOMPSTATION Edit start. Makes burn repair fix husking.
+	if(target.getFireLoss() <= 20) //Dehusk if no longer crispy
+		target.mutations.Remove(HUSK)
+		target.status_flags &= ~DISFIGURED
+		target.update_icons_body()
+	//CHOMPSTATION Edit end.
 	if(!(affected.burn_dam))
 		affected.burn_stage = 0
 	if(istype(tool, /obj/item/stack))


Shadow-Quill:
Scree, test your PRs 🗞️

cadyn:
Not gonna lie, not a huge fan of this. I'm open to having some trickery be possible for dehusking, but this just isn't it. Having doctors spend less than sane amounts of time manually patching up a body with surgery instead of resleeving (sometimes even when resleeving is directly requested) is already enough of a problem in my eyes and this just adds to it. If we're going to add a solution to husking, personally, I'd like to see it be its own thing, and probably have it heal a portion of the damage to give a headstart to the docs trying to save a body. Having burn kits magically de-husk corpses doesn't sit right with me.

LunarFleet:
I'm not really sure where you get '(sometimes even when resleeving is directly requested)', I have literally never had anyone ask me to just resleeve them. In fact, I've had times where people are resleeved, and, after asking IC, they say they'd like to be sleevemated back into their old body, if it can be fixed. 

 This is not as simple as "burn kit magic heal husking". You are having to perform a surgery to fix the damage before you can even remove husking.  Due to the fact that its <= 20 burn damage, this isn't even any slower than normal burn repair surgery, so, 'Having doctors spend less than sane amounts of time manually patching up a body with surgery' is kind of invalid, its not any longer. 

This is just a way to fix husking, which is otherwise unfixable. The husking is being caused by the burns, if you fix the burns, it'd make sense to fix the husking. If people don't want to go through the effort, resleeving is a problem. If people don't like doctors not resleeving them, that is up to the headmins to make a rule or decision about.

cadyn:
> I'm not really sure where you get '(sometimes even when resleeving is directly requested)', I have literally never had anyone ask me to just resleeve them. In fact, I've had times where people are resleeved, and, after asking IC, they say they'd like to be sleevemated back into their old body, if it can be fixed.
> 
> This is not as simple as "burn kit magic heal husking". You are having to perform a surgery to fix the damage before you can even remove husking. Due to the fact that its <= 20 burn damage, this isn't even any slower than normal burn repair surgery, so, 'Having doctors spend less than sane amounts of time manually patching up a body with surgery' is kind of invalid, its not any longer.
> 
> This is just a way to fix husking, which is otherwise unfixable. The husking is being caused by the burns, if you fix the burns, it'd make sense to fix the husking. If people don't want to go through the effort, resleeving is a problem. If people don't like doctors not resleeving them, that is up to the headmins to make a rule or decision about.

Let me elaborate. Right now husking serves as a barrier in that after x amount of damage, the doctor doesn't have the option of manually fixing the body anymore and is forced to resort to resleeving. I think this is a good thing, since I know doctors who would spend hours trying to fix a body rather than just resleeving and players who would find it very frustrating to be out of play for those extended periods of time. If we're going to remove husking as acting as a sort of barrier in that way, I'd like to see time being shaved off how long it takes to fix the body.

Insofar as husking goes, the way I see it is that the body is being damaged by burning to the point which conventional means are unable to recover it. Kind of like the difference between getting severe burns from a fire and being thrown into an incinerator. With one, we can heal them using our flesh mending surgery, with the other, there's no flesh left to be mended, and it follows that after extensive burn damage, much of the tissue and "information" so to speak would be destroyed in a manner not able to be fixed by flesh mending surgery.

My Idea is that rather than doing what is almost the functional equivalent of removing husking, we could have a unique procedure which fixes husking while also bringing the burn damage down so that it adds new mechanics for doctors to learn and experience while also solving the problem of players being forced to spend too much time out of play. I'm not saying that husking shouldn't be fixable, just that I don't think this is the right way to make husking fixable.

BlackMajor:
Bro you can repair and defib a body with damage capped brute damage, no organs sans the brain and no limbs, but can't defib a body if it gets specifically 200 burn, because that toggles husking, which initially existed as an aesthetic thing and shouldn't be a basis to revolve around for medical balancing.

Resleeving should only be a last resort when you've no body to work with.
Trying to validate the existence of husking preventing defibbing by relating it to "realism" is arbitrary at best.
This is a video game, and I am surprised you'd be willing to argue against enabling a doctor to do doctoring better. They can still resleeve if they're lazy, and the duration with which it takes them to repair the body is unphased.

All this PR does is negate the stone walling "Haha you have to resleeve no other way :^)" that husking encompasses.

Should probably add a way to repair toxins on corpses too, since atm that can't be healed either sans corpsemend.

LunarFleet:
That, again, is a rule problem. If people want to be resleeved, there should be a rule to respect LOOC requests. Using your logic, we could just remove external repair surgery in general because its slower than resleeving. We should remove franken surgery as well, as it resets the defib timer. 

You should not punish medical players with no way to fix a problem just because of an OOC issue. And damage is damage, if you heal the damage, its healed, don't need to come up with some incinerator justification.

Regardless, "My Idea is that rather than doing what is almost the functional equivalent of removing husking, we could have a unique procedure which fixes husking". Just like franken surgery. I dig it, _now make it happen, dev boi_ :dab:. If not, this is good, yes yes

cadyn:
> That, again, is a rule problem. If people want to be resleeved, there should be a rule to respect LOOC requests. Using your logic, we could just remove external repair surgery in general because its slower than resleeving. We should remove franken surgery as well, as it resets the defib timer.
> 
> You should not punish medical players with no way to fix a problem just because of an OOC issue. And damage is damage, if you heal the damage, its healed, don't need to come up with some incinerator justification.
> 
> Regardless, "My Idea is that rather than doing what is almost the functional equivalent of removing husking, we could have a unique procedure which fixes husking". Just like franken surgery. I dig it, _now make it happen, dev boi_ :dab:. If not, this is good, yes yes



> Bro you can repair and defib a body with damage capped brute damage, no organs sans the brain and no limbs, but can't defib a body if it gets specifically 200 burn, because that toggles husking, which initially existed as an aesthetic thing and shouldn't be a basis to revolve around for medical balancing.
> 
> Resleeving should only be a last resort when you've no body to work with.
> Trying to validate the existence of husking preventing defibbing by relating it to "realism" is arbitrary at best.
> This is a video game, and I am surprised you'd be willing to argue against enabling a doctor to do doctoring better. They can still resleeve if they're lazy, and the duration with which it takes them to repair the body is unphased.
> 
> All this PR does is negate the stone walling "Haha you have to resleeve no other way :^)" that husking encompasses.
> 
> Should probably add a way to repair toxins on corpses too, since atm that can't be healed either sans corpsemend.

Discussed this a bit with Lunar in DMs, gonna make a separate PR to properly implement a surgical procedure to dehusk, and potentially also one for mending toxin damage on corpses.

BlackMajor:
Bet

cadyn:
> Bet

#744 baby

BlackMajor:
This pleases me.

PR #744, opened 2020-10-22
Adding dehusk and tox/oxyloss surgery

cadyn:
Worked with Lunar for feedback on creating these new surgeries. Adds a new tool called the bioregenerator which can be used to dehusk and do tox/oxyloss surgery, with a nifty little sprite that's totally not a recolored advanced cataloguer with it's handle ripped off and a medical cross slapped on the side, nope, not at all. Gives medical the opportunity to be able to actually heal these cases whereas before they would be cockblocked and forced to resleeve.


Darlantanis:
This one is fine by me. With some of the text in this I can't help but wonder if somebody will try adding localized toxin damage in the future.

cadyn:
> This one is fine by me. With some of the text in this I can't help but wonder if somebody will try adding localized toxin damage in the future.

Illegal

PR #750, opened 2020-10-24
Commenting out micro mech stuff

MisterGrimm:
God what a dumpster fire.. 

- Commented out all micro mech stuff, including mechs, equipment, and build processes. Basically takes them off of the face of the earth.

- More of a sidenote, for some reason the mech medigun is in the micro mech stuff instead of living somewhere else. I don't know why its there, but I made sure it stayed uncommented.


Razgriz1032:
y tho

PR #754, opened 2020-10-26
Fruit TRUE/FALSE

Shadow-Quill:
because using `TRUE`/`FALSE` for bools is Good


cadyn:
using 1/0 is more space-efficient

cadyn:
`ShadowQuill.in_balls = 1`

Shadow-Quill:
> using 1/0 is more space-efficient
Coding standards.
We can use `#define tiles_per_tile 1`, but we don't because standards. We use `#define TILES_PER_TILE` instead to make it clear that it's a macro. same thing with TRUE/FALSE, takes a quick glance to tell that's it's actually a boolean.

Shadow-Quill:
> `ShadowQuill.in_balls = 1`

We'll talk later~

PR #757, opened 2020-10-26
Make the Synth Medichine be more like the organic version

r4d6:
If you compare the Organic Medichine to the Synth version, you would immediately see how much the synth version suck compared to the organic version, despite them costing the same price.

Here is what the Organic Medichine do : 
- Start healing as soon as under 80% health.
- Heal faster when under 20% health and notify the user to go get healed.
- When at -40% health, heal even faster and notify the medical department (Unless in a belly) and put the user in stasis until he is above 20% health.
- Heal Brute, Burn and Toxin at the same time at the rate of 0.1 each healed at the same time.

Here is what the Synthetic Medichine do : 
- Start healing as soon as there is damage.
- Heal 0.1 damage individually.
- Stop healing as soon as the damage is over 5 (Looks like total damage too). Which, from what I was told, is half the damage of a screwdriver stab.
- Does **not** warn Medical if the user is in critical.

As you can see, despite both of them costing 1250 Thalers, the synthetic version of the medichine clearly got the very short end of the stick. So in order to make it a little better for its cost, here is what this PR changed about the Synth Medichine : 
- Stop healing when over 30 combined damage, aka the threshold where the synth need someone else to fix them.
- Make it notify medical when under -40% health, but do not put them in stasis (Synths can't bleedout and I'm not sure if stasis protect against vacuum).

So, while it **does** still suck compared to its organic counterpart, it suck a lot less.

### DIFF
--- a/code/modules/nifsoft/software/05_health.dm
+++ b/code/modules/nifsoft/software/05_health.dm
@@ -134,9 +134,8 @@
 					if(!isbelly(H.loc)) //Not notified in case of vore, for gameplay purposes.
 						var/turf/T = get_turf(H)
 						var/obj/item/device/radio/headset/a = new /obj/item/device/radio/headset/heads/captain(null)
-						a.autosay("[H.real_name] has been put in emergency stasis, located at ([T.x],[T.y],[T.z])!", "[H.real_name]'s NIF", "Medical")
+						a.autosay("[H.real_name] is in critical condition, located at ([T.x],[T.y],[T.z])!", "[H.real_name]'s NIF", "Medical")
 						qdel(a)
-
 
 		return TRUE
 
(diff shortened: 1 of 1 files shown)


Razgriz1032:
I can support this. Lemme ask Nadyr again.

Darlantanis:
I'm a little unsure if this should notify medical or science (or both) about crit given this is for synths. Typically robotics would handle repair but on the other hand medical is the department which rescues dead/dying crew.

Darlantanis:
You also need CHOMpedit comments at each changed line or at the start and end of chunks of new/edited code. Otherwise your changes may be overwritten by upstream pulls.

r4d6:
@Darlantanis I added the tags.

PR #764, opened 2020-10-28
Yummy bugfixes

cadyn:
Fixes:

- Loneliness people detection issues
- Language number not properly updating
- Phoron bore can get infinite materials from 1 piece of phoron.
- Lights don't work with wearable circuits
- Putting more than 5 items into a candle pack results in missing sprite


Darlantanis:
Might be better to send the non-CHOMP specific bug fixes upstream.

cadyn:
> Might be better to send the non-CHOMP specific bug fixes upstream.

Probably is better, but I'm a bit pissed at virgo for some of the steaming shit they've sent downstream lately, and not feeling like waiting a week or whatever for a basic PR to get merged, so not really caring for sending these upstream atm tbh.

Darlantanis:
Relatable, though sending fixes upstream can prevent headaches from upstream conflicts later (or upstream implementing a "fix").

Razgriz1032:
Will the phoron bore changes affect or override or change nothing compared to #759

cadyn:
> Will the phoron bore changes affect or override or change nothing compared to #759

Bah, will need manual conflict resolution but it should override it since the stuff Virgo sent us still has the bug it seems

cadyn:
> Relatable, though sending fixes upstream can prevent headaches from upstream conflicts later (or upstream implementing a "fix").

Yea, and I get that, which is why I say that it generally is the better thing to do, but I feel justified with not sending this specific one upstream since most of these fixes are minor things that virgo may not even ever get around to fixing

PR #776, opened 2020-10-29
Disable living_ch

Razgriz1032:
This is causing a runtime since vore bellies have not been ported to TGUI, I think.

EDIT: I had to disable more stuff and comment out lines of code that was integrated in normal and _vr files.


cadyn:
113y6m

PR #780, opened 2020-10-30
Slip vore is upon us

cadyn:
Slipping into someone's tile now causes you to get vored by them or conversely depending on your spontaneous prefs. Enjoy.


r4d6:
Now the janitors will be the ultimate preds on this station.

PR #781, opened 2020-10-30
Starfire Report Update 104 2564

Darlantanis:
While the Starcaster has not yet been fixed the news is still accessible from the modular computers news program.


cadyn:
F in chat for starcaster.

PR #790, opened 2020-11-01
Map Tweaks

Shadow-Quill:
View mapdiffbot for changes, but:

ID Restoration terminal outside the HOP office (for re-adding access codes to your ID after an unfortunate trip into a vat of acid ~~or a pred's gut~~)

Air alarm in the OR Storage room (freezer preset)

And removing a duplicated firelock from the IAA office. (The game automatically deletes it while initing, but might as well make it do one thing less.


LuciusWorna:
How would the ID restoration terminal prevent someone who isn't the ID owner from restoring the ID?

Shadow-Quill:
> How would the ID restoration terminal prevent someone who isn't the ID owner from restoring the ID?

if realname != ID name, can't restore
```
else if((H.wear_mask && (H.wear_mask.flags_inv & HIDEFACE)) || (H.head && (H.head.flags_inv & HIDEFACE)))	//Face hiding bad
	to_chat(user, "<span class='warning'>Facial recognition scan failed due to physical obstructions. Access denied.</span>")
	flick(icon_fail, src)
	return
else if(H.get_face_name() == "Unknown" || !(H.real_name == inserted.registered_name))
	to_chat(user, "<span class='warning'>Facial recognition scan failed. Access denied.</span>")
	flick(icon_fail, src)
	return
```

PR #793, opened 2020-11-02
Add the Xenobotany ID Hud

r4d6:
This PR add a hud icon for xenobotanist, because I noticed they didn't have one.

Currently just a recoloration of the Botanist's, but someone more skilled can make another one.


Razgriz1032:
Upstream Polaris might like this if this works

PR #795, opened 2020-11-03
Makes door breakage for xenomorphs sound cool

MisterGrimm:
It should be noted that this only works with player xenomorphs.

- Forcing open welded and bolted doors as a xenomorph now causes a more violent sounding creak, complete with gears being forced to turn with hydraulics hissing from their torn tubes.

- Forcing open a door that isn't welded or bolted still has the old creaking sound effect.


code/game/machinery/doors/airlock.dm
```diff
@@ -94,7 +94,7 @@
 			else if(src.density)
 				visible_message("<span class='alium'>\The [user] begins forcing \the [src] open!</span>")
 				if(do_after(user, 5 SECONDS,src))
-					playsound(src, 'sound/machines/door/airlock_creaking.ogg', 100, 1, volume_channel = VOLUME_CHANNEL_DOORS)
+					playsound(src, 'sound/machines/door/airlock_creaking.ogg', 100, 1, volume_channel = VOLUME_CHANNEL_DOORS) //Normal airlock creak since we're not destroying the door
```
Shadow-Quill, line 97: prolly needs a chompstation comment, since new comment that doesn't exist in original file?

Shadow-Quill (approved):

code/game/machinery/doors/airlock.dm
cadyn: I would probably just delete this comment honestly

MisterGrimm: Ill nuke the comment then, I just put that there for understanding reasons.

cadyn: > Ill nuke the comment then, I just put that there for understanding reasons.

Yea, if this was a _ch file this comment would be totally good, and even encouraged, but the problem arises with the whole pulling from upstream thing

PR #802, opened 2020-11-05
Give the Roboticist access to the rest of Science.

r4d6:
This PR give the roboticist job access to Xenobiology and Xenobotany. Mainly for two reason : 
1. So the roboticist don't have to beg for a plant analyzer each time he's asked to make a farmbot.
2. Literally every other job in Science has access to everywhere in Science (except the RD's office, of course), so Roboticists feel a bit left out, being unable to access those locations while those working there can freely access robotics.

### DIFF
--- a/code/game/jobs/job/science.dm
+++ b/code/game/jobs/job/science.dm
@@ -129,7 +129,7 @@ VR edit end*/
 	supervisors = "the Research Director"
 	selection_color = "#633D63"
 	economic_modifier = 5
-	access = list(access_robotics, access_tox, access_tox_storage, access_tech_storage, access_morgue, access_research) //As a job that handles so many corpses, it makes sense for them to have morgue access.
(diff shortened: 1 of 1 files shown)


Razgriz1032:
I do not support.

If anything xenobio needs robotics access removed.

r4d6:
> I do not support.
> 
> If anything xenobio needs robotics access removed.

All I know is that all science jobs got access to robotics, but robotics does not have access everywhere.

Kazzc:
I will say yes, giving them access to the rest of science will mean easier ways to get tools, resources and assisting other scientist in the department.

But just cause they have access they are not expected to do the other areas in the department.  As-long as they don't have access to more restricted areas like the RD. Not to mention will make a bit less of a headache playing science when the round has few crew.

Darlantanis:
I am not in favor, mostly because the meta of what a job is expected to do is ultimately determined by access. I recall an issue with psychologists gaining increased access to medical resulting in the expectation of them performing standard medical treatment despite the job itself being meant for roleplay.

Shadow-Quill:
I could see this getting added onto lowpop access, but don't like roboticists having more access in normal cases.

PR #806, opened 2020-11-05
Merge all of Science's jobs in one job with 10 titles.

r4d6:
While getting suggestions and feedback on https://github.com/CHOMPStation2/CHOMPStation2/pull/802 on discord, someone suggested that if I give all the jobs in science equal access to everywhere in Research, might as well just merge them all in one job with 10+ alt-titles.

And honestly, I feel like it is a better idea than just giving roboticists access to Xenobotany. So I did it.
Here is the list of changes in a bit more details : 

- Commented out all of the science jobs (except RD) in science.dm & science_vr.dm in order to prevent upstream merge conflict.
- Made a new job in the newly made file science_ch.dm, which has all the access of the other jobs.
- Gave said job alt-titles that previously were complete jobs in Science.
- Blob code now look for the three roboticist alt-title instead of the Roboticist job.
- Change the spawning landmarks to only spawn Scientists & their alt-titles instead of spawning roboticists, xenobiologists or xenobotanist.


Shadow-Quill:
As said in the discord chat, I don't agree with this PR at all.

LunarFleet:
Fam, if the first PR didn't go well, why would making a more extreme PR to do the same thing go better?

PR #808, opened 2020-11-06
Add the protector energy gun to loadout

r4d6:
This PR add the Small Energy Gun 'Protector' to the loadout.

Taken from the game : 

> The WT-98a 'Protector' is a common sidearm developed by Ward-Takahashi GMC. It features a powerful stun mode, and an alert-level-locked lethal mode, only usable when the connected jurisdiction allows.

It is a great Self-Defense gun that can only fire a few shots, 3 lethal and 6 stunning with a fully charged Advanced Device Power Cell
It can't fire lethal shots if the alert level is at Green.
It cost 5 points in the loadout (can be changed if people feel it need to cost even more).
It actually spawn a variation that does not give any tech level, so people can't get the gun just to speed research.
Can only be taken by the Site Manager and Security.


Darlantanis:
Yeah guns in the loadout is a hard no.

PR #810, opened 2020-11-07
Allow Robotics to make a cleanbot and a farmbot from the start.

r4d6:
This PR replace two health analyzer and empty medkit normally (never) used for medbots by the stuff needed to make one farmbot and one cleanbot, minus the arms.

Here's the entire list of additions : 
- Watertank
- Plant Analyzer
- MiniHoe
- 2x Bucket
- 1 Proximity Sensor


MisterGrimm:
For one, medbots are used and made alot in robotics. A normal roboticist makes at least two of them at start, one for medical, and one for robotics for prosthetic augmentation purposes. 

Second, you can make buckets, proximity sensors, and plant analyzers in either science, or flat out in robotics autolathe. The only items you actually need from the botanist is a mini hoe, and a water tank. It does not make much sense putting either of those in robotics, when they are usually acquired when a botanist asks you to make a farmbot.

LunarFleet:
Medibots are the most used bot in the game, I'm not sure where you think that they are unused, I consistently see one in the medbay every shift, and usually one in the bar. Just print the things you need for a farm bot, or ask for it.

PR #819, opened 2020-11-08
[MIRROR] Event vendors and entertainer's vendor

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/9286
--------------------
Not really 'new' vendors, so much as adds expanded loadout vendors for use in events, though, there are new food vendors that have just about every kind of food in them. None of these are ever to be mapped in on normal maps. These are sort of a first pass. I confirmed that they work but I want to expand what they offer later.

Also adds an entertainer costume vendor, which is mostly the same as the normal one, but without the costs for costumes, hidden in the entertainer's back room. The main other difference is that it doesn't have the premium/contraband items in it, you still gotta go to the other vendor for that.


code/game/machinery/vending_machines_vr.dm
```diff
@@ -1334,3 +1334,3114 @@
...
+					/obj/item/clothing/accessory/poncho = 5,
+					/obj/item/clothing/accessory/poncho/green = 5,
+					/obj/item/clothing/accessory/poncho/red = 5,
+					/obj/item/clothing/accessory/poncho/purple = 5,
+					/obj/item/clothing/accessory/poncho/blue = 5,
+					/obj/item/clothing/suit/jacket/puffer = 5,
+					/obj/item/clothing/suit/jacket/puffer/vest = 5,
+					/obj/item/clothing/suit/storage/flannel/red = 5,
+					/obj/item/clothing/suit/unathi/robe = 5,
+					/obj/item/clothing/suit/storage/snowsuit = 5,
```
Shadow-Quill, line 3333: Invalid filepath

Shadow-Quill (changes requested):

PR #826, opened 2020-11-10
Allow Phase Weaponry to be printed again.

r4d6:
This pr simply allow R&D's protolathe to print the phase weaponry, which do 5 damages on everything that aren't animals simple mobs, while dealing more damage to them, making it useful for self-defense against wild-life and other mobs.

They all deal 5 damage by default, but when hitting simple mobs : 
The Pistol do 40
The Carbine do 50
The Rifle do 60
The Cannon do 75


Razgriz1032:
I am ok with this.

PR #829, opened 2020-11-11
Cadyn's traits v2

cadyn:
Traits added:

- Extreme slowdown (-8) (four times slowdown from major slowdown)
- Permanent Blindness (-8)
- Extremely low blood volume (-5)
- Low blood volume (-3)
- Schizophrenia (-3)
- Agoraphobia (-3)
- Scrawny (-2)
- Low Blood Sugar (-2)
- Deep Sleeper (-1)
- Shrinking (0)
- Mudking (0)
- Singularity Metabolism (0)
- Big mouth (1)
- Radiation Resistance (2)
- Heavyweight (2)
- Table Passer (2)
- Absorbent (2)
- More blood (3)
- Grappling expert (3)
- Giant mouth (3)
- Extreme radiation resistance (4)
- More blood extreme (6)
- Adrenaline rush (6)
Damn, that's a lot of traits. Thank god for the special_env feature I added for traits *blep*

Also added Tercozolam, a medication specifically for lessening the effects of schizophrenia.
Made with 1 part dylovene, 1 part ethanol, 1 part lithium

### DIFF
--- a/code/game/turfs/simulated.dm
+++ b/code/game/turfs/simulated.dm
@@ -93,7 +93,12 @@
 
 		if(istype(M, /mob/living/carbon/human))
 			var/mob/living/carbon/human/H = M
-			dirtslip = H.species.dirtslip	//CHOMPEdit
+			//CHOMPEdit Begin
+			dirtslip = H.species.dirtslip
+			if(H.species.mudking)
+				dirt = min(dirt+2, 101)
+				update_dirt()
+			//CHOMPEdit End
 			// Tracking blood
 			var/list/bloodDNA = null
 			var/bloodcolor=""
--- a/code/modules/food/food/snacks.dm
+++ b/code/modules/food/food/snacks.dm
@@ -143,8 +143,14 @@
 		if(reagents)								//Handle ingestion of the reagent.
 			playsound(M,'sound/items/eatfood.ogg', rand(10,50), 1)
 			if(reagents.total_volume)
-				if(reagents.total_volume > bitesize)
-					reagents.trans_to_mob(M, bitesize, CHEM_INGEST)
+				//CHOMPStation Edit Begin
+				var/bite_mod = 1
+				var/mob/living/carbon/human/H = M
+				if(istype(H))
+					bite_mod = H.species.bite_mod
+				if(reagents.total_volume > bitesize*bite_mod)
+					reagents.trans_to_mob(M, bitesize*bite_mod, CHEM_INGEST)
+				//CHOMPStation Edit End
 				else
 					reagents.trans_to_mob(M, reagents.total_volume, CHEM_INGEST)
 				bitecount++
--- a/code/modules/mob/living/carbon/human/human_organs.dm
+++ b/code/modules/mob/living/carbon/human/human_organs.dm
@@ -146,12 +146,12 @@
 	// Check again...
 	if(!l_hand && !r_hand)
 		return
-
+	var/adrenaline = has_modifier_of_type(/datum/modifier/adrenaline)	//CHOMPEdit
 	for (var/obj/item/organ/external/E in organs)
 		if(!E || !E.can_grasp)
 			continue
 
-		if((E.is_broken() || E.is_dislocated()) && !E.splinted)
+		if((E.is_broken() || E.is_dislocated()) && !E.splinted && !adrenaline)	//CHOMPEdit
 			switch(E.body_part)
 				if(HAND_LEFT, ARM_LEFT)
 					if(!l_hand)
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -969,6 +969,9 @@
 		if(nutrition > 1000 && species.grows && size_multiplier < RESIZE_HUGE)
 			nutrition_reduction *= 5
 			resize(min(size_multiplier+0.004,RESIZE_HUGE))
+		if(nutrition < 200 && species.shrinks && size_multiplier > RESIZE_TINY)
+			nutrition_reduction *= 0.3
+			resize(max(size_multiplier-0.004,RESIZE_TINY))
 		//CHOMPEdit End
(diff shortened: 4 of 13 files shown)


cadyn:
Me and Sempre also tested these a good bit yesterday, tweaking Adrenaline rush around in particular until we got some satisfactory results. This should be ready to merge whenever, really.

Darlantanis:
Descriptions on these would make it a lot easier to evaluate.

cadyn:
> Descriptions on these would make it a lot easier to evaluate.

There's so many traits though ;o;

cadyn:
> * Extreme slowdown (-8) (four times slowdown from major slowdown)
Self-explanatory
> * Permanent Blindness (-8)
You are now blind, and medication can't cure that. Sorry m8
> * Extremely low blood volume (-5)
40% blood volume of most species
> * Low blood volume (-3)
66% blood volume of most species
> * Schizophrenia (-3)
Has episodes of hallucinations every 5 minutes on average which last ~7 minutes on average with some spicy gaussian randomness. If you take your meds, it changes to once every 50 minutes on average and lasting ~4 minutes on average, with similarly spicy gaussian randomness.
> * Agoraphobia (-3)
More than two people around you? Start losing your shit, similar to loneliness trait.
> * Scrawny (-2)
Harder to secure/hold grabs, and harder to escape from them.
> * Low Blood Sugar (-2)
Not eating enough food? Start to experience some adverse effects including confusion, lightheadedness, and passing out.
> * Deep Sleeper (-1)
Twice as long to wake up from anything which renders you unconscious
> * Shrinking (0)
Shrinks without enough nutrition
> * Mudking (0)
Makes tiles dirty faster
> * Singularity Metabolism (0)
2x reagent metabolism speed, 32x hunger rate. (is it worth it?)
> * Big mouth (1)
Half as many bites to eat any food item
> * Radiation Resistance (2)
Radiation dissipates from your body faster, you are able to withstand more radiation before it does damage, and radiation does slightly less damage
> * Heavyweight (2)
Nobody can push you or swap tiles with you. Can still be pulled, though.
> * Table Passer (2)
Can move across tables similar to a teshari.
> * Absorbent (2)
Absorbs dirt and blood on tiles similar to a promethean.
> * More blood (3)
50% more blood than most species.
> * Grappling expert (3)
You have an easier time escaping from grabs, and other people have a harder time escaping from yours.
> * Giant mouth (3)
Takes a quarter as many bites to eat most things.
> * Extreme radiation resistance (4)
Radiation resistance but crank up the numbers a bit.
> * More blood extreme (6)
150% more blood. Juicy.
> * Adrenaline rush (6)
When you get below you 0 health, you get a boost of adrenaline which increases your speed, attack abilities, brings you up from the ground, causes you to ignore all pain/stun/disabling effects, so that you can get yourself out of a sticky situation. Lasts 30 seconds before being replaced by "adrenaline detox" which lasts for ~9 minutes, and causes slowed pulse, slower metabolism, slightly increased oxy damage, great slowdown, increases duration of disabling effects, and usually leads to you being disabled.

Razgriz1032:
Dragonshark had an idea from log ago that I couldn't figure out - Narcolepsy. A -2 pt trait that puts you to sleep for a few seconds every so often.

Razgriz1032:
I'd consider Adrenaline rush as like a 9 point trait. I am also going to laugh when someone takes the extreme slowdown trait, then when he's about to die, starts running around at ludicrous speed.

cadyn:
> I'd consider Adrenaline rush as like a 9 point trait. I am also going to laugh when someone takes the extreme slowdown trait, then when he's about to die, starts running around at ludicrous speed.

Hmm, I did try to balance it out with the adrenaline detox, and the fact that you can't have another adrenaline rush for like 30 minutes I think it was... 9 does seem a bit excessive, considering this likely won't even save you from death, just giving 30 seconds to **maybe** deal with a single enemy if you're brave(it's still pretty easy to during an adrenaline rush), and then gtfo to safety.

Darlantanis:
Alright, feedback time. If I don't bring up a trait assume I'm okay with it.

Permanent Blindness
One of the effects of blindness is to prevent players from seeing "Me" posts. While I understand the logic behind this (you can't see anything hahah) it doesn't make much sense in a roleplaying server where "Me" posts are overwhelmingly used for roleplay (i.e. not mechanics) and describe many senses other than just sight. It would probably be prudent to tweak this before offering blindness as a trait (otherwise we're going to get a lot of players never realizing they're missing posts). 
If you ever get inspired to make a permanent deafness trait I'll note that deafness prevents players from receiving subtles. 

Schizophrenia
Trait itself is fine but it's a bit of an oversimplification of schizophrenia (which makes me wonder how a schizophrenic would feel about it if we represent the condition as periodic hallucinations). Perhaps "Episodic Hallucinations" would be a better name? Additionally, unless I'm missing something there's nothing (outside of code) actually informing players which med is needed to reduce hallucination episodes.

Grappling Expert
Could see this one becoming meta. I may recommend a cost increase if I see it abused.

Adrenaline Rush
This is a _really_ potent modifier even if just for thirty seconds, sort of to the level where I could see it reliably turning the tide of PvP combat (not that we often have PvP combat). I can see this getting nerfed hard depending on how players use/abuse it.
I also see it tweaks bleeding_rate_percent. Will this cause it to cure haemophilia for anybody with that trait/disability?

Side thoughts, would any of the negative traits fit in the disabilities tab for non-custom species? It's been on my mind since haemophilia was added there as it can be interesting to voluntarily weaken characters.

cadyn:
Adrenaline Rush is obviously the biggest question mark here, so will be keeping an eye on it, just added a call to log and message admins every time the ability is activated, so should help us keep track and take notes on how it's used.

Added a warning to the blindness description. Todo, potentially add deafness, and to fix the stupid code that says blind people can't see Me's right in from them.

Schizophrenia renamed to episodic hallucinations, bc it makes what the user is signing up for a lot more clear anyways and the feedback given was very valid. Wish I thought of this myself :v

PR #830, opened 2020-11-11
Added spicy tools

cadyn:
There are now spicy tools. Why do we need spicy tools? Yes. What do they contribute to the game? Yes. Spicy tools.

No current way to obtain them other than admin spawn, so mostly for admins or potentially POIs or events in the future :p


cadyn:
fuck you travis

PR #835, opened 2020-11-13
Actually give CMO teleport access.

cadyn:
no this can't be in the medical_ch file because of course medical_vr compiles after _ch :seensomeshit: if only we were zhomp station instead.


Razgriz1032:
Thanks obama

cadyn:
> Thanks obama

Fucking obamacare man.

PR #836, opened 2020-11-13
New, improved, reflector turrets

cadyn:
New improved reflector turrets added. They use a subsystem and some big brain tactics to make them capable of handling big-boy level shit and not lagging the server or breaking. Can be built or ordered from cargo. Cargo version cannot be deconstructed.

Also added a secret engineering area with a few treats for any spicy engineers. Has 10 Gyrotrons, 5 Gyrotron Computer circuits, 5 Gyrotron circuits, 5 Reflector boxes, a canister of 2.7k phoron, and a canister of 100mPa oxygen.


cadyn:
image

LuciusWorna:
The secret room seems a bit much - the number of gyrotrons is pretty high and this map change is for one or two people only due to the content. If people want the ability to change the configuration of the engine as a map default, add a second hard storage in the same construction theme as the first.

cadyn:
> The secret room seems a bit much - the number of gyrotrons is pretty high and this map change is for one or two people only due to the content. If people want the ability to change the configuration of the engine as a map default, add a second hard storage in the same construction theme as the first.

This is not another engineering hard storage, it's a secret room intentionally. Mostly because I'm worried about newbie engineers just stumbling across it and going "oH wOndeR wHaT all ThIs is For", this way at least, when it goes shown to people they will hopefully have someone that knows what they're doing tagging along to explain things. Also the I could totally see more gyros than that being needed or wanted for specific engine setups. The purpose of this room is to encourage a different type of playstyle regarding engineering that encourages exploring different setups, gaining a better understand for how things actually work, and yes part of it is for "engineer posting" as it's called.

If I wanted this to be for just one or two people, as you say, well I wouldn't make a unique room for it then. This is something meant to be shared. Because I want people to learn not to fear the SM engine, but to understand it. 

> If it is to remain secret, remove the APC and air vents as well as the secret door, replacing the reinforced walls with regular ones like every other secret room on the map.

It's not "any other secret" on map. The and wires leading up to it actually blend in rather will with the whole thing going on in atmospherics unless you're actually aware of the piping, and what it all does and where it should be, in which case you're probably at the least more than a newbie engineer, so I don't mind you discovering this too much.

cadyn:
> > If it is to remain secret, remove the APC and air vents as well as the secret door, replacing the reinforced walls with regular ones like every other secret room on the map.
> 
> It's not "any other secret" on map. The and wires leading up to it actually blend in rather will with the whole thing going on in atmospherics unless you're actually aware of the piping, and what it all does and where it should be, in which case you're probably at the least more than a newbie engineer, so I don't mind you discovering this too much.

Actually adding onto this further, the reason most secret rooms don't have APCs or air alarms is because they're part of maints, and this room again, doesn't fit into that same category, and vents, scrubbers, an APC, and an air alarm could be needed especially if the room is in active use.

cadyn:
I am going to make this because obviously this is more of a controversial PR than I anticipated. Here is a quick FAQ/concern answering:

Why does it have to be a secret room?:
It doesn't have to be a secret room, but the intention behind making it a secret room is multi-pronged. First of all, newbie engineers are kept out for the most part, because even if they ghost and see it, not many of those items are worthwile to them since they have no idea what they do. The second part is that players are meant to be "introduced" to the room by another play who is familiar with at least some of those items purposes, and how to use them. This way, hopefully we can introduce understanding into the playerbase without them just assuming things and inevitably incorrect information becoming "common knowledge" among players. 

Why does engineering need these things anyways?:
Engineering doesn't need these things. What engineering needs is something more than the typical playstyle of set the same engine up the same way every time, and do the same things then just go do whatever and fix things when they break. This is attempting to introduce a new playstyle that involves experimenting, trying to improve engine setups, and gaining true understanding for how things work. This is my best idea for implementing an approachable solution to that, which will hopefully be spread through playerbase by our engineers who are most familiar with this type of play.

Why these parts in specific? Why this many?:
Options. The answer is options. People will take as many as they need, and this can always be tweaked in the future to decrease the amount of certain items and potentially replace them with other new items. At the moment though, I want players to be able to either use one extra gyrotron, or go big and make an engine that makes use of 9 gyrotrons with reflectors. 100mPa canister of oxygen? Meant to be used for an oxygen engine setup to get past having to wait on like 7 oxygen canisters to fill to 15mPA. 2.7k Phoron Canister? Meant to be used for radiation collectors and potentially pumped into the engine cooling loop to compensate for an oxygen hot loop, since the gas will expand over time as the engine gets hot anyways. These help the engineers do these changes without spending 3-6 hours of a shift on them waiting for various factors. Also helps be a platform for again, new players trying to learn more advanced engineering.



In the end it's my personal belief that this is one of those things that people will "look a gift horse in the mouth" because they're scared of change. If this really ends up a problem after this gets merged, we can always just go back and undo the map changes in specific, but I've done some good thinking about this and I really think it's just the boost that engineering needs.

LuciusWorna:
Not seeing where the controversy is at unless it is happening in a chat I don't have access to.

My point remains is that this change is highly specific to few engineers. You want to give us more options? Include the circuit boards for a R-UST, singulo or tesla as well as the components required. Don't include finished parts and twenty gyros is too much outside of a warehouse at central or already installed into the engine.

I highly disagree with keeping newbies out as, if they walk through atmospherics, they will see the room because mesons at a part of engine setup.

Again, make it a second hard storage. One with a vault door that requires CE access like secure technical storage. In-universe, why would NT leave this stuff unsecured?

Kazzc:
Map wise this is very counter intuitive to have a ``secret room`` with a ludicrous amount of expensive equipment. 10 Gyrotrons is just outright excessive,
Particle emitters can already be abused by antags and can be e-magged to be a nasty weapon. Gyrotrons not only shoot faster, have longer range, do more damage and charge things faster. 
Not to mention that this defeats the purpose of having cargo... if any department needs additional equipment they should be ordering it from cargo to get those 10 Gyrotrons, 
I am not against the idea of loading in extra equipment for engineering to experiment but mapping it in to a ``secret`` room is very poor design for the station as a whole. If you want additional protection from it being abused then maybe only the CE (Chief of Engineering) should have easy access to the room that contains such items.
I will not mind implementing this but you have to massively reduced the amount of expensive equipment you get out of the bat, and have it mapped in engineering in a more cohesive location.

image

I can see using these two empty rooms as additional engineering storage to place such Gyrotrons... I will have to move things around to make it look right.

Darlantanis:
Canonically durasteel would be a perfect material for a reflector (though I guess it's a bit more annoying to get than diamond). 

10 gyrotrons is a bit much to stick in secret room on station. Considering they're essentially upgraded emitters that's a lot of firepower for a secret room. Kinda unsure about the gas canisters, I could see it becoming meta for toxins to run off with both. 

The room itself is somewhere between normal and secret which is inconsistent with the station mapping standards; it needs to be aesthetically set up like another secret room (maintenance flooring, no APC/air alarm/etc.) or like a proper department room. I am going to direct you to Kassc for mapping as he will go much better in-depth about it than I can (plus he's got a better eye for design).

Your intention for the room seems to be offering options for easily pushing the engine to further extremes. I wonder if this might be better handled with a plains POI or perhaps a static, derelict engine structure near the Sif outpost. That would admittedly be less visible than a room like this... but on the other hand it seems unlikely many engineers will see this room and understand its intent when extreme engine modification is a small niche.

cadyn:
So I have removed all of the mapping changes from this PR. After reading your responses, I've decided that such changes are likely meant for a different PR anywho. In the meantime, though, the least we can have is working reflector turrets :p

PR #844, opened 2020-11-14
Current Metroid Progress

Razgriz1032:
This is the current progress of metroids. Just updating on master for now (because it somehow snuck in when we were setting up the mirror bot). They WILL be enabled in game, but please do not spawn them.

Current (and biggest goal) is to get the gamma, zeta, omega, and queen to move and shoot. Gamma is what I have been using to test. Current behavior is that it sees target, takes a long time to shoot and waits in place while doing so, performs the shot, then begins rushing the target, but then stops as it begins readying another shot.

Last goal is just to refine and verify projectiles for the above listed.


Razgriz1032:
why

PR #849, opened 2020-11-15
TGUI changes from upstream are a nightmare now.

cadyn:
Thanks @Darlantanis wow you just had to make starcaster in tgui huh


Shadow-Quill:
those mirror prs that are currently conflicting should just need a bundle rebuild

cadyn:
> those mirror prs that are currently conflicting should just need a bundle rebuild

that's what I did, but doing it for a PR you don't own is painful, git GUIs don't like it

PR #860, opened 2020-11-18
Unfuck security access

Razgriz1032:
Loosens the :choke: on HoS access that Virgo had. They now have Polaris access, which should include most hallways of departments. If you catch the HoS anywhere and they're not currently pursuing someone or conducting security business, that's tresspassing, a regulation violation.

Grants the warden access to the red armory.


cadyn:
seems legit enough

PR #862, opened 2020-11-18
Make the web bindings a subtype of the straight jacket

r4d6:
Right now they are just like any normal clothing, you can put them on and take them off without any problem.

This PR make them a subtype of the straight jackets, so that they inherit their effects, and make them actually able to restrain people.

### DIFF
--- a/code/modules/vore/appearance/spider_taur_powers_vr.dm
+++ b/code/modules/vore/appearance/spider_taur_powers_vr.dm
@@ -1,4 +1,5 @@
-/obj/item/clothing/suit/web_bindings/
+// CHOMPEDIT : Make the web bindings a subtype of the straight jacket
+/obj/item/clothing/suit/straight_jacket/web/
 	icon = 'icons/vore/custom_clothes_vr.dmi'
 	icon_override = 'icons/vore/custom_clothes_vr.dmi'
 	name = "web bindings"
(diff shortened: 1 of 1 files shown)


cadyn:
Proper Chompedit tags please. They should be on the same line, not the line before, and every line that has an edit needs one.
`
edited.code = edited.value  //CHOMPEdit
if(nonedited.boolean)
       nonedited.do_thing()
else
       edited.do_thing()   //CHOMPEdit
`

Razgriz1032:
I don't like that it's a straight jacket because you cannot escape straight jackets on your own. It might be better as reskinned legcuffs. Also I'm not gonna like the fact that it'll mean spider people will become legcuff printers, but it might be a risk worth taking.

r4d6:
> I don't like that it's a straight jacket because you cannot escape straight jackets on your own. It might be better as reskinned legcuffs. Also I'm not gonna like the fact that it'll mean spider people will become legcuff printers, but it might be a risk worth taking. Admins will discuss.

I mean, from the code that I can see [here](https://github.com/CHOMPStation2/CHOMPStation2/blob/master/code/modules/clothing/suits/miscellaneous.dm#L243), you can resist out of them, you can't just take them off, but you can resist out of them, even if you can't take it off like every other suit slot clothing.

Darlantanis:
This one is a hard no from me. Web bindings originally worked as straitjackets but were nerfed to roleplaying tools because it turns out it's really OP for a species to be able to spit out straitjackets on demand. There's not a need for them to have mechanical effects (though on that note I could see it interesting if resisting generated flavor messages of some sort despite being able to just equip/unequip whenever).

Unless there have been recent upstream tweaks, straitjackets function as an inescapable hand/legcuffs combo which prevent pretty much everything but slowly walking. You can resist out of a buckled chair in a straitjacket though it takes time.

PR #872, opened 2020-11-19
ranged_cooldown_time variable for mobs

Razgriz1032:
This is the addition of a new mob variable that will allow mobs to move and shoot at the same time. This section here is special snowflake code for metroids only, or for whatever else in the future that you want to have move and shoot at the same time. Basically, this is a non-stupid version of "ranged_attack_delay." "ranged_attack_delay" is stupid because it sleeps the entire mob. This new ranged_cooldown_time is smarter in the sense that it is an internalized timer. Try not to confuse the names.

### DIFF
--- a/code/modules/mob/living/simple_mob/combat.dm
+++ b/code/modules/mob/living/simple_mob/combat.dm
@@ -98,6 +98,16 @@
 		if(reload_count >= reload_max)
 			try_reload()
 			return FALSE
+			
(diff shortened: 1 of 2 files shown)


cadyn:
I'm thinking maybe you should just override this proc for metroids. Biggest concern is that this will be adding this code and running it for every single simple mob that shoots any projectile, even if it's just skipping over the code after the if statement, that's still an extra branch for the CPU and could add some unwanted slowdown.

Razgriz1032:
No because I want to add this for taser spiders, come to think of it. Taser spiders was the reason I made this in the first place on old chomp

Razgriz1032:
Also we can apply this to xenomorphs since the ones that shoot projectiles stand still and don't do anything unique and can be cheesed easily. Might be better to change their AI to use more unique tactics as well while not spamming people with their attacks.

Razgriz1032:
Merging because this has further applications than metroids.

PR #874, opened 2020-11-19
Fixing metroids

cadyn:
Metroids can now run and shoot, and they don't spit out runtimes anymore :v This should be better than adding the changes to the low-level simple_mob shooting code.


Razgriz1032:
But what about taser spiders

Razgriz1032:
All these values are wrong from my final intent in https://github.com/CHOMPStation2/CHOMPStation2/pull/878

But nutrition is a good.

Razgriz1032:
Your nutrition fix has been merged into #878

PR #902, opened 2020-11-23
Finally optimize radios

cadyn:
Finally got around to this! Should bring some major performance boosts for us.
Oh yeah also fixed a bunch of markings for us since apparently at some point our _vr file became not so _vr and actually different from virgo.


Shadow-Quill:
Could upstream use this?

cadyn:
> Could upstream use this?

Upstream can come and get it if they want it :v

PR #911, opened 2020-11-24
Spear and Gun sizes

Razgriz1032:
-Made all gun and spear sizes larger so that they will no longer fit into backpacks.
--Energy spears and energy swords will fit while deactivated, but not while active (unchanged)

-Tweaked the Floragun have no recoil for micros.


Razgriz1032:
wtfffff where are the changes

PR #917, opened 2020-11-24
Change frontier phaser beam-type to ACTUALLY BE a phaser

Razgriz1032:
Because FUN FACT: Phasers were actually using regular beams.

After looking through the code during my gun and spear size PR, I made the realization that there are two types of phasers, but we were all under the impression that they only did 5 damage to people and full damage to mobs. This is true of phasers. However the Frontier Phasers are a Virgo weapon and apparently are not phasers at all, they are lasers, and do indeed use lasers, NOT phasers. So I am fixing that.

Pros of change:
-I created a new beam type for the phaser beam. The OG Phaser uses a short-ranged, slow-moving projectile. So I made that, but made it look like the default laser beam. I also copied the damage numbers, so you do lots more damage to mobs of MOB_CLASS = ANIMAL now. 5 burn damage, depending on armor, plus a straight-up FLAT BOOST to damage, afaics.
-You won't friendly fire people as badly because people were under the impression that these only did 5 damage to people.

Cons:
-No cons anymore after class updates


cadyn:
please god no. Phasers have been this way since forever.

Razgriz1032:
Beam has been tweaked to now do damage against synthetics

Razgriz1032:
And xenos

Razgriz1032:
Ok, now it will kill mercs, animals, mechs, and drones.

r4d6:
> Ok, now it will kill mercs, animals, mechs, and drones.

1. MOB_CLASS_SYNTH include FBP and everything that pass the is_synth() check.
2. At this point just set SA_vulnerability to null to include every simple mob, but not player.

Razgriz1032:
ugg

cadyn:
Make sure to test this before merging it to see what the bonus actually works on.

cadyn:
Well, looks like I'm gonna be testing this since I decided that these needed to be converted to orange beams and set myself out to do it

cadyn:
Tested this and it's golden, so I'm merging 👌

PR #923, opened 2020-11-25
Making radios even more optimized

cadyn:
had a big :brain: moment and realized that we can make this even faster by not calculating intercoms' broadcast tiles every time someone uses a radio, since they literally can't move at all except with shuttles, and just pray that whatever niche case that might move them involves using forceMove 👍


cadyn:
Git CI only takes 7 minutes vs like 15 minutes from travis why have we not had this sooner aa

cadyn:
> Git CI only takes 7 minutes vs like 15 minutes from travis why have we not had this sooner aa

Should probably wait until we find out whether or not the machines controller change fixed it so we can know which one fixes the lag if any because shotgunning variables is maybe a bad idea

PR #974, opened 2020-12-06
Liquid Bellies TGUI Port

### LINKED ISSUE #889: Nadyr TODO: Reimplement liquid vore bellies
Fix liquid vore bellies with tgui

Darlantanis:
Ports liquid bellies to tgui. Also fixes issues with reagent mode addons (probably) which apparently broke months back due to refactors to digest modes.
Increases the complexity limit of lint to let tgui compile, I've been at this for around 33 hours so reducing the complexity of VorePanel.js is quite a bit beyond me right now.

The ui stuff itself could use some better design but as I'm tired and not super adept with ui visual aesthetics that will have to come later. Combining the "Liquid Options" and "Liquid Messages" tabs would be a good start (and hopefully maybe get the complexity below 50).

Screenshots
image
image

As I've never used liquid bellies before I'd appreciate if somebody familiar with the feature tests this (or we could just do the real dev thing and test it live).

Fixes #889


Razgriz1032:
LGTM

PR #1041, opened 2020-12-21
[MIRROR] Exploration/Gateway Balance Update

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/9478
--------------------
- Adds a new Scatter type of gateway, abductor, which strips the person's gear and deposits them somewhere else
- Nerfs Expedition Phasers and Holdout Phasers
- Reduces the number of full size phasers in Exploration from 4 to 3
- Slightly buffs melee mobs
- Removes humanoid mobs from Virgo 3B Wilderness
- Removes ADG from one of the POIs from V2


Darlantanis:
Do we actually want this?

PR #1139, opened 2021-01-19
Make it so Alraune can properly choose what they grow

LunarFleet:
No one plays Alraune, I figure this is part of the reason why, even though they are a pretty neat race.
If this is gonna be an issue as is, tell me, and I will nerf the production rates or somethin'
Don't just abruptly close it, I will become sad, and call you a mean name >:V

### DIFF
--- a/code/global_vr.dm
+++ b/code/global_vr.dm
@@ -34,6 +34,7 @@ var/global/list/acceptable_fruit_types= list(
 											"durian",
 											"eggplant",
 											"grapes",
+											"grass", // CH changes
 											"greengrapes",
 											"harebells",
 											"lavender",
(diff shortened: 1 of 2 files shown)


MisterGrimm:
I don't get the point of this at all, why even do this?

LunarFleet:
Because being limited to one fruit per round is kinda problematic; I want to be able to do fluff things, like give people flowers and such. That and the race has very few redeeming qualities in general. 

As I said, it is for the fluff of it. If there are any mechanical issues with it, I can nerf it or change the list of what they can grow.

LunarFleet:
I was in a call with Kassc while making this, he suggested I add a few things to the list of what they can make, so there we go

PR #1149, opened 2021-01-21
Adds Sharkpile prototype + 939

Sharkmare:
This adds a variant of trash piles I spaghettied up.

These piles have the ability to override the default spawn tables (changeable in mapping) and also can be given a % chance to just yeet themselves out of existence instead of spawning to randomize where they are. (Random placements woo)

 Each % value can be adjusted in mapping nothing is hardcoded. This also makes making unique subtypes of trash piles simpler

Not his pile does not spawn sharks (yet)


cadyn:
no idea what the point of the piles are but why not

PR #1161, opened 2021-01-23
[MIRROR] Add in extra Ram horns.

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/9588
--------------------
https://cdn.discordapp.com/attachments/746752031657558057/802392014574190632/unknown.png

image

Adds in some horns for character creastion as... ya know we could never have to many lego pieces!


Razgriz1032:
Needs icons/mob/vore/ears_vr.dmi. That is missing from the mirror for some reason.

PR #1174, opened 2021-01-24
Law and subtype preparation for new drones

Sharkmare:
Defines law subsets and typings for drones, this is step 1 of two, second step will be adding back the tool sets for the unique drones


cadyn:
Please god, no spaces in file names. This triggers me

Sharkmare:
> Please god, no spaces in file names. This triggers me

that should be an underscore, lemme fix that real quick

PR #1176, opened 2021-01-24
Vore slug and forest turf fix

### LINKED ISSUE #1173: Sif Forest/Planetuse tiles use icon_state sif_grass instead of sif_grass_dark
Title. Uncertain when this broke. My suspicion is all sif grass now using "/decl/flooring/grass/sif " as flooring. Will bundle the fix into my next PR if my suspicions are correct.

Darlantanis:
Adds a passive slug mob which leaves a trail of sticky slime. Run in said slime and you'll trip, calling the slug to slowly wander over and eat you.
-Vore mob with unique vore behavior which triggers it to try eating you if you get stuck in its trap.
-Rarely spawns on Sif near water.
-Slow and hard to kill with melee, incapable of being more than a nuisance to anybody with mob vore toggled off.
Open to feedback and suggestions on the timers and probability checks and etc.
Sprites by @JesusChristItsJasonBourne 

Additionally,
-Fixes Sif forest turfs to use the correct icons. Fixes #1173 
-Fixes an oversight with the leopardmander.

Slugs do NOT spawn with this frequency
image


cadyn:
These are actually pretty nice, good job Nadyr.

PR #1187, opened 2021-01-25
[MIRROR] Chopped up the wideborg files.

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/9609
--------------------
I divided the dogborg sprite files by department and made the code use those. This should fix the cache jank.


cadyn:
do we even need or want this

Darlantanis:
Unsure. Conflicts may be tricky though due to yw borgs.

Sharkmare:
Why would anyone want wideborgs to be across multiple files, this just boggles my mind

Darlantanis:
Apparently having them all in one file causes crashing for some clients.

Razgriz1032:
I think we can accept these, I don't think we're missing out on anything. I will test regardless.

PR #1200, opened 2021-01-27
Energy Weapons Rebalance and Misc Fixes

LuciusWorna:
Allows magnetics to be modified, adds more shots to Lasercannon and balances the laser sniper.

Magnetics can have their capacitors switched out and no longer use standard power cells by default.
Laser Cannon has been buffed to 6 shots.
Energy Sniper has increased effective range, less penalty for shooting without scoping, 6 shots by default, can switch the power cell and slightly increased cooldown between shots.

Laser Shotgun and Burst Laser are no longer holsterable.


cadyn:
Gonna try to get this and #1209 discussed today

Kazzc:
I discussed this with Lucious for an hour or so. With a 35 DPS limiter and a 350 damage potential I will consider this a small buff for these weapons.

PR #1218, opened 2021-01-31
[MIRROR] Fixes station shieldgen icon jank (supposedly)

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/9626
--------------------
Looks like there was a special type of new and completely different in every code side aspect shield gen that was added for overmap ships, but for some reason that type was using the same dmi as the other type, and while at it, copletely replaced the sprite file rather than adding to it or using its own. Now it uses its own file and the old shieldgen got its original file back.


Razgriz1032:
Fuck it. Ugly shields it is.

PR #1260, opened 2021-02-08
optimizations

cadyn:
Stole TG's Garbage collector and imported their extools.dll, updated our own rust_g with something that may or may not get added upstream: https://github.com/tgstation/rust-g/pull/60 made widelists faster, also made turfs use widelists for footstep sounds, etc.


cadyn:
Gonna actually add more to this, so don't merge just yet

Novacat:
This is definably something that should go upstream.

cadyn:
> This is definably something that should go upstream.

I never even finished this, i think there was something fucky with the garbage collector. Only thing that was functional was the widelist for footsteps, which i either moved to a different pr or gave up on, my xxhash function got accepted to the official rust-g repo though if curious, not sure whether they made any new releases since then though

Razgriz1032:
Closing because old

PR #1270, opened 2021-02-09
[MIRROR] Moves egg type options to vorepanel

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/9666
--------------------
Also makes them belly specific.


Darlantanis:
I dunno what's taking integration tests so long but the bundle is good now.

Darlantanis:
This compiles so merging

PR #1287, opened 2021-02-11
Remove YW change; allowing tesh to facesit 

LunarFleet:
Tesh thighs, do, afterall, matter
I'm cripplingly tired, I'm not sure why it thinks I changed the wikilinks when it seems to be the same, but, I'd appreciate someone who isn't sleep deprived checking too.


LunarFleet:
Okay, I think I broke this, let me try this again

PR #1300, opened 2021-02-12
Tesh wings/tails/antenna time

LunarFleet:
- Adds teshari variants of various tails/ears/wings, specifically...

- Teshari dragon fly and bee wings

- Teshari bug abdomens

- Teshari antenna 

- Adds a new neutral trait, insect stinger, which lets you inject 3 units capsaicin into someone adjacent, as an ability. Has a cooldown, of course, and it takes a while to work. Can move it to a positive trait if this is not acceptable, but, succubus drain is neutral, so I figured it is fine too

(I was in a wasp mood, what can I say)


Sharkmare:
Based commit we need more tesh body parts

LunarFleet:
There, it is positive now, 1 point cost

PR #1304, opened 2021-02-12
Fixes improper dock tag

ExabyteDiscord:
Shuttle previously could take off, but couldn't land back in the hangar once it left. This fixes that.

Single line fix go brr


Kazzc:
Yeah, I noticed that and I could not figure out how to fix it. I guess it was a code line I missed.
I had a similar issue with Ursula where it wont land back on the dock, I may have messed up the code on some of the shuttles that may need a revise.

PR #1325, opened 2021-02-14
Adds a wide selection of moth wings

Sharkmare:
Adds a wide selection of moth wings, mirrored from BEE station (adds as in we can now implement them, these are not selectable just YET)

https://github.com/BeeStation/BeeStation-Hornet/blob/master/icons/mob/moth_wings.dmi


Razgriz1032:
Says empty file

PR #1338, opened 2021-02-15
Kassc's bedtime story

LunarFleet:
Kassc requires his bedtime story
I'm not fucking with config files, but Kassc wanted a physical reference to the wiki, so here, a book, with a link, in parentheses. 
This is his favorite bedtime story.


LunarFleet:
That should do it, yes yes

PR #1343, opened 2021-02-16
Solidifies Exploration Department under the HoP

LuciusWorna:
Gives access to the HoP, revokes access and radio channels from the RD and updates the occupation tooltips appropriately.


DerBlitzkrieg:
I won't go too indepth as I already have, but I heavily disagree with this change. Its just following an example from upstream and fixing something that doesn't exist as a problem. Explo being in command purgatory is a problem, but the HoP is the worst choice of head to put explo under.

LunarFleet:
Neither HoP or RD fit well, and there is no good reason for them intervening. Personally in favor of explo being its own mini-department, but, it isn't like this is going to matter enough to be worth it either way, do whatever.

PR #1344, opened 2021-02-16
PAI changes, no suicide, move wipe

Sharkmare:
Removes PAI suicide
Moves WIPE and makes it a bit more clear
And adds the ability to unbind master, for those awkward times when your master doesn't want you anymore or you two just aren't compatible,
image


Sharkmare:
Dont merge yet, I added something new and need to test if it works because i just hacked together this web edit

Sharkmare:
It compiles!

PR #1361, opened 2021-02-18
[MIRROR] [MIRROR] Fixes Telecom Storage Access

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/9740
--------------------
Original PR: https://github.com/PolarisSS13/Polaris/pull/7874
--------------------
Fixes issue #7870


Razgriz1032:
We don't want this merged obviously, but we should remember that we need to fix it on our end as well.

PR #1476, opened 2021-03-03
adds "fitness instructor" alt title for entertainer

kcin2001:
had hannah in mind for this one and entertainer is pretty much the lump anything in job title
its a sensible job to have with our gym (makes more sense than some of the other entertainer job titles)


Sharkmare:
I'd suggest changing it into Alt titles_zz_ch.dm so that all overrides can be in one file and to have us simply separate the departments in code via comment blocks.

This way the file is easy to disable and edit.

kcin2001:
Ill do that in the morning

PR #1499, opened 2021-03-05
Shadekin Content Update + Whitelists for traits

Sharkmare:
Adds Shadekin species-specific traits to change their eye color adaptation.
Reworks phasing to take into account light condition and observers
Adds a new attack to shadekin that allows them to drain energy from carbons via attacking (this is for the aggressive boyos)

Each eye type has its own stats just like the mob versions had.

Essentially this update makes carbon shadekin be on par and even better than mob shadekin for most purposes.
As well as giving counterplay to phase-shifting.


Darlantanis:
Not a fan of that big banned species list on the kintype datum since it seems to be intended to blacklist every other species other than shadekin. Is it possible to just have an allowed species list instead so new species don't gain access to the trait when they're invariably not added to that list?

Sharkmare:
> Not a fan of that big banned species list on the kintype datum since it seems to be intended to blacklist every other species other than shadekin. Is it possible to just have an allowed species list instead so new species don't gain access to the trait when they're invariably not added to that list?

Tat's just how the trait system works, yes I too would like a white system that if not null restricts it to the listed species. But currently, this is the correct way of restricting species until someone codes an allowed species list.

Once someone does that I'll happily use the other system.

Sharkmare:
Y'know what, I'm gonna be that guy, I'm gonna make that whitelist
Gimme a sec

Sharkmare:
@Darlantanis Feature made

cadyn (approved):
Approving this after I spoke with shark in DMs about some changes, and got permission to make them myself. We will likely be seeing more content additions in the future related to carbon shadekin, and balancing changes will also likely need to be made. This PR achieves what it sets out to achieve, which is pretty much making a foundation for the rest of the changes to come.

Very-Soft:
Just peeping in here to say, this looks good to me~ Thanks for nudging this closer to how they're actually supposed to work! <3

Kazzc:
Approving based on DM discussion I originally had with them, more intent towards RP and nerfing how OP they can be. we still intend to keep them admin spawn only.

PR #1510, opened 2021-03-07
Mining additions

Holdelta:
Given that we are to be more action-oriented, I believe that mining could do with a few small updates from more action-oriented servers.

- Adds larger ore satchels
- Adds(re-adds?) cooldown upgrades for kinetic accelerator
- Re-adds the diamond pickaxe

image

Late addition of:
image


Holdelta:
Adding bluespace satchel for Research to create - putting in Draft.

Darlantanis:
So that's where the diamond pickaxe went. Glad to see it back, it gets no use with its current implementation.

PR #1512, opened 2021-03-07
[MIRROR] Restricts some size-mechanics with prefs

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/9862
--------------------
Makes step mechanics entirely an opt-in pref (~~off~~ on by default) - available in vore panel, saves
Makes being picked up/picking up others a combined opt-out pref (on by default) - available in vore panel, saves
Makes picking up require target be on Help intent
Adds a toggle unrelated to the picking up pref that allows you to not pick up others even if all conditions are met (for micro petting) - in IC tab, does not save


MisterGrimm:
I personally don't think half of this is a good idea at all. Being picked up can be instantly reverted just by resisting, since there is no resist chance to fail with it. Step mechanics exist to prevent a micro from stopping you from moving, since a 2 inch gnome isn't going to be able to stop a 6 feet person from walking over them.

The only things I like about this are the toggle to prevent yourself from picking people up, since that can be a annoying issue, and having it where you can only pick people up if they are on help intent. Although I have no clue how dead mobs will react to intent if you try to pick them up.

kcin2001:
grimm the step mechanics isnt the passible part i believe, i think it refers to the mechanics that happen on intents other than help (aka the grabbing pinning and harm stepping)

MisterGrimm:
> 
> 
> grimm the step mechanics isnt the passible part i believe, i think it refers to the mechanics that happen on intents other than help (aka the grabbing pinning and harm stepping)

`if(step_mechanics_pref && tmob.step_mechanics_pref)
				if(handle_micro_bump_helping(tmob))
					now_pushing = 0
					return
` Not sure if im looking at this wrong, but this looks like it is saying that if one side does not have the prefs it will not allow you to stand on the same tile, aka they will react like a normal mob when it comes to tiles.

Darlantanis:
Size mechanics were never intended for combat to begin with (despite being implemented with pretty clear combat advantages). I imagine with stepability off you'd just push micros just like everybody else instead of stepping on them.

kcin2001:
So this pr is pretty bad other than the temp setting to turn off pickups

PR #1517, opened 2021-03-08
Fix space transit, properly lists space station as space map

Razgriz1032:
Now that our station is defined as space map and not *just* map, we now have established it as a "multi-z complex", I think. 

Anyway it also means that I can add all the other levels as extra_z_levels and solidifies Vir as a fully functional overmap system.

:cl:
bugfix: Players can hear from the Transit level over comms again
/:cl:


Razgriz1032:
will fix later

Razgriz1032:
why

Razgriz1032:
Wrong file. I'm gonna close this and start again

PR #1540, opened 2021-03-09
Update hard override file and downports Vorestation changes of kin

Sharkmare:
This updates our override file to no longer override the functions that are now just default handling

This also a replacement for the mirror


Sharkmare:
Oh yeah, obligatory, yes I tested this. 
Had no issues with it at all.

PR #1569, opened 2021-03-10
The missing icons

Kazzc:
So the mirror bot broke and didn't get all the icons, hopefully this fixes that issue.
I got around testing it on a test server and everything seems to work properly. All the ores, materials, TGUI on the autolathe and material should work now. Cause of the silly need of everything needing copper. . . I will be adding copper to science.


Razgriz1032:
Depreciated by https://github.com/CHOMPStation2/CHOMPStation2/pull/1571

PR #1611, opened 2021-03-17
[MIRROR] Fixes prefless carbonmobs defaulting to fuzzy mode

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/9951
--------------------
This fuzzy var is basically just reserved for character pref use and a semirecent fix on another thing moved the part of the code using this var from character spawn loading code to icon update code and therefore made it non-irrelevant for non pref-loaded humanmobs (such as monkeys)

Partially fixes #9950


Razgriz1032:
ALSO freaking requires #1582

Issue #1171, opened 2021-01-24
red spacesuit replica broken sprite


kcin2001:
#### Brief description of the issue

red spacesuit replica doesnt have a body sprite
#### What you expected to happen
to wear the spacesuit

#### What actually happened
had a default oopsie sprite

#### Steps to reproduce
- (Step 1) get a red spacesuit replica
- (Step 2) wear it

#### Code Revision
- 58e59eaa6b05167b773d672d7d26d2722e03e043
 
#### Anything else you may wish to add:
![image](https://user-images.githubusercontent.com/16435949/105623454-a94d6580-5dce-11eb-9c2e-78fb6c498cde.png)

CommodoreYouth:
You wore this Suit as a Taur.
Many Taurs do not have Sprites for certain clothing items, because the clothing items generally cover the entire body, legs and feet included.  Taurs have special sprites for their legs and feet that need extra spritework to be compatible with these "missing sprites".

kcin2001:
I thought it defaults to the non taur ver if there is no special sprite?

CommodoreYouth:
Mechanically, suits/helmets that don't need to be cycled, like the Red Space Replica Suit, have a sprite that generally fits all humanoid characters, even if it makes them look odd, like missing tails or ears.  Taurs have the lower half of their body as a completely different sprite, and each Taur choice needs its own spritework done to make the suit fit THE ENTIRE TAUR SHAPE, not just the front legs.  Yes, the entire taur body, in addition to the upper body, needs a sprite.

Without the sprite, most outfits like these default to that ERROR-like greentext of just "suit".

kcin2001:
![image](https://user-images.githubusercontent.com/16435949/110072453-a8dfad00-7d32-11eb-8c7a-82983bdc0b41.png)
its broken for non taurs too...

PR #1636, opened 2021-03-21
Fixs issue 1171

Armyguy:
https://github.com/CHOMPStation2/CHOMPStation2/issues/1171

Namely by pointing it NOT to the in hand.

### DIFF
--- a/code/modules/clothing/suits/miscellaneous.dm
+++ b/code/modules/clothing/suits/miscellaneous.dm
@@ -106,6 +106,7 @@
 /obj/item/clothing/suit/syndicatefake
 	name = "red space suit replica"
 	icon = 'icons/obj/clothing/spacesuits.dmi'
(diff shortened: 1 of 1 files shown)


Razgriz1032:
This is wrong. What you are changing is for the object before you put it on, not the worn sprite.

Armyguy:
I Think? that will fix it, I'm not 100% on icon_overrides. but its this or editing the .DMI

Sharkmare:
> I Think? that will fix it, I'm not 100% on icon_overrides. but its this or editing the .DMI

You could always launch up a test server and look if it fixed it, if ya need help doing so i can provide

Razgriz1032:
I tested it myself, it works.

Darlantanis:
Will note icon_override may break species-specific icons, if they exist.

PR #1646, opened 2021-03-21
Fixes and optimizes vorepref loading.

### LINKED ISSUE #1533: carbon mobs issue
through one of the updates, the ability to vore any and all carbon based mobs has been disabled 

for mobs to not be effected by this
#### What actually happened

mobs were effected by this
#### Steps to reproduce
- (Step 1)
- (Step 2)
- (and so on)

#### Code Revision
 
i just want this to be fixed

Verkister:
-Voreprefs are now loaded when the character loads as they used to be. The initial load only gives one usable belly while the rest are loaded when the vorepanel is opened.
-Now that the prefs are loaded at an appropriate time again, the forced defaults to disable everything are no longer a prefbreak concern.
-Un-spaghettifies an "optimization" into an optimization.
-Fixes voreprefs not loading when they should load.
-Adds on-demand belly loading for simplemobs. Voremobs now only get their bellies loaded in when they try to use them or open their vorepanel (in case of player-controlled voremobs)

Fixes #1533


Verkister:
Adding something for simplemobs too before merge.

Verkister:
Annd done~

PR #1688, opened 2021-03-28
Enables Novel Posting

R3dtail:
I HATE running into the 2048 character limit.  So, I figured I'd make a PR to double the maximum number of characters allowed for an emote post, because wallposting is good, IMO.


Shadow-Quill:
Needs a chompstation edit tag

Darlantanis:
Checked with BlackMajor, code elsewhere should ensure this can't be abused for DDoS spam.

PR #1710, opened 2021-03-31
Added safety lock feature to phaseguns, same as frontier phaser

Razgriz1032:
:cl:
add: Added safety lock feature to phaseguns, same as frontier phaser.
/:cl:


CommodoreYouth:
We'll never get crates open in Cargo ever again.
Augh.

We also won't be able to shoot Blobs with their biggest weakness.
Double AUGH.

CommodoreYouth:
Dangit Raz.

icowom:
Uh. Why was this needed? Some folk don't want to have to melee bash station event mobs too when Sec is nonexistent or just one dude. Phase weaponry is the only ranged option we have.

PR #1753, opened 2021-04-07
Fix hang for actions on servers that don't have a database

Razgriz1032:
What's happening is some attack logs are getting recorded to SQL database, and if there is no SQL database, it hangs and gets angry while it tries to figure out how to write to a non-existent database. It doesn't affect live, but it will affect people running test servers or private downstreams.

Amending #1460


Razgriz1032:
I actually have no idea what I'm doing, lol

cadyn:
Oops, was trying to look at it on mobile

Razgriz1032:
NVM, just turn this off in config.txt

## Enable/disable SQL connection (comment out to disable)
SQL_ENABLED

PR #1759, opened 2021-04-07
[MIRROR] Trait Rebalance and Refactor

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/10033
--------------------
- New typepaths added for Positive, Neutral, and Negative Traits
- Colorblindness traits moved to Neutral traits, no longer refund points
- Custom species now start with 2 trait points
- Major Conductivity only refunds 1 point now
- Major Hardy only reduces item slowdown by 75%, costs 3 points
- Hardy cost upped to 2 points
- Reduced some extremely high/low values in some traits (flash_mod is the most notorious offender)
- Hard Feet and Flight reduced to 0 points


cadyn:
Brain't

PR #1764, opened 2021-04-08
[MIRROR] Make unnatural resizing even better and more efficient with the power of DCS:tm:

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/10095
--------------------
Title. Makes #10090 and #9545 better while retaining the fixes required by #9950.
This is almost a revert of #10090 because I removed most of what it did in favor of this approach.

Basically, the way this works is that whenever resize() is called, the following will happen:

 - If uncapped = TRUE and your area permits this:
   - If you are scaling to an unnatural size, it will add a one shot component to return you to capped sizes when you move outside of this area
   - If you are scaling to a size within bounds, the component will be deleted if present.
 - If uncapped = TRUE and your area does not permit this, it will assume adminbus and permanently set your size to that value.
 - If uncapped = FALSE, the component will be deleted if present and your size will be capped.

The component in question is referred to as `one-shot` because it will delete itself as soon as it detects that you're resized out of bounds when you're not supposed to after resizing you down. The way it does this is by listening to `COMSIG_ATOM_ENTERING`, which is called from the base of atom/Entered() *on* the atom that just moved into a new place. This covers basically every possible way to leave an area, as any time you enter a new turf or area, you'll have your area checked for size limitations and the component will fire and reset your size before deleting itself.

This doesn't seem to have much performance impact at all.
> /datum/component/resize_guard/proc/check_resize   0.000   0.000   0.000  0.000  70


Darlantanis:
Merge #1760 first

PR #1842, opened 2021-04-22
Catalogue update

Razgriz1032:
Supplants #1827 because stop paring useless crap with lore crap.


Darlantanis:
I'll get to this eventually I swear

PR #1844, opened 2021-04-23
[MIRROR] tumby icons

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/10195
--------------------
maybe I will do more later or something I donno. Wanted to make some nice animated ones but these will do for now.

I should note. These aren't supposed to be anything in specific beyond 'fleshy fleshy' so, you know. Use your imagination or whatever.


Razgriz1032:
Ahh yes, a nice blank PR because we have modified the icons before.

Razgriz1032:
image

Razgriz1032:
Here's ours with the two new ones. IDK if we should accept these because they aren't grayscale 
image

PR #1860, opened 2021-04-27
Adds Liquid Copper/Aluminium to Sheets Reactions

ZVB334:
So first off the agony that is the reagent of aluminum being called 'aluminum' and the sheets being called 'aluminium'.

Second, you can now use xenobotany frost oil to make copper and aluminum sheets.


Darlantanis:
This would be best either added to Chemistry-Recipes_ch.dm (our file for chomp-specific chemistry recipes which means no future conflicts) or sent upstream.

ZVB334:
> 
> 
> This would be best either added to Chemistry-Recipes_ch.dm (our file for chomp-specific chemistry recipes which means no future conflicts) or sent upstream.

My bad, my mind completely skipped over doing that. It's in the modular _ch file now.

Darlantanis:
No worries, it can be super unclear where to put code when working in a downstream codebase.

PR #1876, opened 2021-05-02
Webedit ore spawning fix

Darlantanis:
So #1863 converted ore lists to a static list. Problem is we already have a global list defined for ore_types which was never updated for new ores and seems to override the new, updated list. This PR deletes the now unneeded list, hopefully fixing the bug.

Webedit PR so untested/etc.


Razgriz1032:
Tested, seems to work.

PR #1883, opened 2021-05-05
[MIRROR] Adds various new fat taur variants

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/10256
--------------------
wow look at that who knew I could sprite?
Made these a while back but the person I asked to actually do the code side forgot to add them here instead of just another server.

Have some example pics in dumb colours
Fat lizard:
image

Fat synthlizard:
image

Fat synthfeline:
image

Fat synthwolf:
image


Darlantanis:
These already exist in a _ch file. Conflicts maybe?

PR #1942, opened 2021-05-11
[MIRROR] Tweak synth stuff

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/10298
--------------------
isSynthetic() on humans was severely overloaded with performing work when it's just an accessor function, so I moved the work to the appropriate places (and also reverted the lynx quick fix):

- Mechanical stomach provides the reagent purge verb (orchestrator on proteans)
- Diagnostic organ provides the self-diagnostics verb (orchestrator on proteans)
- Head organ provides monitor set verb if robotizing to a head that supports it, otherwise removes it
- Torso sets the `synthetic` var on owner to robolimb singleton (previous isSynthetic was handled by checking torso, so all cases of a human being isSynthetic() true should be the same: ones where torso is robotized)

Edit: Also ports /tg/'s current glob Initialize(), due to https://github.com/tgstation/tgstation/issues/26954 showing it's head here.


Darlantanis:
This is missing like 11 files for some reason

PR #1943, opened 2021-05-12
Pred POI's

Kazzc:
Adds two more wilderness POI's that has two very hungry large preds~


Kazzc:
Finally

Zeldazackman:
blackwolf-1 has an incorrect sprite when facing to the left and instead seems to have the sprite state from whitewolf-1

PR #1955, opened 2021-05-14
[MIRROR] Adds in Booze hounds.

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/9969
--------------------
Adds in the booze hounds! Big boyos that are here to serve drinks, and be a dunk tank. They're effectively service hounds with slightly less, but are all sprited and animated. A new addition for those who finds them interesting.


Darlantanis:
I guess I should say the discussion tag is whether we take these or comment out and keep ours.

Razgriz1032:
I'm scared, but we should probably accept this.

Razgriz1032:
We'll take it and hopefully unfuck the YW implementation

PR #1983, opened 2021-05-20
Fixes voremobs not loading bellies when fed/dropped/slipped into

Verkister:
Fixes voremobs not loading their bellies when fed or dropped/slipped into


Verkister:
oops i may have messed up. fixing rq

PR #2054, opened 2021-05-27
[MIRROR] [MIRROR] Linter Introduction + Cleanup

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/10433
--------------------
Original PR: https://github.com/PolarisSS13/Polaris/pull/8085
--------------------
This adds the required defines to use the SpacemanDMM DreamChecker linter, which can be used inside of certain IDEs such as Visual Studio Code. It can also be integrated into CI, so that someday in the future we can have our CI point out spotty code.
Documentation for that can be seen here: https://github.com/SpaceManiac/SpacemanDMM/tree/master/src/dreamchecker

Unfortunately, Polaris has over 500 problems raised by the linter, and I assume Virgo has even more, so I figure I would do my part and try to reduce the number down so that someday we can become A Prim And Proper Codebase:tm: like Nebula. I've shoved linter error fixes into seperate commits so that it might be easier to review, or to be cherrypicked into a seperate PR if @  Atermonera wants that. 

Incidental fixes as a result:
* This also fixes a bug with fax machines, where you cannot actually fax the `Supply` admin department due to the check if it was SifGov always evaluating to true.


Razgriz1032:
"Unfortunately, Polaris has over 500 problems raised by the linter" 
I bet you do

PR #2081, opened 2021-05-30
Vox Recoloring Edition: COLORFUL VOX, BAWK!

Somememeguy:
Hey you! Yeah I'm lookin' at you, our five entire vox player base.

Have you wanted to be able to... recolor those colorful lil' scales of yours?

Have you looked at all the other vox and thought, 'Man, why can't I be different?!'

Well WAIT NO LONGER, for today I give to you... recolorable vox!

HZOEYervk6

Become a pretty pink vox like you always wanted to be! Sadly, hair is yet to be recolorable, but hopeffully that will change one day!

Also, as the old version is now recolored, another version of old vox has been added so as to make sure there is always an un-colored version of the original.


Darlantanis:
Each tweak in vox.dm needs a //CHOMPedit comment, otherwise I'll end up overwriting this if upstream touches the file.

Somememeguy:
> Each tweak in vox.dm needs a //CHOMPedit comment, otherwise I'll end up overwriting this if upstream touches the file.

Done!

Darlantanis:
I missed this when reviewing but modifying any upstream .dmi files will result in said icons getting overwritten if upstream modifies them. This is typically best resolved by tweaking code to read from a _ch.dmi file instead. Though I can merge this if that's an acceptable risk given upstream hasn't touched either for four years now.

Somememeguy:
> I missed this when reviewing but modifying any upstream .dmi files will result in said icons getting overwritten if upstream modifies them. This is typically best resolved by tweaking code to read from a _ch.dmi file instead. Though I can merge this if that's an acceptable risk given upstream hasn't touched either for four years now.

Yeah it's an acceptable risk. Idk if they'll ever update vox, and if they do, I'm sure they'd make it recolorable.

PR #2131, opened 2021-06-05
Enables vore for raccoons

Verkister:
The sprites for it exist so why not


Razgriz1032:
Does the vore sprite for it exist? I can't check now because I am on vacation.

Verkister:
kuva

PR #2137, opened 2021-06-06
Fixes great wolf den minions being hostile

Verkister:
Presuming they were intended to be nonhostile according to the note mapped in.


Kazzc:
It's a very simple change, they changed the AI holder to a retaliate AI holder. easy fix.

PR #2164, opened 2021-06-09
Adds belching whilst full

radiantAurora:
Adds a new vore panel setting to enable belching whilst full. Basically the opposite of current hunger noises.
Also adds in 15 belches (from Jeschke's free-to-use 2020 Vore Day sound pack) that the system and potential future systems I plan to make will use.

More details on the additions:
Mobs with the setting enabled will belch when over 500 nutrition, with the probability (and therefore frequency) of belching rising with nutrition up to 4075 nutrition at 10%

Addition of an 'audible belches' preference so clients can disable belch sounds. This defaults to SILENT.


Darlantanis:
This needs chompedit comments on setting_datums.dm. It would also be preferable to add to the comment summary in VorePanel.js. Chompedits are summarized in lines 48-168 as tgui .js files do not seem to support in-line commenting.

radiantAurora:
Oh wait I should probably comment here so you know I added the comments oops

Razgriz1032:
Aurora, you need to pull from master into your fork again. Get it all updated. And TGUI bundle needs to be updated after.

radiantAurora:
Updated

PR #2169, opened 2021-06-11
Fixes an artefact ignoring player TF preference

### LINKED ISSUE #2132: Passive radiation/artifact mutations ignore Spontaneous TF prefs
Note that I was told that this might be related to a xenoarcheology artifact instead of your standard radiation, probably, but the point still stands. I was getting green messages like "Your whole body vibrates" when I was near an area with potential artifact spots in the mines, and yet I wasn't getting irradiated like I expected. No deformities or stinging from toxins. Only realized what was happening when I suddenly grew larger; under my hardsuit, I was TF'd into a monster of various markings/parts. This didn't obey my "no spontaneous TF" pref setting, either, which is the main reason I'm reporting this.

To not be randomly TF'd into a chimera and need admin help to go back to normal.

#### What actually happened
My hands became purple wings and my head turned green, amongst other things that shouldn't have happened.

#### Steps to reproduce
- (Step 1) Go dig around near patches of glowing cracks.
- (Step 2) Get unlucky with either radiation or this mysterious artifact I'd never noticed.
- (Step 3) Stay a while and watch yourself TF out of control, regardless of pref settings.

#### Code Revision
- d5ca48ba11fc67b334f7ccf2ab4e774cbd5d43ee
 
- For what it's worth, I had it happen to me in the southwest corner of the mining area.

radiantAurora:
Title. Hopefully. Should fix #2132 
This is a webedit done from mobile but in theory this should work. I can't test though and I'm about to sleep.

### DIFF
--- a/code/modules/xenoarcheaology/effects/dnaswitch.dm
+++ b/code/modules/xenoarcheaology/effects/dnaswitch.dm
@@ -13,7 +13,7 @@
 
 /datum/artifact_effect/dnaswitch/DoEffectTouch(var/mob/toucher)
 	var/weakness = GetAnomalySusceptibility(toucher)
-	if(ishuman(toucher) && prob(weakness * 100))
+	if(ishuman(toucher) && is_preference_enabled(/datum/client_preference/allow_spontaneous_tf) && prob(weakness * 100))
 		to_chat(toucher,pick("<font color='green'>You feel a little different.</font>",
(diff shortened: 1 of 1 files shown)


radiantAurora:
fuck

Darlantanis:
untested webedit bad

radiantAurora:
Don't see myself working on this any time soon, closing.

PR #2219, opened 2021-06-18
[MIRROR] TGUI 4.3

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/10636
--------------------
I'm putting this up because I need help reviewing it. This update was performed using the "replace the TGUI folder" approach, and this means that a lot of properties on components and .scss file tweaks were wiped out, rendering a number of our UIs subtly but annoyingly different. I need to go through each changed file and look for changes like this:
 

The linter can't tell us about missing properties, but they will nevertheless cause issues. Basically, we need to check for regression.


Razgriz1032:
This goes perfectly with no conflicts. This also modifies 332 files while Virgo's modifies 333 files. 
one fear

Darlantanis:
I found it

PR #2233, opened 2021-06-18
Added a gooey shadekin

SarmentiCampbell:
Uh, I'm adding a subtype of Rakshasa since the current one way too op currently and can't be spawned without having to VV each time


Razgriz1032:
No. This can't pass because it would get access to admemey verbs as a subtype. It's not the stats alone that make it what it is.

Issue #2190, opened 2021-06-13
Torta's in-hand sprite is a pink square


radiantAurora:
#### Brief description of the issue
Picking Torta up TFs you into a pink square

#### What you expected to happen
See an armadillo in my hand

#### What actually happened
I'm now a pink square

#### Steps to reproduce
- Pick up Torta (and presumably any other armadillo)

#### Code Revision
Commit: b6dbf2cbb5ae2072de74c5a000a2f146718e1373
 
#### Anything else you may wish to add:
![image](https://user-images.githubusercontent.com/80002839/121814027-cc39c280-cc66-11eb-8808-10a9ad3fb9cd.png)

Darlantanis:
Dug a bit, I missed this because it only occurs with the left-hand sprite. Right-hand defaults to a blank sprite while left defaults to a pink square.

PR #2238, opened 2021-06-19
[MIRROR] [MIRROR] Adds blank icon state to lefthand_holder.dmi

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/10692
--------------------
Original PR: https://github.com/PolarisSS13/Polaris/pull/8137
--------------------
This should prevent mob holders from turning players into a pink square if there's an error with the icon_state.
image

righthand_holder.dmi already uses a blank icon by default if the icon state doesn't exist but for whatever reason lefthand_holder.dmi does not.


Darlantanis:
Fixes #2190

PR #2259, opened 2021-06-21
[MIRROR] [MIRROR] Refactors posters to use decls

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/10698
--------------------
Original PR: https://github.com/PolarisSS13/Polaris/pull/8128
--------------------
Tested: they work the same (or perhaps better, testing showed that it might not've been possible to previously set posters in a dmm to reliably have the same serial number). They still flick() the animation when you put them up, but it actually uses a do_after() like modern code.

In the following, all 5 posters were mapped in. The first 3 from the left had their state explicitly set by setting the `poster_decl` var to the path of the desired decl, while the other 2 were random.
https://user-images.githubusercontent.com/8475252/120964258-21298600-c707-11eb-8103-660fbaa2a3c9.mp4

Removed posters and haircombs from the contraband random item spawner because those uh... aren't contraband?
And reduced the cost of the security posters crate to 10 because c'mon, it's just posters

Specifically requested by @  PrismaticGynoid for some PoI work for the Cynosure.


Razgriz1032:
Useless feature

PR #2263, opened 2021-06-21
Decrease mine mob spawner chance

Razgriz1032:
This should halve the total amount of spawners in the mines.


Razgriz1032:
CI don't wanna look at this, I guess.

PR #2278, opened 2021-06-23
Fixes vorepanel screen overlay menu

Verkister:
Apparently this fixed it even though I have not the faintest clue about how :v


Darlantanis:
I don't get it either.

PR #2301, opened 2021-06-27
xeno hunter dome bump

kcin2001:
i apparently missed bumping this up a pixel the last time i fixed this, also removed some stray pixels that i found


Darlantanis:
This should probably be sent upstream, otherwise it will get overwritten if upstream touches these .dmi's.

kcin2001:
ah yeah i forgot i meant to put this in ch dmis
shouldnt be an issue to just copy the sprites over to a ch one and change the code and leave the modified one in the _vr?

kcin2001:
and looking at the ch files... i did do that, but the code is unchanged

kcin2001:
eventually i should just make them ears, the reason i did this like this intially is lessened and ill be able to use the multi colors

PR #2315, opened 2021-06-30
[MIRROR] Port /tg/ vis_contents openspace

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/10764
--------------------
Removes the openspace subsystem, as it's not needed now!
Much less CPU intensive on the server, though get_mobs_and_objs_in_view_fast() could use some looking at.
Also allows multiz lighting (sort of).

Also ports glass floors.

zrJbeRe6WG


Razgriz1032:
Missing files are 
code/game/atoms_movable.dm
code/game/objects/items/weapons/storage/storage.dm
code/modules/mob/inventory.dm
code/modules/mob/living/carbon/human/inventory.dm
code/modules/mob/living/inventory.dm

Razgriz1032:
Missing files are apparently not needed at all because #2316 already reverts them. FROM WHAT I CAN TELL, this PR is already seeing the changes from THAT PR and is like "bruh, I don't need these files, they're gonna get reverted anyway." It's playing 4D chess. I think this PR is self-aware. In addition to Unit Tests, I wonder if CI can do Turing Tests.

PR #2493, opened 2021-07-10
[MIRROR] Move like every clothing icon file

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/11002
--------------------
MMMMMM. That was fun.

Note to downstreams: This removes the `_s` required at the end of every uniform-slot-clothing icon state. You might need to just rename some states in your dmis. Shouldn't need any code changes. If this DOES mean you need to do code changes, then you were probably unnecessarily specifying what icons clothes use when they could have inherited it.

image

I think this will be easier for people to understand. Glasses in eyes/item.dmi for the items, eyes/mob.dmi for the mob, etc.
No more "Well it's icons/obj/glasses.dmi for the item and icons/mob/eyes.dmi for the mob sprites" or just having zany names.

I'm sure uh, I broke a thing or two, it was a lot of lines changed. I tested some 'sensitive' stuff like teshari sprites, and fixed up a few backpack state names while I was in there, and merged some of their needlessly separate files.

PR #2503, opened 2021-07-10
[MIRROR] Port a big blast door, add to ship

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/11015
--------------------
I know it's super unrelated but also fixes a couple suits that were incorrectly coded to use icon_override


Darlantanis:
Relies on #2493

PR #2535, opened 2021-07-13
Adds bellyrub mechanics

Verkister:
Adds a bellyrub verb to manually trigger a belly processing cycle for the target's selected vorgan to help it process its contents faster using a lighter version of the regular belly cycle to avoid straining the subsystem. The bellyrubs can be accessed via an ability tab verb, and can also be easily applied by targeting groin for a hug alternative. This also works on simplemobs and simplemobs can also target groin on help intent to provide rubs.
Also fixes tamed oties smashing through shit that gets in their way.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/species.dm
+++ b/code/modules/mob/living/carbon/human/species/species.dm
@@ -431,7 +431,7 @@
 		H.visible_message( \
 			"<span class='notice'>[H] boops [target]'s nose.</span>", \
 			"<span class='notice'>You boop [target] on the nose.</span>", )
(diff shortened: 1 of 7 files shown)


Zeldazackman:
Can't wait until someone speedruns making a fat bald golden "buddha" assistant

Darlantanis:
Might be better to use BP_GROIN instead of hard-coding the zone selection. Otherwise looks good to me.

Verkister:
should probably udate the rest of the hug alternatives to use the proper definitions too, but that's something for an upstream fix later

PR #2644, opened 2021-07-24
Tiny ddos fix

BlackMajor:
Examine spam of this forces the server to generate and compress the icon's .dmi into the chat window for every examine.


Darlantanis:
Needs CHOMPedit comment or I will overwrite this later.

BlackMajor:
Why do I always forget those

PR #2654, opened 2021-07-25
Less Ringa ding and work in progress new casino manual

Shadowfire117:
Now people cant spam the slot machines and get platinum coins, they are once more special! Also new codex casino manual. Albeit still entries needs to be done at later moments but is being added for safety reasons and to access while I only have a laptop to work on.


Darlantanis:
Map edits required.

Shadowfire117:
Thank you, only realized this morning I forgot to replace book pathway for the casino.

PR #2738, opened 2021-08-06
Adds a new experimental item digestion mode

Verkister:
Adds a new experimental item digestion mode that divides the digestion damage between all items within the gut instead of going through one item at a time. Disabled by default.
Experimental because I don't know how badly it affects performance and because I'm feeling too lazy to make a vorepanel button for it rn, so it's a varedit toggle for now.


Razgriz1032:
Default off is good idea when releasing something new into the wild

PR #2747, opened 2021-08-07
Fix being unable to put micros in shoes

Razgriz1032:
Thanks to TheGreatKitsune for this fix.


CommodoreYouth:
Looks like the checks all passed.

PR #2748, opened 2021-08-07
Protean now gets default survival box

Razgriz1032:


Razgriz1032:
This actually did not work. IDK what I was thinking. I passed the fix to Rykka for #2798

PR #2773, opened 2021-08-10
Adds a new chompspecific debug map 

lbnesquik:
## About The Pull Request
This PR adds a new debug map. It has mechs, guns, medical, research and hopefully a working supply shuttle soon, along with most of the vendors in the game. The engine bits of southern are also literally tacked on, if that's something you want to test.

This PR also makes this map the default map to load.

I just need to make the spawns work on it, and then we'll probably be good.

Virgo minitest is way outdated and i don't feel like touching their files, so i made a new one. 
Also for something like that, it is just as easy to start again than improve again.

It is the default map for efficiency reasons, making it easier to code on this place as it would compile and startup much faster without having to load SIF.


lbnesquik:
Lacking motivation to work on this codebase.

PR #2777, opened 2021-08-11
Gives Medihound borg module surgery Tools

User-U-U:
This PR gives Medihound borgs the following surgery tools

- Scalpel
- Hemostat
- Retractor
- Hemostat
- Cautery
- FixOvien
- Bone setter
- Circular Saw
- Bone Gel
- Surgical Drill

The medihound module often has nothing to do, and since the medical system here has a decently high chance of broken bones and they can't perform surgery often times this module feels useless. Since I would like to play medihound more I made this PR hoping it would help me feel less like a third wheel. 

There are still some operations they can't do, I left out the organ gripper so transplants are not possible, neither are NIF installs.

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_modules/station_ch.dm
+++ b/code/modules/mob/living/silicon/robot/robot_modules/station_ch.dm
@@ -6,3 +6,17 @@
 /obj/item/weapon/robot_module/robot/engineering/New()
 	..()
 	src.modules += new /obj/item/weapon/pipe_dispenser(src)
+
+
+/obj/item/weapon/robot_module/robot/medihound/New()
+	..()
+	src.modules += new /obj/item/weapon/autopsy_scanner(src)
+	src.modules += new /obj/item/weapon/surgical/scalpel/cyborg(src)
+	src.modules += new /obj/item/weapon/surgical/hemostat/cyborg(src)
+	src.modules += new /obj/item/weapon/surgical/retractor/cyborg(src)
+	src.modules += new /obj/item/weapon/surgical/cautery/cyborg(src)
+	src.modules += new /obj/item/weapon/surgical/bonegel/cyborg(src)
+	src.modules += new /obj/item/weapon/surgical/FixOVein/cyborg(src)
+	src.modules += new /obj/item/weapon/surgical/bonesetter/cyborg(src)
+	src.modules += new /obj/item/weapon/surgical/circular_saw/cyborg(src)
+	src.modules += new /obj/item/weapon/surgical/surgicaldrill/cyborg(src)


CommodoreYouth:
No mention of the Bioregenerator, which is the tool we use for Dehusking, Oxygen, and Toxin damage?
The Normal Surgical Module has one.

User-U-U:
> No mention of the Bioregenerator, which is the tool we use for Dehusking, Oxygen, and Toxin damage?
> The Normal Surgical Module has one.
> And I thought hounds were by nature not designed to be as functional as a Standard Cyborg, instead being more intended for RP and Vore?

I did not include that since it inst 100% necessary and to add some distinction to surgeon borg. Also, some people want to play the game and play hound modules because they prefer the look, just about every hound module can compete with it's non-hound counterpart, like engineer, science, security, as of now medical can not and is rather boring.

lbnesquik:
This is the entire point of the surgery cyborg, though. I understand having nothing to do, but this might be a little too strong.

User-U-U:
> This is the entire point of the surgery cyborg, though. I understand having nothing to do, but this might be a little too strong.

No one plays surgery borg

User-U-U:
Even then I don't think it really makes sense to split up the normal medical medical modules either, if you don't have surgery tools as a medical borg you are next to useless.

AdmiralDragon:
So if we look at other departments. The same borg sprites for example Sec ALL have the same load-outs unless it is a special borg. So to me, medical borgs all being specialized in one area or another is tedious and annoying.

So while 1 Sec hound can take up 1 officer position. (The same applies to Engineering borgs)
On medical it would take, 2-3 depending on the exact details of the load-outs, which then means 1 medical Hound takes up 1/3'ed of a medical doctors position.

This is where power-gaming should be considered though as Medical has split jobs. Medical Docs who do all the fixing, chemists who make the meds, EMT's and so on.
Just like other departments, 1 borg can fill the entire department if they are robust enough, but any respectable borg player can let the chemist do chems, or the doc fix peeps or so on just like they do in other departments.

"No one plays surgery borg" - User-U-U
This is very correct and because the server is a Vore server, who the heck wants to hang with a robot that looks like the one from star wars that fixes up Darth Vader? ew no.

User-U-U:
> So if we look at other departments. The same borg sprites for example Sec ALL have the same load-outs unless it is a special borg. So to me, medical borgs all being specialized in one area or another is tedious and annoying.
> 
> So while 1 Sec hound can take up 1 officer position. (The same applies to Engineering borgs)
> On medical it would take, 2-3 depending on the exact details of the load-outs, which then means 1 medical Hound takes up 1/3'ed of a medical doctors position.
> 
> This is where power-gaming should be considered though as Medical has split jobs. Medical Docs who do all the fixing, chemists who make the meds, EMT's and so on.
> Just like other departments, 1 borg can fill the entire department if they are robust enough, but any respectable borg player can let the chemist do chems, or the doc fix peeps or so on just like they do in other departments.
> 
> "No one plays surgery borg" - User-U-U
> This is very correct and because the server is a Vore server, who the heck wants to hang with a robot that looks like the one from star wars that fixes up Darth Vader? ew no.

I am not sure what you mean, any doctor, emt, chemist can overlap with nearly every job in the department already, even the normal surgeon model can do emt work, it is just dragging someone from a to b, if you want the beaker removed to prevent chemistry work I can implement that? Though I really don't see your point here, even engineering borg can take up every job in engineering, they can set the engine, and do atmos work and that is normal, so can science borg, they can do R&D, robotics, and xeno bio(with issue on the xeno bio part) this is a normal thing that literally only doesn't apply to the Medihound module

Darlantanis:
Imagine balancing being a dog with restricting the utilities other borgs get.

We generally do not enforce department subjobs here.

PR #2778, opened 2021-08-11
Change occupation selection color of pilot to be clearly listed as ex…

Razgriz1032:
…plorer job


Darlantanis:
I guarantee you I will end up overwriting this later if there's not a CHOMPedit comment.

PR #2779, opened 2021-08-12
Lowers the cost of producing liquids, increase the max capacity & indents a comment

lbnesquik:
## About this PR
It brings the cost of producing any liquid from 15 to 5.
It increases the max capacity to 200.
It fixes a little indent so things look prettier

Liquids currently give you really, really poor levels of nutrition, and feeding scenes are rare. It is just silly to me that we have to spend half our nutrition on it. 
Max capacity increase is for those people who want to actually have a lot of liquid. I can't see this breaking anything or unbalancing anything.

### DIFF
--- a/code/modules/vore/eating/belly_obj_vr.dm
+++ b/code/modules/vore/eating/belly_obj_vr.dm
@@ -237,7 +237,7 @@
 		START_PROCESSING(SSbellies, src)
 
 
-	create_reagents(100)	//CHOMP So we can have some liquids in bellies
+	create_reagents(custom_max_volume)	//CHOMP So we can have some liquids in bellies
 	flags |= NOREACT		// We dont want bellies to start bubling nonstop due to people mixing when transfering and making different reagents
 
 
--- a/code/modules/vore/eating/vorepanel_vr.dm
+++ b/code/modules/vore/eating/vorepanel_vr.dm
@@ -1119,25 +1119,25 @@
 				if("10 minutes")
 					host.vore_selected.gen_time = 0
 				if("30 minutes")
-					host.vore_selected.gen_time = 2
+					host.vore_selected.gen_time = 4
 				if("1 hour")
-					host.vore_selected.gen_time = 5
+					host.vore_selected.gen_time = 10
 				if("3 hours")
-					host.vore_selected.gen_time = 17
+					host.vore_selected.gen_time = 34
 				if("6 hours")
-					host.vore_selected.gen_time = 35
+					host.vore_selected.gen_time = 70
 				if("12 hours")
-					host.vore_selected.gen_time = 71
+					host.vore_selected.gen_time = 142
 				if("24 hours")
-					host.vore_selected.gen_time = 143
+					host.vore_selected.gen_time = 286
 				if(null)
 					return FALSE
 			. = TRUE
 		if("b_liq_reagent_capacity")
-			var/new_custom_vol = input(user, "Choose the amount of liquid the belly can contain at most. Ranges from 0 to 100.", "Set Custom Belly Capacity.", host.vore_selected.custom_max_volume) as num|null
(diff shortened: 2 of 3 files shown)


Razgriz1032:
Noodleserg warns that increased max cap will fuck with the examine text

Darlantanis:
I don't see an issue with this given the general uselessness of nutrition but if examine options are affected by this variable then they need corresponding tweaks to avoid breaking things.

Verkister:
I think the gen_time var means affects time, not capacity, and probably shouldn't be doubled there.

Verkister:
Also the examine amounts are located in living_ch.dm line 48 and so on.

code/modules/vore/eating/vorepanel_vr.dm
```diff
@@ -1119,25 +1119,25 @@
...
 				if(null)
 					return FALSE
 			. = TRUE
 		if("b_liq_reagent_capacity")
-			var/new_custom_vol = input(user, "Choose the amount of liquid the belly can contain at most. Ranges from 0 to 100.", "Set Custom Belly Capacity.", host.vore_selected.custom_max_volume) as num|null
+			var/new_custom_vol = input(user, "Choose the amount of liquid the belly can contain at most. Ranges from 0 to [host.vore_selected.custom_max_volume].", "Set Custom Belly Capacity.", host.vore_selected.custom_max_volume) as num|null
 			if(new_custom_vol == null)
 				return FALSE
-			var/new_new_custom_vol = CLAMP(new_custom_vol, 10, 100)
+			var/new_new_custom_vol = CLAMP(new_custom_vol, 10, host.vore_selected.custom_max_volume)
```
Verkister, line 1140: Also hardcode these to 200 if you don't wanna end up every future setting to be limited to the earlier one, such as never being able to set the belly above 50 unit again if you happen to lower that even once.

code/modules/vore/eating/vorepanel_vr.dm
```diff
@@ -1119,25 +1119,25 @@
...
+					host.vore_selected.gen_time = 34
 				if("6 hours")
-					host.vore_selected.gen_time = 35
+					host.vore_selected.gen_time = 70
 				if("12 hours")
-					host.vore_selected.gen_time = 71
+					host.vore_selected.gen_time = 142
 				if("24 hours")
-					host.vore_selected.gen_time = 143
+					host.vore_selected.gen_time = 286
```
Verkister, line 1132: Doubling these will only double the amount of time it takes to produce a single unit and that would mess things up.

code/modules/vore/eating/belly_obj_vr.dm
```diff
@@ -237,7 +237,7 @@
 		START_PROCESSING(SSbellies, src)
 
 
-	create_reagents(100)	//CHOMP So we can have some liquids in bellies
+	create_reagents(custom_max_volume)	//CHOMP So we can have some liquids in bellies
```
Verkister, line 240: Might want to hardcode this too in order to avoid things becoming impossible to increase after decreasing it even once.

Verkister (changes requested):
custom_max_volume seems to be the customizable capacity setting and shouldn't be used as a solid cap. Looks like just changing the number on the if("b_liq_reagent_capacity") and create_reagents parts to 200 would have been enough to raise the cap

Verkister:
And while all the other available liquid options might be near worthless on nutrition, honey still gives 10 per unit. I'd personally prefer to have some costlier liquid options too because I often tend to use it to get rid of some extra lard without having to spam the gym for hours, ~~and on one character as a way to produce something fuffed as a precious non-abundant resource~~, but then again I've been planning to add some more options for that myself anyway

lbnesquik:
This change shouldn't be so hard. Me angry.

lbnesquik:
This PR bloated out of the reach i appreciate. I'll get back to this another time.

PR #2791, opened 2021-08-13
[MIRROR] Przy creations port from Chompers

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/11398
--------------------
Przy/Serdy was like 'psst look at my hotties' and I was like :0 and he was like 'you COULD port them' and I was like OK

- Replaces ERT borg sprite he made in 2015 with a nicer one he made in 2021! Removes it's baton because it has ZAPJAWS now. Also a unique sleeper.
- Leopardmander! Normal glassbelly and bloo.
- Dragons! Boss mobs that let you pick sprite accessories for them and have telegraphed attacks sometimes.
- Big waffs! They're big and they're wolfeh.

To Chompers: I didn't change any of the functionality of these except for a commented-out `belly_fullscreen_color` in bigdragon.dm because we don't have that feature and I'm too lazy to port it. So! You should just be able to delete your _ch files. I also ported autotransfer so you can remove that from the belly obj file you have.

dreamseeker_t4o1BSMDJL


Arokha:
Just delete your files and uncomment `belly_fullscreen_color` in bigdragon.dm

PR #2802, opened 2021-08-15
Bandaid fix for inexplicably missing staff titles

Razgriz1032:
I think we are having database issues, or it seems to not be wanting to read from admin_ranks.txt. This is a bandaid fix by adding the titles to the else statement. Other workaround is just to edit database manually.

It works fine offline, but on live it has this problem. Only diff I can think of is the active database.


Razgriz1032:
nvm found the problem

PR #2816, opened 2021-08-16
Adds 'Horizontal on rest/stun' proc for simplemobs, assigns to shadekin

Raeschen:
The first time i've tried to do something a bit more serious. Please be constructive if something is wrong!

This allows specified simplemobs to go horizontal like carbon/humans when laying if the proc is assigned as an update_transform() override for a simplemob subtype. If an icon_lay sprite is defined, that should take precedence over this code for forwards compatibility (I.e. no horizontal will occur).

currently assigned to shadekin and nothing else.

Tested with sizeguns, seems ok.
Tested other simplemobs unaffected, seems ok.


Raeschen:
will probably tweak this later to put the horizontal laying code at a lower level isolated proc and have the shadekin-level layer simply call that proc in an update_transforms override

then any other simplemobs that want that functionality can just call the proc in their own update_transforms override if that's better

Darlantanis:
Preferably a large chunk of code like this should have //CHOMPedit end at the bottom to make it clear what's ours and what's upstream.

Raeschen:
if simple_mob.dm is not the correct place for the extra proc, let me know and i can relocate it. 
the previous layer mob/living/ has it in living.dm so i assumed it was okay

Darlantanis:
As far as conflicts/comments go, the easiest approach is to put out code in a _ch.dm file where possible. For here, that would be something like creating a simple_mob_ch.dm (if it doesn't exist) and shadekin_ch.dm and putting your code there.

Raeschen:
> As far as conflicts/comments go, the easiest approach is to put out code in a _ch.dm file where possible. For here, that would be something like creating a simple_mob_ch.dm (if it doesn't exist) and shadekin_ch.dm and putting your code there.

This has now been done :)

PR #2817, opened 2021-08-16
Adds step mechanics when the target is lying down

TheGreatKitsune:
Allows you to step on/over micros that are lying down.

Possibly needs some more testing to see if this code ever triggers at the same time as the Bump-based trigger, though it... really shouldn't.

### DIFF
--- a/code/modules/vore/eating/slipvore_ch.dm
+++ b/code/modules/vore/eating/slipvore_ch.dm
@@ -43,11 +43,22 @@
 			perform_the_nom(src,target,src,src.vore_selected,1)
 			target.slip_vore_in_progress = FALSE
 			target.is_slipping = FALSE
+			return
 		else if(can_be_slip_vored_by(target) && !src.slip_vore_in_progress && !target.slip_vore_in_progress) //Otherwise, if they can vore us, make it happen.
 			target.slip_vore_in_progress = TRUE //Make them stop slipping
 			perform_the_nom(target,src,target,target.vore_selected,1)
 			target.slip_vore_in_progress = FALSE
 			target.is_slipping = FALSE
+			return
+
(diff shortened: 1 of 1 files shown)


Darlantanis:
I don't see a benefit to this. Size mechanics are already abusable (and occasionally abused) in their current state, expanding on that seems unwise. 

Throwing returns into Crossed() for mob/living also seems particularly risky and likely to break code elsewhere considering Crossed() is a fundamental proc expected to get called frequently.

TheGreatKitsune:
The returns can be removed without causing issues, I believe. And if not, I can add a few more checks to see if a slipnom did occur to the rest of the step mechanic checks.
As for why, it was requested by Youth in the Discord, and I can see where it can be used in RP. And I would assume Youth's reason is so that certain ERP scenes can be played out more easily with game mechanics, though asking them directly might be better.
And if the potential to abuse it is great enough, there is the option of adding another preference to handle this, and possibly other situations where stun-locking or similar can occur.

CommodoreYouth:
Haha look, I got mentioned in a PR.
That's MY name.  I'm here too.  :D 
As for the benefits, we do have a sizable crowd of players who enjoy Size Mechanics and Foot/Pawplay RP, as well as a subgroup of players who like taking these things to more excessive levels.  Size Mechanics themselves are here because some people prefer size difference, and as for abuse, that's Ahelpable, and Pref-Locked; We have a functional Pref that prevents resizing, and another Pref that prevents stepping, which limits the targets of these Size Mechanics to those who actually have them enabled.
And if a certain interaction is not in someone's OOC Notes, then people should be LOOCing each other for specifics, or otherwise figuring out with each other the other's limits.  If they're not, then any cases of abuse cause notifications in the Admin logs anyways, correct?

Darlantanis:
I keep forgetting there's finally a pref-toggle for step mechanics.

PR #2819, opened 2021-08-17
Brings back teshari in-hand sprites

Raeschen:
This also changes the default holder of the teshari (e.g when tesh-grabbing) to the same holder as when picking up a micro.

Special behavior added for when grabbing a teshari specifically to show the inhand sprite as well as color it correctly without coloring the icon shown in the hand slot.


Raeschen:
_ch file'd my code from request from a different PR.

PR #2832, opened 2021-08-23
Vorepanel complexity reduction

TheGreatKitsune:
Reduces the 'complexity' (See https://eslint.org/docs/rules/complexity) of several parts of the VorePanel tgui interface to enable potentially adding more features. (And to also re-enable changing the visibility of individual fullness levels again.)

This is a draft, as I'm unsure if this should instead be proposed upstream, and if it would be accepted here, I would want to update the list of changes made to VorePanel before this is merged. (Unless that isn't as important as I assume it is.)


Darlantanis:
I should probably state this here instead of just on discord. This would be best sent upstream, changes this extensive will make future upstream tweaks to VorePanel.js difficult to resolve.

PR #2892, opened 2021-08-25
Ports Digitigrade legs

Raeschen:
image

This is a big one, and is the first draft of it here.

In some places I feel like I perform some rather questionable methods to achieve this? Good to have it reviewed, but locally it works.

Some more sprites still need to be added by me, like the species-specific hardsuit variants.

This is probably a thing which needs to be testmerged before any kind of real merge to help gather up some issues from the community. 

- Adds a character setup menu item; "Digitigrade?" only selectable if species is flagged to use digi legs
- Reskins original limbs to digitigrade appearance based on mob variable
- Adds a DNA datum Digitigrade variable for the purposes of transferring the data to resleevers/bioprinters
   - This does not affect DNA SE/UI strings at all. it is ONLY used for transferring data to reprinting machines
- Clothing with digitigrade sprites will show that sprite if the mob wearing it has a set digitigrade var AND if a replacement sprite exists - if no replacement sprite, it uses the original.

Known (and accepted) issues:

- Underwear sprites bad. If you want digitigrade, do not use underwear/socks.
- Some missing clothing sprites will be present (Engineering species-specific voidsuits, some shoes, ...), more will be added in the future.
- No body markings that cover the legs will work (Workaround fix below so you can keep your body markings, if you want digitigrade, expect no marking on the legs.)

Tested/fixed things:

- [x] Overlays on legs (Currently digilegs delete any overlays from themselves. Can add digi-specific overlays in future)
- [x] round-start spawn, latejoin spawn, resleeving spawn
- [x] Bioprinter
- [x] Clothing sprites
- [ ] Finish testing time period


Razgriz1032:
I'll get it test merged later in the week

Raeschen:
> I'll get it test merged later in the week

Thankyou!
I recommend the following approach:
> Nudge me when it's TM'd (I'll make a thread for feedback)
> TM it for a while (A week?)
> Any big/coding issues? Immediately un-TM and let me fix.
> If the TM time is up and there's no huge issues? (Missing sprites/small sprite issues can be fixed in later PRs) then merge.

that sound good?

code/modules/mob/living/carbon/human/species/station/station_ch.dm
```diff
@@ -38,4 +38,32 @@
 
 /datum/species/vox
 	spawn_flags = SPECIES_CAN_JOIN | SPECIES_IS_WHITELISTED | SPECIES_WHITELIST_SELECTABLE
-*/
\ No newline at end of file
+*/
+
+//Can use digitigrade flags
```
Rykka-Stormheart, line 43: Are Xenochimera not allowed to use Digitigrade legs, or do they go off the sprite choice under "sprite base"?

Raeschen: Added in commit below.

Raeschen:
Main bugs to fix:

- [ ] Fix leg overlays looking strange
- [ ] ...

Raeschen:
I learnt a few things from previous bugs. Somewhat changed how the preference variable is applied to the mob 

Also removed setting this on the species datum because that's the wrong place to have a preference toggle variable. - it's on the mob instead. 

Awaiting more bugs but the TM could potentially be updated here - might be good to re-check my code.

Raeschen:
I think you updated to the wrong commit!

Raeschen:
@Razgriz1032 
This is in a suitable state to update the TM with i'd say

Raeschen:
well i haven't had any huge complaints or bug reports with this over a few days. If maintainers are happy with it, i am, can be merged.

code/modules/organs/organ_icon_ch.dm
```diff
@@ -0,0 +1,33 @@
...
+		digitigrade = owner.digitigrade
+	else
+		digitigrade = dna.digitigrade
+
+	if( digitigrade && ( istype(src,/obj/item/organ/external/leg) || istype(src,/obj/item/organ/external/foot) ) )
+		var/IB = species.icobase
+		var/list/markingsTmp = markings
+
+		markings = null
+		species.icobase = species.icodigi
```
Verkister, line 25: This might be causing some issues and possibly play a part with client caches and stuff bugging out and crashing. All non-custom species use a single shared species datum between all mobs of the same species, and 2 users of the same species with different digi settings might have some nasty conflicts going on with their icon generation.

Raeschen: I undid any editing of the species datum by this code and instead copy some of the icon getting code from the original to be performed in a different way specifically for digitigrade here.

code/modules/client/preference_setup/general/03_body_ch.dm
```diff
@@ -0,0 +1,28 @@
+// Chomp additions to character load/save logic
+// /datum/category_item/player_setup_item/general/body/content cannot be overridden here easily as the proc creates a list of html data
+// perhaps one could get that list here, and sort through sections to rewrite it's content...? that seems inefficent having to loop through the list.
+
+
+/datum/category_item/player_setup_item/general/body/copy_to_mob(var/mob/living/carbon/human/character)
+	. = ..()
+	if(character.species.digi_allowed)
+		character.digitigrade = pref.digitigrade
```
Verkister, line 9: I feel like this spot could use a sanity check to make sure the pref has this option saved, otherwise it slaps that toggle var into a null due to old savefile lacking the value. Also there seems to be a lot of defined 0 vars in this PR that should probably use FALSE instead for consistency.

Raeschen: I wanted to keep digitigrade itself as a value so i could allow for future different leg styles but there may be a better way to do this (?) - i will however add the sanity check and check elsewhere for this

PR #2893, opened 2021-08-25
Cleanup subsystem

cadyn:
Mmm, I think it's new subsystem hour.

So, during Serdy's recent event, which was great fun by the way, we noticed that as ammo casings were piling up on the ground it was causing great amounts of lag for the server. It was confirmed to be the source of the lag when the admins used SDQL2 to delete all the empty casings, which brought the time dilation from about 150 to 50.

What's my solution for this? Well I figure we can probably gain some performance in general from deleting things that aren't needed. So far, I only have dead simple mobs, dead monkeys, and expended casings being added, but more stuff can be added to it without much effort.

The subsystem has an urgency feature which takes a measure of how long the queue is and how laggy the server is to decide how long it should wait before deleting stuff. For casing it's currently between 5-10 minutes, for mobs it's 15-30 minutes.

We also have some special cases for the subsystem, but these are ignored if the urgency is too high and the system goes into "Panic mode":

1. All atoms have their deletion timer paused if they're in not on a turf (i.e. inside a locker, or storage.)

2. Mobs have their deletion timer reset if they change turfs.

3. Mobs are removed from the queue if they are alive again during the check

4. All atoms are removed from the queue if they were unable to be deleted after 120 minutes of trying.

I also made an effort to leave comments all over the subsystem, so it's easy to understand the code, or hopefully it is at least. I'm taking feedback for adding more special cases as well for more things to have deleted by the subsystem.

EDIT 1:
Lots of small tweaks and cleanup. Thanks to Aronai for the feedback, and pointing out a really big oversight I made.

In addition, magazines have been added to the system. If a magazine is removed from a gun, and it has no bullets stored in it, it will be added.
The new special condition for magazines is:
5. Magazines that are in the subsystem and have bullets added to them will be removed from it.


code/controllers/subsystems/cleanup_ch.dm
```diff
@@ -0,0 +1,165 @@
...
+
+/datum/controller/subsystem/cleanup/proc/get_points()
+	var/temp_points = processing.len
+	for(var/mob/mob_entry in processing)
+		if(istype(mob_entry))
+			temp_points += 2 //Mobs are worth 3x as much as atoms, since they're counted once in the first definition for temp_points, and twice here.
+	for(var/mob/living/carbon/carbon_entry in processing)
+		if(istype(carbon_entry))
+			temp_points += 2 //Carbons are worth 5x as much as atoms, since they're counted once in the first definition for temp_points, twice for being a mob, and twice for carbon.
+	return temp_points
```
Arokha, line 40: These istypes don't make any sense, because the for-loop is a typed for-loop, which does istype already. The istype() inside the loop on the first line will always be true, so is pointless.

You could avoid the istype() over all these items entirely if you kept the lists separate, then you could just check the len. Like if you had a proc to add to 'any' list, the subsystem could handle which list it should go into (or the thing itself can, I dunno).

code/controllers/subsystems/cleanup_ch.dm
```diff
@@ -0,0 +1,165 @@
...
+	flags = SS_BACKGROUND|SS_NO_INIT
+	wait = 1200 //Fires every 2 minutes.
+
+	var/stat_tag = "Cl" //Used for logging
+	var/urgency //How desperate are we to delete stuff?
+	var/points //How much complexity is in our queue?
+	var/list/processing = list()
+	var/list/currentrun = list()
+
+	var/datum/cleanupentry/current_thing
```
Arokha, line 18: Just curious, is this only used for the Recover()?

code/controllers/subsystems/cleanup_ch.dm
```diff
@@ -0,0 +1,165 @@
...
+	var/time_entered_objective
+
+/datum/cleanupentry/New(atom/cl_atom)
+	entry_atom = cl_atom
+	data = list()
+	time_entered = world.time
+	time_entered_objective = world.time
+
+/datum/cleanupentry/proc/Check()
+	return entry_atom.CleanupCheck(src,world.time - time_entered)
```
Arokha, line 104: See other comment.

code/controllers/subsystems/cleanup_ch.dm
```diff
@@ -0,0 +1,165 @@
...
+		currentrun = processing.Copy()
+		SScleanup.urgency = get_urgency()
+	var/list/current_run = currentrun
+	var/current_check = 0
+	while(current_run.len)
+		current_thing = current_run[current_run.len]
+		current_run.len--
+		if(QDELETED(current_thing) || QDELETED(current_thing.entry_atom))
+			processing -= current_thing
+		current_check = current_thing.Check()
```
Arokha, line 70: This is probably the biggest problem I see here. QDELETED returns true if the arg is null as well (because maybe it's already been GC'd).

In the case where the datum is null, this will just runtime and break the subsystem since you call Check() on it in the next line.

In the case where the atom is null, you will generate a runtime inside Check() below, which assumes the atom exists and calls CleanupCheck on it. This is much more likely, and I would imagine you'd encounter this probably somewhat often, maybe even once a shift: imagine a dead mob getting butchered, which would seemingly always trigger this.

You should probably `continue` inside this, after removing it from processing.

Arokha (commented):
I like this idea!

code/controllers/subsystems/cleanup_ch.dm
cadyn: This was me being a 4head, and a pretty big oversight lol. If I add continue here, I will probably want to also move the tick check to the beginning of the loop, so hmm, I will have to think about this. The alternative would just be adding an else statement and putting the rest of the code under that, and in the end it doesn't really matter other than for cleanliness.

code/controllers/subsystems/cleanup_ch.dm
cadyn: I guess this is mostly just habit from dealing with this codebase and wanting having redundancy checks everywhere to make sure shit doesn't randomly break everything. I'll get rid of these.

code/controllers/subsystems/cleanup_ch.dm
```diff
@@ -0,0 +1,165 @@
+#define FIRE_PRIORITY_CLEANUP 20
+#define MAX_CLEANUP_QUEUE_TIME 120 //Max time in queue before we give up.
+#define URGENCY_PANIC_MODE 0.75 //Max urgency before we start to skip checks
+#define SMOOTHSTEP(variable) (max(min(6*(variable**5) - 15*(variable**4) + 10*(variable**3),1),0))
+//This is the cleanup subsystem. Its responsibility is to clean up unneeded atoms such as empty bullet casings and dead mobs in an effort to boost performance.
+SUBSYSTEM_DEF(cleanup)
+	name = "Cleanup"
+	priority = FIRE_PRIORITY_CLEANUP
+	flags = SS_BACKGROUND|SS_NO_INIT
+	wait = 1200 //Fires every 2 minutes.
```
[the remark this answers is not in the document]
cadyn: Probably lol

cadyn:
Mostly for my own 4head brain so I can remember:
TODO:
1. resolve Aro's feedback.
2. Add more items to be put in queue.
3. Potentially add a verb for people to remove items from the queue (Potential for this to be abused, make sure this is a good idea, and add some warning that explains what the verb actually does before the user actually goes through with it.)

code/controllers/subsystems/cleanup_ch.dm
cadyn: If we're talking about current_thing, then it's used for Recover and in the actual fire proc, I believe all of these were just copied from other subsystems though to sort of follow the standard for them.

code/controllers/subsystems/cleanup_ch.dm
Arokha: > This was me being a 4head, and a pretty big oversight lol. If I add continue here, I will probably want to also move the tick check to the beginning of the loop, so hmm, I will have to think about this. The alternative would just be adding an else statement and putting the rest of the code under that, and in the end it doesn't really matter other than for cleanliness.

I'm not sure I'd worry too much about that honestly. If you can think of an elegant way to do it, sure. But the 'worst case' for this is 'the entire processing list is QDELETED(x)=TRUE items' (so every single item processed goes into the inside of this if() statement), and I have trouble imagining that clearing the list out will be that expensive. 150 shell casings or whatever, aka removing 150 nulls from this list (just imagining that the shell casings were all destroyed by an explosion or something so they're all deleted and now null), should be baaaasically instant.

I guess if it gets into tracking like 100k+ things then I'd worry about the performance in the off-chance that you end up with 'large' runs of nulls in the list.

cadyn: > > This was me being a 4head, and a pretty big oversight lol. If I add continue here, I will probably want to also move the tick check to the beginning of the loop, so hmm, I will have to think about this. The alternative would just be adding an else statement and putting the rest of the code under that, and in the end it doesn't really matter other than for cleanliness.
> 
> I'm not sure I'd worry too much about that honestly. If you can think of an elegant way to do it, sure. But the 'worst case' for this is 'the entire processing list is QDELETED(x)=TRUE items' (so every single item processed goes into the inside of this if() statement), and I have trouble imagining that clearing the list out will be that expensive. 150 shell casings or whatever, aka removing 150 nulls from this list (just imagining that the shell casings were all destroyed by an explosion or something so they're all deleted and now null), should be baaaasically instant.
> 
> I guess if it gets into tracking like 100k+ things then I'd worry about the performance in the off-chance that you end up with 'large' runs of nulls in the list.
> 
> EDIT: I definitely wouldn't move the tickcheck to the top though, since that could result in the subsystem not doing ANY work during superlag times where the MC tells the subsystem it's out of time already as it's starting. It should do at least 1 item per run, so if everything is going to hell, it can at least slowly work on it.
> 
> EDIT EDIT: Actually yeah, wait, if everything is going to hell and lagging, don't you want to NOT stop working if the MC says it's real laggy? The whole TICK CHECK stuff is to give up working to allow other things time to do their jobs when it's laggy times, but like. If the server is turbo-lagged don't you want to ESPECIALLY do work with this subsystem then? HMM.

Hmm. I'll think about this, I could keep it at the end and then use an urgency check to decide whether or not to keep going, but even then it will probably only be up to like 5 entries per tick, because it should still keep firing every tick afterwards until it gets through the queue.

Razgriz1032:
I'll get this test merged later in the week

cadyn:
This is ready to be fullmerged

code/controllers/subsystems/cleanup_ch.dm
```diff
@@ -0,0 +1,229 @@
...
+/datum/controller/subsystem/cleanup/proc/get_points()
+	var/temp_points = processing.len
+	for(var/mob/mob_entry in processing)
+		temp_points += 2 //Mobs are worth 3x as much as atoms, since they're counted once in the first definition for temp_points, and twice here.
+	for(var/mob/living/carbon/carbon_entry in processing)
+		temp_points += 2 //Carbons are worth 5x as much as atoms, since they're counted once in the first definition for temp_points, twice for being a mob, and twice for carbon.
+	return temp_points
+
+/proc/cleanup_adjusted_points(points)
+	return (log(points)-2.5)/3
```
Verkister, line 48: This spot runtimes when points is 0

cadyn: oopsie poopsie, I really should have predicted this -w-

Cheers.

Verkister:
I have a suggestion at least for the mob cleanups to have it skip the deletion if there's a player within the view range. Kinda jarring ingame to see the corpses pop out of existence right in front of your character.

PR #2921, opened 2021-08-27
Adds some recyclers to the disposal room

lbnesquik:
## About this pull request
This pull request adds recyclers, a sorter and a sheet stamper to disposal. 
Those recycle most of what you'll be throwing down disposals. I've had to rework the disposal area for this a little bit.

## Why is it good for the game
Fancier trash area that doesn't just _**space**_ everything they are done using.
Makes it worth going over there for some recycled material on occasions.

## How does it look 

image


cadyn:
Wait how would people actually be able to get the sheets out of the machine?

lbnesquik:
> 
> 
> Wait how would people actually be able to get the sheets out of the machine?

Open the door and get them.

Razgriz1032:
Unless the new machines block air movement, you should put the plastic flaps back. Those block air movement and won't suck the whole system.

Kazzc:
Anyways, now that I mapped something more cohesive and tested the recycling machines. . . I can say that this PR is now irrelevant. Even more, as the machines can't recycle everything. . . So this set-up will literally clog and not send the trash into space. Bad design and poorly tested.
Please reference here. https://github.com/CHOMPStation2/CHOMPStation2/pull/2942

PR #2922, opened 2021-08-27
Makes the central ring 3 doors wide instead of simply 2 seperated by glass

lbnesquik:
## About this pull request
This makes central ring more pleasant because you don't have to aim as hard for the doors. Yes this is annoying.
Also move some maint areas around near EVA.

Less bumping your face into glass.


cadyn:
Dunno about this one. Personally, I think the glass in the middle is more aesthetically pleasing than three doors. Make a fancy single door that takes up three tiles, though? Then we'll be talking.

WTFIsmyname:
Fuck windows. All my homies hate windows

WTFIsmyname:
Fuck windows. All my homies hate windows

Razgriz1032:
Idk about this one, chief.

Kazzc:
Eh, generally it's not a good idea to have two doors next to each other, even less 3 doors all next to each other.
Since the hallways are 3 blocks wide it's aesthetically pleasing to leave it like this, unless we have 3 block-wide door and atmos doors, having 3 doors next to each other just looks silly. (Tho I can't imagine a 3 block wide door. . . will make atmos issues even worst if there was a breach)

Just by how it looks and how negligible this change is I will say no.

lbnesquik:
image

lbnesquik:
> 
> 
> Eh, generally it's not a good idea to have two doors next to each other, even less 3 doors all next to each other.

It works just fine on delta, tether or northern star. 

> Just by how it looks and how negligible this change is I will say no.

I am absolutely willing to make this change extend to every hallway if that makes it not negligible.

lbnesquik:
Ah, whatever. This is not going to get merged.

PR #2937, opened 2021-08-30
Altevian & Squeakish Language and ATC Addition.

Enzo-Leon:
With approval from @Darlantanis on the lore, and @BlackMajor with the go-ahead to handle the first set of code changes. This is the start of the altevian species lore and language addition! This adds a language that AI/borgs and others will be able to understand with no real restriction, and adds their ships and other data into the ATC. While a small addon for them, this will start their integration into becoming an official species for others to easily utilize at their leisure.


Darlantanis:
Any edits to upstream files (typically anything that's not _ch.dm or _yw.dm) will need //CHOMPedit comments (otherwise I'll end up overwriting the changes if a mirror conflicts).

Ideally anything that can be put in a _ch file should be put in a _ch file. This helps Mirrorbot not have an aneurysm when code conflicts.

PR #2942, opened 2021-09-03
Adds a recycling room for Cargo

Kazzc:
After testing it for quite a while, there are some minor issues with the recycling machines.
-Will not recycle trash.
-Will recycle some specific items, but if there is to much trash between the things it can recycle and what it can't, it will not recycle anything. (It will need to manually be unclogged)
-They do not consume power. (Tho I think they should)

I like the idea and I liked what mapped in for it, but this is essentially an extra step for Cargo to sort trash out. But I made sure to make it optional so if cargo players do not want to deal with recycling, then they now have an extra step of clicking the lever 3 times to send it all to disposals. (Please forgive me for making you click 3 more times now)

Feedback will be appreciated, but I am indifferent if we should add this or not. . . I think the recycling machines need more code work to have it recycle more things and/or give some other basic materials than just steel. (Aka, cardoard, plastic, glass and so on from trash itself)

*SouthernCross Deck-2
-Expanded cargo and turned its small warehouse in to a hybrid warehouse/recycling facility.
-Added warning signs on Cargo.
-Removed a trash-pile that was cluttering Disposals
-Added a few Library and Gym signs on the halls by the civilian area.
-All the loot piles, trash piles, resources, and items were shifted near its maintenance area. No resources were deleted.


Kazzc:
image
image

PR #2951, opened 2021-09-08
Fix the DNAModifier so that genetics can be a thing again.

TheGreatKitsune:
Only the change on line 330 is required, forgot to revert the other two changes made before committing. I can revert the unnecessary changes if desired though.


TheGreatKitsune:
FixSoon

Darlantanis:
I recommend sending this upstream.

PR #2952, opened 2021-09-09
Remove the audio clip from the initial solar storm alert.

TheGreatKitsune:
Because seriously, it makes me always double check the announcement whenever I hear it. (It'll still get the ping sound by default, which should be good usually.)


Rykka-Stormheart:
The audio clip exists for a reason. Solar storms still carry radiation, and as such, it's far better if they're read out. 

Unless you have a replacement audio file, this is a positive feature removal for what seems like a personal pet peeve?

How does this benefit the station?

TheGreatKitsune:
Hmm... That is true. I'll let this get closed then, and look into what would need to be done for a replacement audio file.

PR #2955, opened 2021-09-09
Enables hands for raccoons

Verkister:
Go, do a crime.


Darlantanis:
This is a terrible idea.

PR #2960, opened 2021-09-13
Revert "Fix the DNAModifier so that genetics can be a thing again." for test merge

Razgriz1032:
Reverts CHOMPStation2/CHOMPStation2#2951

In wake of a lot of reports about TGUI breaking, we are going to try to revert this and test merge it.


Razgriz1032:
nvm not testing yet, going to try #2961

PR #2962, opened 2021-09-13
Change energy cell sprites and add new giga cell

Razgriz1032:
Changes super and hyper cells to utilized some new and unused sprites. Adds a new energy cell past the hyper cell to also take advantage of new sprites. Requires precursor tech 1 and power tech 5


Razgriz1032:
Tested. Pretty
image

PR #2968, opened 2021-09-16
Southern Cross STS shuttles reach max

Darlantanis:
After a year and a half of nonstop grinding, Shuttles 1 and 2 have finally reached level 99 Sailing. As a result of finally maxing, shuttle transport speed is now 4 times faster. Transit between station and planet now takes 1 minute. As a result of this new speed, shuttles now wait only 1 minute for passengers to board.

WARNING: Attempting to pilot a level 99 vessel without maxed skill may result in spontaneous combustion, public lynching, code nerfs, passive-aggressive complaints, and/or death. Please drink responsibly.


Kazzc:
vibe

PR #2974, opened 2021-09-17
Traits balancing changes

cadyn:
So, I kinda realized that traits needed to be fixed a bit after I found myself with a few grievances about the balance of certain traits. I'm 100% open to feedback on these changes, but I'm feeling pretty confident about the changes I want to make. This is isn't entirely a nerf, but will likely feel like it, because I'm specifically going after traits that I consider to be "free points" and that I know lots of people use as such.

Changes and why:
Negative traits:
High pain intolerance: Now 100% more pain instead of 50% to push it over the boundary of actually making a significant and noticeable difference in gameplay.

Conductive: Now -1 point instead of -2 points and 75% instead of 25% increase. Even still this is going to make a pretty negligible difference in gameplay, hence -1 point.

Lightweight: Now -1 point instead of -2. Having lightweight trait will not change the outcome of a combat situation 99/100 times. Only way that somebody could take advantage is if they even knew, it occured to them, and they could get close enough. All this trait really does is provide a minor annoyance.

Less blood: Now -2 points instead of -3. Less blood is impactful, but not that impactful. When I was originally making this I overestimated how much of a part blood plays.

Less blood extreme: Now -3 instead of -5. Again with above, the thing is despite being 40% the amount of blood, it only affects you if you start bleeding and can be fairly easily combatted.

Positive:

Traceur: Now 1 point instead of 2. It's my opinion that all "fluff" traits, and being able to flip more reliably is a fluff trait in my book, should be 1 point.

Weaver: Now 1 point instead of 2. Same as above, this constitutes a fluff trait, therefore only 1 point.

Rad resistance: Now 1 point instead of 2. Having to deal with radiation is pretty rare, and this doesn't provide that much of a buff against it.

Extreme rad resistance: Now 2 points instead of 4. Same with above, although this one does provide a more substantial buff, radiation remains something that just doesn't impact gameplay often enough to justify 4 points.

More blood: Now 2 points instead of 3. Same argument with negative blood traits, they just don't impact the game enough to have this much cost.

More blood extreme: Now 4 points instead of 6. I would have gone with 3, but this trait does give you a lot of blood, making you almost immune to blood loss. Reason for decreasing it is similar to decreasing negative blood traits.

Grappling expert: Now ~2x harder to escape instead of ~4-6x, this is just something that I didn't test enough originally. I discovered only on the server that it is pretty much if not entirely impossible to escape a level 3 grab with this trait.

Giant mouth: Now 2 points instead of 3. This is another fluff trait, really. I would leave it at 1 point, but wanting it to still be more expensive than big mouth, I'll settle it at 2.

Absorbent: Now 1 point instead of 2. This is yet another fluff trait. Neither nutrition nor dirt play a major part in gameplay.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/negative.dm
@@ -106,24 +106,26 @@
 	cost = -1
 	var_changes = list("pain_mod" = 1.25)
 
+//CHOMPEdit Begin
 /datum/trait/negative/pain_intolerance_advanced
 	name = "High Pain Intolerance"
-	desc = "You are highly sensitive to all sources of pain, and experience 50% more pain."
+	desc = "You are highly sensitive to all sources of pain, and experience 100% more pain."
 	cost = -2
-	var_changes = list("pain_mod" = 1.5) //this makes you extremely vulnerable to most sources of pain, a stunbaton bop or shotgun beanbag will do around 90 agony, almost enough to drop you in one hit
+	var_changes = list("pain_mod" = 2) //most things should down you in 1 or 2 hits
+//CHOMPEdit end
 //YW ADDITIONS END
 
 /datum/trait/negative/conductive
 	name = "Conductive"
-	desc = "Increases your susceptibility to electric shocks by 25%"
-	cost = -2 //CHOMPEdit
-	var_changes = list("siemens_coefficient" = 1.25) //This makes you a lot weaker to tasers.
+	desc = "Increases your susceptibility to electric shocks by 75%"
+	cost = -1 //CHOMPEdit What the fuck this is not worth 2 points.
+	var_changes = list("siemens_coefficient" = 1.75) //This makes you a lot weaker to tasers.
 
-/datum/trait/negative/conductive_plus
+/*/datum/trait/negative/conductive_plus //CHOMP Removal No we are not giving people a free 3 points with this trait.
 	name = "Conductive, Major"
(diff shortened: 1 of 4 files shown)


Razgriz1032:
Haste should cost more or be less effective.

cadyn:
> 
> 
> Haste should cost more or be less effective.

Honestly, I'm not in agreeance with this. If it was equal to teshari haste then I would say yeah, but it isn't. I really wish speeds were in percentages like literally everything else but it's not so ree. haste is -0.5 slowdown, and teshari have -1 slowdown

Cyrelius:
Haste is extremely powerful, I'd like a lesser and cheaper version of it.

Cyrelius:
I would also argue that the fluff ones should take a trait slot, but not cost any points at all.

cadyn:
> 
> 
> I would also argue that the fluff ones should take a trait slot, but not cost any points at all.

Moved weaver to neutral, but the other two feel more fitting with some cost. Just having them take up a slot doesn't feel like enough of a cost when players have 8 available slots.

cadyn:
> 
> 
> Haste should cost more or be less effective.

After sleeping on it, and running around with haste for a bit myself, I've decided to bump the price up to 4, because it is admittedly very impactful on gameplay, but at the same time the speed boost isn't crazy. I'd guess it's less than a 50% increase in speed.

Razgriz1032:
Looking good so far for the rest of the choices

Cyrelius:
Honestly, big mouth and giant mouth don't deserve to be positive traits at all.  They're about as meaningful as spice tolerance.

PR #2976, opened 2021-09-17
Lightens up digilegs and fixes leopardmander runtime

Verkister:
Adjusts the digilegs to not be several shades darker than the rest of the body.

Also fixes leopardmander runtime.


Rykka-Stormheart:
Flamethrower fluid is throwing runtimes on a Sif POI.

May want to check that something isn't spawning the effect, iirc.

Razgriz1032:
Leopardmander runtime was already fixed in your compiler PR

Verkister:
Nah, the leopardmander runtime was caused by a missing chomp edit in the upstream file that had replaced ours in the dme, and was an ingame runtime that wouldn't happen during compile anyway.

PR #2977, opened 2021-09-18
Digi markings and sprite improvement

cadyn:
Self explanatory, added digitigrade specific markings which only show up for digi legs, also improved the digitigrade sprite. Have some visual aids. This took me way too long to make, and I know my sprite work isn't the best, but I just wanted to nail out some basic markings so people could actually customize their digi legs.
beforeafter
allmarkings


Razgriz1032:
This will conflict with 2976

Verkister:
Looks better than mine anyway. The only other thing in my pr was swapping leopardmander back to _ch file rather than the upstream one that was runtiming due to missing chomp edit.

Darlantanis:
Why are there VOREstation comments in a _ch file?

cadyn:
> 
> 
> Why are there VOREstation comments in a _ch file?

It's because the code was copied from the original file, which was organ_icons.dm IIRC

cadyn:
I've added a comment stating exactly where the code is copied from and why VOREStation edits are kept in (namely that I feel like it's worth keeping the code as 1:1 as possible and keeping track of which code is from virgo). If you really want the vorestation comments removed, though, I can, as it's not a big deal.

PR #3026, opened 2021-09-24
[MIRROR] Plains POI Audit

CHOMPStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/11606
--------------------
Works over every active Plains POI and adjusts their turfs to use new v3b-specific turfs (added in `plains_turfs.dmm` using `VIRGO3B_TURF_CREATE`). Some minor adjustments were made to them where necessary, but contents/etc. are otherwise unchanged.

Pairs with #11605 to (hopefully) squash active edge errors coming from the plains. Will cut down on weird cases of buildings starting full of breathable air despite being partially or completely exposed to the V3b atmosphere.

Also removes a couple of unused duplicates that were simply just "POI X but with no areas".


Kazzc:
Ignore this one, replacing a lot of tiles with new tiles from Virgo that we don't have.

PR #3039, opened 2021-09-27
Add hovering to shadekin when phased

TheGreatKitsune:
Fixes leaving footprints in snow, burning from lava, slowdown in water, and probably falling off cliffs while phased out.


Kazzc:
Thank you very much!, I'll have one of my devs review the code before I merge.

PR #3051, opened 2021-09-28
add a method of limiting areas shadekin can phase in/out of

TheGreatKitsune:
Currently only the vault has this enabled, but this could be applied other places, including live during rounds via variable editing.

### DIFF
--- a/code/game/area/Space Station 13 areas.dm
+++ b/code/game/area/Space Station 13 areas.dm
@@ -1819,6 +1819,7 @@ NOTE: there are two lists of areas in the end of this file: centcom and station 
 	name = "\improper Vault"
 	icon_state = "nuke_storage"
 	ambience = AMBIENCE_HIGHSEC
+	limit_shadekin_phasing = TRUE // CHOMPEdit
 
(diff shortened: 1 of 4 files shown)


ExabyteDiscord:
Maybe add the armory to this too? You could probably also simply use the variable some areas already have: The armory won't allow bluespace interference such as teleporting

TheGreatKitsune:
I could add the armory pretty easily, though... placing the restriction on only the vault is what Kassc requested from me so far. It is separate from any existing variables because, for one because I didn't notice any other variables that could serve this purpose, and for another because it sounds like the list that variable would effect and what is currently slated to be affected by this would be different.

Kazzc:
This is pretty good, I'll have it checked by a dev and get this merged.

TheGreatKitsune:
Honestly if there was another area or two that I think could use this limitation, it would be the R&D research servers, and telecomms. Both are... not often accessed by players (And even when there are people in them they aren't hanging around for long as both are at least partially kept quite cold), and can cause fairly annoying or even major issues if messed with.

PR #3065, opened 2021-10-07
Fix size gun settings in station/outpost dorms

TheGreatKitsune:
Just for the dorms, not going to touch any other areas at the moment. (But for context, walking out of these areas will return you to 25% or 200% if you were set below/above those limits by non-admin means.)


LordSaladin:
hyper macro oppai

PR #3072, opened 2021-10-13
Fix shuttle flooring generating insane lag

Razgriz1032:
/turf/simulated/shuttle/proc/underlay_update() had a "return 0" for it. Return 0 would add a blank value to a LIST and there was no GC for it. Nothing else seems adversely affected by this change to remove the 0 and just have it return.


Psik9941:
Will have to keep an eye out for other return 0 bits in code, especially on stuff that's called repeatedly that could cause lag buildup and crashes.

PR #3083, opened 2021-10-18
Guilmon horns and markings - Skyrat markings 

Echopsychotic:
markings from skyrat this adds guilmon Horns and the markings. This is complete with digitigrade leg markings as well.


JoltDerg:
Who originally made these? I'm trying to find the proper creator so that I can port these to RS and it's fucking me up

PR #3186, opened 2021-11-10
Arachnid Slammer

Shadowfire117:
Made a new drink derived from stimulants, energy drink and hearty ale! Good to warm the heart and soul when mining and slaying spiders!

Also made a recipe for making some absolute loko energy drinks!


LordSaladin:
ROCK AND STONE!

PR #3191, opened 2021-11-14
quick mapping tool pr

BlackMajor:
Been trying to figure out how to get lighting to work when you late-load a new z-level.
This is as close as I can get.

Honestly just made this PR so map lighting works during event maps.


BlackMajor:
Made this in like half a second for someone's event, I'll probably come back and refine it

PR #3199, opened 2021-11-23
Map Adjustments (Good bye Beepsky)

harry3124:
Deck 2 - Beepsky and nav beacon deleted. Food vendor placed on the brig.
Deck 7 - Randomous DS helmet deleted. Cannot find relating lores to support why helmet is at there.
Deck 8 - Light item move between the sword and P90 location within armory. New button, separating the blast doors leading to each weapons and mech armory for admin/event managers.

 - Reason of why Beepsky has been removed : It caused too many issues. Examples being...
   - Beepsky left the Deck 2 by following someone on the teleport. 20 minutes or more the peoples suffered from intense lag as no admins were able to help at the time.
   - On peculiar event, Beepsky was able to track and beat the antagonist to death who was within the vent pipes.
   - Often bothers the event mobs, resulting sour, unintended relationship between crews and the event mobs.
   - It is able to open the emergency shutters for itself trying to move around on weed infestation.


Razgriz1032:
`Beepsky and nav beacon deleted`
The nav beacons do not need to be deleted. They don't add any load to the server and would be required again if beepsky were to come back.


`Deck 7 - Randomous DS helmet deleted. Cannot find relating lores to support why helmet is at there.`
Can you find lore to support _why not_ the helmet is there? There's wrecked mauler, phazon and dark gygax there, some heavy duty hardware. It's an alien ship, could be any number of reasons why it would be there. It's something special to find.

`Deck 8 - Light item move between the sword and P90 location within armory. New button, separating the blast doors leading to each weapons and mech armory for admin/event managers.`
Buttons are, I guess, handy. But normally we just delete the doors if it's needed. Also being able to control what weapons are available is good because the northern bit of the armory is actually Asset Protection gear and not normally granted to ERT, and a button would just allow free access.

harry3124:
> `The nav beacons do not need to be deleted. They don't add any load to the server and would be required again if beepsky were to come back.`
 Very well, I'll get those restored.
> `Can you find lore to support _why not_ the helmet is there? There's wrecked mauler, phazon and dark gygax there, some heavy duty hardware. It's an alien ship, could be any number of reasons why it would be there. It's something special to find.`
 Simply putting only the phazon, dark gygax along with little bit of rubbish did not really convinced why the DS helmet is at the first place. No notes to convince the peoples why the ship is there, or which sides the crew on the ship was been. A DS helmet within an _alien ship_ even further confuses the ship's faction.
> `Buttons are, I guess, handy. But normally we just delete the doors if it's needed. Also being able to control what weapons are available is good because the northern bit of the armory is actually Asset Protection gear and not normally granted to ERT, and a button would just allow free access.`
The said buttons are behind invincible windows, walls and centcom/ERT level access dozens of doors, and from ERT room to the office, the peoples also need to bypass two blast doors. If said 'we' are representing 'entire staff team' are simply deleting the doors, the other button that was been there before; Being the button that opens **both doors** which is more insecure, should've been rather deleted it if the security is a problem like you are pointing out. Adding those buttons and adjusting the blast door ID code simply allows the blast doors to be opened separately, allowing the admins/event managers to open the singular area rather than opening both places.
However, if still makes you uncomfortable, I shall change the access of the door being much more higher, allowing only centcom officers able to interact; Even when it is already _only_ centcom officers can interact and both get an access path towards the gears.

Kazzc:
hmmm. . . I am still doing a side project on reworking all of CentCom into it being more appealing and somewhat more functional. But these changes are good, as much as I don't want to get rid of Beepsky it causes issues here and there.

PR #3201, opened 2021-11-25
reverse vorespawn

BlackMajor:
Does exactly what you might expect.

While the vorespawn PR enabled prey to spawn directly from the lobby into a pred's gut, this PR does the reverse. Allowing a predator to spawn right ontop of a prey from the lobby, voring them in the process.

Someone suggested this to me so I figured I'd make it a reality.


Arctaisia:
I am specifically making a synth to do Titanfall jokes with now, and no one can stop me

_Hhh, still really good though,,,_

PR #3204, opened 2021-11-26
Probably fixes a little autotransfer bug

Verkister:
Probably fixes a bug where a prey-ghost queued for single-slot autotransfer vanishes from existence, leaving their position on the queue as a null, which neither has the autotransferable var or a thing to remove from that now-occupied-by-a-null list, which basically broke both parts of the thing by having an already "full" queue that wasn't clearing itself and also a blank un-skipped null in the contents list bumping up the tally and hitting the max amount to skip further handling of the contents. Or something like that idk. Thinkmeat too depleted to ~~(legit so depleted i completely forgot the word that was upposed to go here smh. it started with c tho i think but even google ain't helping rn lmao)~~ **elaborate** further.

### DIFF
--- a/code/modules/vore/eating/bellymodes_vr.dm
+++ b/code/modules/vore/eating/bellymodes_vr.dm
@@ -39,7 +39,7 @@
 					continue
 			tally++
 			M.belly_cycles++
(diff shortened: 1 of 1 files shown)


Darlantanis:
smacking CI

PR #3205, opened 2021-11-26
Fixes dirt vulnerability

Verkister:
Fixes dirt vulnerability trait being completely bonkers broken and causing slips on every single step regardless of whether the floors were even clean or not and also makes it not affect walk intent.


Darlantanis:
Smacking CI

PR #3206, opened 2021-12-02
Add some missing modules to the medihound

TheGreatKitsune:
Adds the bioregenerator, organ gripper, and a dropper to the medihound, which should be the last few modules they were missing.
The bioregenerator because it's provided in the operating rooms by default, so... would make sense that the medihound also gets it.
Organ gripper because without it there is literally no way for a medihound to replace lost limbs. The research hound, on the other hand, has the robotics organ gripper, so...
Dropper because otherwise you have no way of fixing necrosis, besides amputation.


Darlantanis:
Smacking CI

PR #3231, opened 2021-12-03
[MIRROR] operation RAD APES

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/11831
--------------------
- crusher gauntlets no longer have power draw when active
- phoron bores now update their parts on every attackby()
- adds the unobtainable Radiation Absorption Device and Atmospheric Protection Enhancement Suite
- the RAD provides full radiation protection at the cost of 250 power per tick
- the advanced RAD does the same thing as the RAD, at the cost of 50 power per tick
- the APE provides infinite protection against atmospheric heat and pressure, at the cost of 250 power per tick


Darlantanis:
remirror

PR #3234, opened 2021-12-03
[MIRROR] Some scoopy holder nom tweaks

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/11908
--------------------
Moves the part of the code handling the holder things on scooped up snacking to where the actual snacking happens, and also clears and deletes the holder afterward.
Also contains checks to make sure any undeclared stowaways in holder contents also get eaten (or dumped on the ground if that somehow fails) instead of getting deleted with the holder. Probably fixes a lag or garbage collection jank bug that sometimes makes the holder dump the prey back on ground after they've already been eaten, and also fixes the thing with the holder item still lingering in hand for the entire process cycle duration after the prey has already been eaten.
Also makes holders use Entered/Exited to do their things with the contained mob to make them function more reliably. (now makes sure only one mob can enter a holder, and now also clears the held_mob var and returns the held mob's transform appearance back to the mob without requiring the ground dump proc, which probably also fixes the thing with eaten held mobs sometimes getting released with messed up scaling etc)
Tested working nicely.


Darlantanis:
remirror

Darlantanis:
Implemented by #3198

PR #3249, opened 2021-12-03
[MIRROR] [MIRROR] Fixes rocket launcher

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/11886
--------------------
Original PR: https://github.com/PolarisSS13/Polaris/pull/8344
--------------------
The launcher type guns use throw_at instead of the fancy bullet projectile features, so the fancy bullet projectile mechanics were unable to make things go boom by having several dead-end returns happen before the projectiles could even reach the on_hit, which already bonks into a return in the ..()'d parent proc anyway.


Darlantanis:
Implemented by #3188

PR #3260, opened 2021-12-04
[MIRROR] Remove pkg-config:i386 from build

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/11927
--------------------
Ubuntu (or perhaps GitHub's image of Ubuntu 20.04) messed up the deps for pkg-config:i386 (namely libglib2.0.0:i386). pkg-config:i386 doesn't appear to actually be required for rust though, so unsure why it was in this CI.


CHOMPStation2:
Already done

PR #3284, opened 2021-12-06
[MIRROR] Fixes a vchat runtime

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/11942
--------------------
Sometimes the proc above returns something that is not a list.


CHOMPStation2:
Already ok

PR #3286, opened 2021-12-06
[MIRROR] Fixes missing ladder assembly icon

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/11919
--------------------
When did that happen?

[issue link stripped]


CHOMPStation2:
Our file is actually correct this time. I wonder if a previous PR messes with this...
/master/code/modules/multiz/ladder_assembly_vr.dm

PR #3287, opened 2021-12-06
Snowfield Revamp

harry3124:
Gateway map change, from a rectangular, single building in the small 'snowfield' that does not have much story onto it, to a base for the proper exploration with a story.

Oh, however, please do not merge until bottom is merged/finished.
(https://github.com/CHOMPStation2/CHOMPStation2/pull/3203#issue-1064051265)


harry3124:
Github you ffffff**ffffffFFFFFF**

This shouldn't be re-opened after the close because of same title

PR #3289, opened 2021-12-07
Adds more firedoor

harry3124:
Adds more firedoor, specifically on Deck 3, making the station more safer.


Kazzc:
Surprised I missed those, nice catch. merging.

PR #3452, opened 2021-12-10
[MIRROR] EVENTS AND LANDMARKS - OH BOY MORE COMEDY

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/11848
--------------------
The following PR will do a FEW things

* Adds a new Jellyfish Migration Event
* Updated Vermin Infestation to now spawn via landmarks vs. areas and airvents (This is to allow it so there are more spots and unique areas we can drop the critters at)
* Updated Carp Migration & Rogue Drone event numbers, and also added landmarks around
* Updated Drop Pod Event with new locations and updated the flavor text of the event
* Disabled Meteor Strike RNG event as it didn't make much sense with the current set-up


CHOMPStation2:
infestation.dm got totally fucked so I copied and pasted original back over. They changed their system to be a completely random spawn over landmarks without specific areas being chosen.

PR #3491, opened 2021-12-11
Rocket launcher damage increase

harry3124:
Previous : Pewww _*slap*_
Yes, it was 30 damage. Somehow. Explosion included? 40.
Now, have a real damage.


Cyrelius:
This is too much, especially with the armour penetration.

harry3124:
> This is too much, especially with the armour penetration. It would make them a one-shot-kill weapon on basically anyone for how much damage it does to a single location, and those are not fun.

On the balance work at the moment, yep. This is not a final value.

harry3124:
Doesn't requires alteration.

PR #3499, opened 2021-12-20
Haste trait removal

harry3124:
Title says it all, the haste trait removal.

Looking by different perspectives.
Reason why the haste trait should be in there?
 - Gameplay gets easier.
   - Easier to respond on the emergency.
   - Easier to kite the enemies.
   - Easier to avoid dangerous situations.
   - Easier to run around.

Reason why the haste trait shouldn't be in there?
 - Gameplay balance makes it super hard.
   - Either map makers/event managers need to make the strongest mobs to counter the haste trait users; making it terribly difficult to the non-haste trait takers, or be balanced overall; making so easy to break through the situations.
 - Tons of other servers gives a method; such as adrenaline rush implant, rather than being 'real free estate.'
   - Speed is LITERALLY the critical in the ss13 combat and so many combat-focused servers are restricting or harshly punish whoever gets that speed; Example having a higher income upon damage and immediate internal bleeding (CM, perhaps dozens of more) or occasional stun if the implant is being used too long (Paradise) 
   - There were thousand times 'balancing' the speed in combat, and being a hot potato upon balance of the game play. That is solely upon the 'ITEMS' and 'CHEMICALS' that increases the speed even in dozens of servers, while we here have a free speed boost with dozens of negative traits.
  **NO, PLEASE don't speak out "We take so many bad traits in return." There are so many negative traits, but mostly very minor/not affecting the combat (Example of mental issues; such as aggrophobia and so on) and it is so easily be abused, and it IS getting abused.**
   - Same reason above - There are dozens more methods to compensate if one really wants to avoid rather than "god mod" (Example of only from traits being: High Endurance, Traceur, Snow walker, Photosynthesis, etc etc. Meds being Hyperzine, speed-boost module for the hardsuit and so on.)
  

I am well aware this will be like provoking the bee hive, but I shall take it so, because it is ridiculous.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/positive.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/positive.dm
@@ -1,24 +1,6 @@
 /datum/trait/positive
 	category = 1
 
-/datum/trait/positive/speed_fast
-	name = "Haste"
-	desc = "Allows you to move faster on average than baseline."
-	cost = 4 //CHOMPEdit
-	var_changes = list("slowdown" = -0.5)
-	excludes = list(/datum/trait/positive/hardy,/datum/trait/positive/hardy_extreme,/datum/trait/positive/hardy_plus,/datum/trait/positive/speed_fast_minor)
-	
-//ChompEdit
-
-/datum/trait/positive/speed_fast_minor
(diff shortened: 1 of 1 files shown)


DerBlitzkrieg:
No lmao

PR #3512, opened 2021-12-31
Adds Honk-Hound

Verkister:
Absolutely cursed in all the right ways~

Also fixed the regular confetti cannon's flaccid throwing performance while at it.


Skysteak:
You've done God's work.

PR #3546, opened 2022-01-09
[MIRROR] Makes messages for climbing up/down rafters a bit more obvious

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/12003
--------------------
Adds a proper message for both when someone starts and ends climbing, that is visible to both those above and below. A single ambiguous 'you can hear something climbing' when you can actually SEE someone climbing was silly.


Darlantanis:
Already implemented at some point.

PR #3575, opened 2022-01-17
Yet another patch for borgs.

Cyrelius:
Adds phase rifles to miner borgs. 
Adds slime baton and taser to sci-borg, which was on the non-doggo, but not the doggo.
Adds the ability to grab NIFs to the medical gripper, letting medical borgs handle nif installs.


Darlantanis:
phase.dm needs //CHOMPedit comments

Cyrelius:
Should be good now.

PR #3617, opened 2022-02-06
[MIRROR] Item bank!

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/12155
--------------------
It always sucks when you get an item you know you would love to use in a specific scenario, but, for one reason or another, you know that's probably not happening this shift. 

It would be cool if you could save that thing you found until you know it's time to use it!

So I made an item bank.
This stores and saves a single item based on your ckey!

This is intended to allow people to save an item they know they may want later but have right now. It is not intended to allow one to keep an item forever, or to store multiple items. Ultimately I think the rarity of certain items is a big part of what makes them special!

But, if you find it and can't use it right now, it'd be cool if it was there for you when you can use it. 

Restrictions:

-Any item that has been retrieved from storage can not be stored again! (To prevent people from keeping an item forever.)

-One can only store any item one time per shift! (To prevent abusing possible contents of items.)

-Certain items cannot be stored! (See: ID cards, pdas, communicators, radios (and their keys), holders (micros, pets, etc), most storage items, unique items, certain weapons, etc)

This system stores the item type, so any modifications to the item will not be saved, it will be as if the item is new (because it is).

This means that it will be possible to duplicate the contents of things that spawn with contents. I have thought some about what things this might be a concern for, but there will invariably be things that remain that this will be possible for. If it becomes a problem, it is always possible to apply further restrictions. (this is why most storage items are disallowed. Like, any medkit or gun case or things like that, would be able to duplicate their contents if they were able to be stored, and I feel like this is a better solution than making it so you cannot store something and retrieve something in the same shift.)


Darlantanis:
This can't go wrong.

PR #3632, opened 2022-02-07
Makes toilets slightly more fun

Verkister:
Instead of instantly just zapping the contents away, they now swirl the stuff around before making it vanish one item at a time, keeping the things contained for a 20 sec cooldown time before it gets zapped to the exit landmark, this serving as both, spam protection and a chance for the possible player victim to watch the action finish before they're bluespaced away.

Not even gonna bother marking as chompedit for now cause it was already all my code and I doubt anyone else up there's gonna be touching the file and might just end up porting it there anyway if I somehow get a charitable urge to provide them with non-critical funfeatures again :v

### DIFF
--- a/code/game/objects/structures/watercloset_vr.dm
+++ b/code/game/objects/structures/watercloset_vr.dm
@@ -4,6 +4,8 @@
 /obj/structure/toilet
 	var/teleplumbed = FALSE
 	var/exit_landmark
+	var/refill_cooldown = 200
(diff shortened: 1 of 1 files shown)


Verkister:
hngh goddamn poi mercs runtiming my unit tests smh

Verkister:
Closing this one in favor of a bigger PR including a map tweak.

PR #3633, opened 2022-02-07
Makes toilets slightly more fun

Verkister:
Instead of instantly just zapping the contents away, they now swirl the stuff around before making it vanish one item at a time, keeping the things contained for a 20 sec cooldown time before it gets zapped to the exit landmark, this serving as both, spam protection and a chance for the possible player victim to watch the action finish before they're bluespaced away.

Also makes the exit landmark use a new contained item retrieval system instead of just making a messy lagpile in the open. The machine also processes down certain things, these icluding foods, organs, and vore eggs, making the thing not be useful as a prank food vendor for your dear enemies, and popping open vore eggs so that nobody needs to end up blindly guessing their mystery contents from the retrieval menu. Also the machine has a funny little feature to get rid of the unwanted contents as well. ~~(SPOILERS: it's a "Muffin Monster" mode with a twist. The monster is an actual voremob contained within the machine, and if active, the contents will end up in its gut. Deactivating will release the contents back into the machine. Sounds like some sorta Bethesda level brainfart, right? But I did it on purpose for another even funnier feature: Emagging the machine while the mode is active will release the monster as a hostile mob lmao)~~

Not even gonna bother marking as chompedit for now cause it was already all my code and I doubt anyone else up there's gonna be touching the file and might just end up porting it there anyway if I somehow get a charitable urge to provide them with non-critical funfeatures again :v


Razgriz1032:
Muffin monster... This name. It scares me.

PR #3641, opened 2022-02-08
Update to Sheet Materials

Holdelta:
While this doesn't revert all the changes upstream to the mining materials, it does replace most of them with sprites from other servers, which I feel are more appropriate here.

image

Also, phoron is now a forbidden ice cube.
image

If some of these are unwarranted, I can revert them to what they originally were, but I feel like the slight modern touch to some of these materials do look nice.


Razgriz1032:
These look delicious

PR #3651, opened 2022-02-10
Couple more flush tweaks.

Verkister:
Fixes an oversight of picking up or moving contents off the tile mid-animation not saving them from getting flushed.
Also makes it possible to flush small (unbuckled and knocked down) mobs up to 75% size :v


Razgriz1032:
`Also makes it possible to flush small (unbuckled and knocked down) mobs up to 75% size :v`
IT BEGINS

PR #3660, opened 2022-02-13
Removes Command Secretary from whitelisted jobs

BlackMajor:
It's just assistant with bridge access, made for ~~being ploughed~~ intermingling with the command staff.

This doesn't need to be whitelisted.


Rykka-Stormheart:
Thank you! I'd been asking for this for a while.

PR #3665, opened 2022-02-13
Removes Baymiss

BlackMajor:
Does exactly what it says in the title.
I have not adjusted any damage values. So yes, buckshot shotguns are the most damaging weapon class because of this.
Damage balance can come in later. Removing misschance is a big change to ranged combat and I'd rather not run guesses about what needs changing without useable data first.

Removes the layers upon layers of RnG fiasco for making bullets sometimes miss.
Seriously it was like four layers deep what the hell, there was even a bug in there about it, evasion as a stat didn't work either.

But hey that's fixed now.
By default, all projectiles / thrown items (crossbows) have a 100% hit chance.

I've also removed the obnoxious screenshake that occurs when getting hit by a bullet. Yeah, sure, you'd reasonably jump when being shot, but showering your screen with earthquakes is just destroying your agency.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -404,11 +404,15 @@ emp_act
 			zone = ran_zone(BP_TORSO,75)	//Hits a random part of the body, geared towards the chest
 
 		//check if we hit
+		/*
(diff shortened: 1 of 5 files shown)


CommodoreYouth:
More like BasedMajor

PR #3679, opened 2022-02-14
Fixes communicator watches breaking gun reloading

BlackMajor:
Webedit PR moment
Yeah so anything in the "gloves" slot gets checked for bullet forensics.
Communicator watches aren't a /gloves subtype so they're not inheriting the correct vars.
Lazy fix, so I added one.
Complicated fix would be changing the watch into a gloves subtype.


Razgriz1032:
WHAT THE FUCK. WHY IS A VAR REQUIRED TO SHOOT A GUN. WHO THE FUCK. WHY THE FUCK. HOW THE FUCK.

PR #3712, opened 2022-02-19
Brings back pref ignoring micro stomps

BlackMajor:
Except I did it smartly.
See, stomping on someone was a mechanic to balance out people being two pixels tall and impossible to click in melee combat.

But then some genius made it a toggleable pref, so a cigarette with a fireaxe is the most dangerous thing to any sec team.

I'm bringing back no pref micro stomping in a smart way.

If you have step mechanics turned on, all the fetishistic messages about foot stuff, grabs and taking damage are included.

If you have step mechanics off, and are 100%+ scale, you cannot be stomp stunned. Macros can still walk over you and you can still walk under them, tho.

If you have step mechanics off, and are below 100% scale, anyone that's 50% or more scale above you can briefly stun you with a disarm/grab/harm step with a simple message of being knocked down. Nothing more, no foot stuff, no damage. Just a stun.

Hopefully, this should make micros non-threatening in melee combat, as intended, while also not being aggravating to people with step mechanics turned off.

### DIFF
--- a/code/modules/mob/living/living_movement.dm
+++ b/code/modules/mob/living/living_movement.dm
@@ -112,14 +112,14 @@ default behaviour is:
 
 			//VOREstation Edit - End
 			forceMove(tmob.loc)
-			//VOREstation Edit - Begin
+			//CHOMPSTATION Edit - Making macro/micro step mechanics mandatory again for balance, but removing the fetish aspects if pref denied.
+			//There's nothing fetishistic about politely stepping past someone.
 			// In case of micros, we don't swap positions; instead occupying the same square!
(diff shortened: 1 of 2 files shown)


Darlantanis:
Cutoff should be 50% imo, seems unfair to punish anybody playing playing an easy-to-see scale when this is only a problem for actual micro sizes.

We may need to add a "using this as a combat weapon against other players is prohibited" warning to the sizegun if we add more micro-punishing mechanics. Otherwise this will encourage a "keep resizing toggled off" meta which I would rather not see in a vore server.

PR #3733, opened 2022-02-20
Rocket Launcher Damage Increase

harry3124:
Damage has been severely low on the previous value.
Splash damage/radius has been decreased to compensate.

### DIFF
--- a/code/modules/projectiles/guns/launcher/rocket.dm
+++ b/code/modules/projectiles/guns/launcher/rocket.dm
@@ -38,4 +38,6 @@
 		rockets -= I
 		return new I.projectile_type(src)
 	return null
+
+/obj/item/weapon/gun/launcher/rocket/handle_post_fire(mob/user, atom/target)
 	..()
--- a/code/modules/projectiles/guns/launcher/rocket.dm
+++ b/code/modules/projectiles/guns/launcher/rocket.dm
@@ -38,8 +38,4 @@
 		rockets -= I
 		return new I.projectile_type(src)
 	return null
-
-/obj/item/weapon/gun/launcher/rocket/handle_post_fire(mob/user, atom/target)
-	message_admins("[key_name_admin(user)] fired a rocket from a rocket launcher ([src.name]) at [target].")
-	log_game("[key_name_admin(user)] used a rocket launcher ([src.name]) at [target].")
 	..()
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/guns/launcher/rocket.dm
```diff
@@ -38,8 +38,4 @@
 		rockets -= I
 		return new I.projectile_type(src)
 	return null
-
-/obj/item/weapon/gun/launcher/rocket/handle_post_fire(mob/user, atom/target)
```
Rykka-Stormheart, line 42: This should be left in (not the admin log/message, just the proc) - launchers may need to do a callback post-fire, but if you're set on removing it, remove the ..() below.

Rykka-Stormheart (changes requested):
The proc should be left in, admin log + messages can safely be removed, but if you're removing the entire proc, remove the ..() too

Cyrelius:
The armour penetration might be a little much, but this looks very good overall.  My worry is one-shotting people, it's not fun when you get oneshot-killed.

Razgriz1032:
To do that much armor pen and have a 4 tile explosion radius would be too much.

I'd honestly make multiple rocket types, HE with no armor pen and the 4 tile radius and lower damage, and HEAT with large pen and small radius and high damage.

Rykka-Stormheart:
> To do that much armor pen and have a 4 tile explosion radius would be too much.
> 
> I'd honestly make multiple rocket types, HE with no armor pen and the 4 tile radius and lower damage, and HEAT with large pen and small radius and high damage.

This would be awesome, choose your ammo based on situation, rather than one-size-beats-all.

Razgriz1032:
Ahh, missing comments. See this for comments guidelines. 
https://github.com/CHOMPStation2/CHOMPStation2/blob/3fe84f61e8fff3cc962c803207b2f0acf80d3576/.github/CONTRIBUTING.md

==============
When changing any code in any stock Polaris .DM file, you must mark your changes:
For single-line changes: //CHOMPEdit - "Explanation" (Edit can also be Add for new lines or Removal if you are commenting the line out)
For multi-line additions: //CHOMPEdit - "Explanation" and then at the bottom of your changes, //CHOMPEdit End
For multi-line removals: Use a block comment (/* xxx */) to comment out the existing code block (do not modify whitespace more than necessary) and at the start, it should contain /* CHOMP Removal - "Reason"
==============

PR #3856, opened 2022-03-18
Reorganize map defines to match in-game z-levels

Razgriz1032:
Because not having them match in-game z levels causes issues


Rykka-Stormheart (commented):
Casino will need the datums commented out at lines 246 -> 306

PR #3864, opened 2022-03-19
Add northern star mines

Razgriz1032:
Attempting to bring back the old Northern Star mine as a visitable area in space.
Slides the Aerostat off of kara.dm.

kara.dm holds the overmap object
northern_star.dm holds the datums and the areas and the landmarks.

Apparently I still need to solve for active edges

The mines have no cave gen or ore gen. I would like to get cave gen to work in the future. Not sure how to do that on a late load map tho. But I don't think we'll include ore gen for now for lore reasons.


Razgriz1032:
Active edges solved

PR #3907, opened 2022-03-22
[WIP Don't Merge] POI Sanity Project

EveryStepYouTake:
Point of interest submaps are a bit of a mess. Sometimes there are multiple versions of the same submap (i.e. house1, house1_airless, house1_airless_vr, house1_vr) when only house1 is used and the others have absolutely zero application and might as well be deleted. Sometimes the versions used have outdated item paths (common for ammunition) which means they are spawning with missing items and not giving the intended experience. Sometimes they are named badly - 'old house' is more of an old library, 'smol3' is a merchant stand, and quite dangerously 'shantak den' is actually a dragon den now!

This project is just here to fix those problems and also organises file paths and names a bit better.  Unused unique submaps are categorised, so far the only one I have reenabled is 'ruptured canister' for plains as it seems fairly mundane and I think it wasn't intentionally removed. POI areas will also use their zlevel name now i.e. 'POI _Plains_ - Boathouse' which makes them a bit easier to find in the teleport menu.

This is a bit boring and very tedious but will make any submap work in the future considerably less perplexing.

- [x] Plains Submaps Done
- [ ] Mountain (Mine) Submaps Done
- [ ] Wilderness Submaps Done


Kazzc:
I love what you are doing here and this is pretty good content, I been planning on organizing and making more POI's so this already tacks something off my check-list. But!. . .
We are a downstream of Virgo so a lot of this "POI - Duplicates" are from the downstream. Deleting it's not really a good option as it can cause issues later down the road, so instead I opted in to making "CH" or so chompstation variants that work on our code.

PR #3910, opened 2022-03-23
Organize POIs

Razgriz1032:
Now with 100% less drama


Razgriz1032:
Map Diff Renders not working because too many maps are being modified, lol

Good news is that I can ignore that as I activated the map test flag and it should be coming back good now

PR #3917, opened 2022-03-25
[MIRROR] Auto-Resleeving

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/12529
--------------------
Adds auto-resleeving!

Under normal conditions, this allows one to resleeve their character after they have died in the round. 

The auto-resleever requires a bit of time after death to be used:

-Characters who died due to vore reasons must wait 15 minutes

-Characters who died for other reasons must wait 30 minutes

The auto-resleever only requires power, there is no biomass requirement.

The auto-resleever also works for species that can ordinarily not be resleeved, since the potential for abuse there is minimal.

To use the auto-resleever, one only needs to click it as a ghost. If it is ready to revive you, it will pop up a confirmation dialogue.

Additionally, the automatic transcore notification, and the 'Cancel Transcore Notification' verb has been removed. Ghosts have a 'Notify Transcore' verb, which can be used around 5 minutes after death if they wish to notify medical.

This should eliminate false positives for medical, because no one likes resleeving people who don't need it.

This does not otherwise alter the original resleeving process or machines.

The auto-resleever for the tether is on Z1, just south of the psych office. It also features a return chute for items right next to the door. (I wanted to put it in the resleeving room's cleaning room, but it's really far into medical)

The auto-resleever for the stellar delight is in the resleeving's cleaning room.

I also fixed the access on the tourbus's APC while I was there.
fixes https://github.com/VOREStation/VOREStation/issues/12525

image


Razgriz1032:
Give it a week before we start working on this because of the shitstorm of fixes rolling in from upstream

Rykka-Stormheart:
Active Edge unit test failed, everything else passed. 7o

PR #3964, opened 2022-03-30
Unfuck sawed off shotgun sprites because we have better sprites

Razgriz1032:
Virgo hardcoded sprite recolors.


SpadesNeil:
You broke sawnoff shotguns with this.
https://github.com/CHOMPStation2/CHOMPStation2/issues/6143

PR #3967, opened 2022-03-31
Fixes things not passing unit tests

Verkister:
Fixes the bridge plank turfs starting with wrong air temperature.
Also fixes the sky island open space having wrong air temperature.
Also fixes sheep vore init runtime.
Also fixes air temperature on some abandoned mine poi turfs.
And some borg lab poi turfs.
Also earlyports a fix for viscerator runtime.


Verkister:
Like a kidney stone :dab:

PR #3971, opened 2022-04-01
Alters the server tick_lag from 1ds to 0.5ds (10tick to 20tick) 

Raeschen:
To allow more granular movement control

Standard tick rate is 10, or one tick per 100ms. This allows a movement speed granularity MAXIMUM of 0.1s delay per step.

this is an OPTIONAL resolution to the haste/trait tweaks, because you cannot alter your slowdown currently at a resolution below 1/10th of a second.

This change will increase the server's tickrate to 20, to allow a doubling of the minimum resolution from 0.1 to 0.05. that -0.5 --> -0.45 change should be possible.

This is a fundamental change in how the server works, which COULD increase server load and cause other unknown problems.

Tested:
Mob AI, seems fine.
Atmos, seems fine.

MC Functions appear to be on a timebased tick, so tick rate changes "shouldn't" alter the MC

***TESTMERGE ONLY*** to check for unexpected consequences.

Alternatively, you can live with a maximum movement speed delay adjustment of 0.1s / 100ms steps instead.

### DIFF
--- a/code/modules/mob/mob_movement.dm
+++ b/code/modules/mob/mob_movement.dm
@@ -288,7 +288,7 @@
 					direct = turn(direct, pick(90, -90))
 					n = get_step(my_mob, direct)
 
-	total_delay = DS2NEARESTTICK(total_delay) //Rounded to the next tick in equivalent ds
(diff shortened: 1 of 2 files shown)


Raeschen:
Ticklag is a config value, my bad

PR #3972, opened 2022-04-01
Fixes config datum ignoring the TICKLAG config setting. 

Raeschen:
Polaris moment. This will be updated upstream too. 

Removed the debug verb because it was only good for testing.


Raeschen:
Confused on the use of world.ticklag vs world.fps, closing this for now

PR #3996, opened 2022-04-03
Remove hand tele 5% chance to space

Raeschen:
Hand tele is supposed to be a reasonably powerful sensitive item for heads.

Removes the 5% chance to just space you


Razgriz1032:
No. This doesn't make sense to make it perfect because it is a hand teleporter, not like a full teleportation station or a big quantum teleporter. Hand tele is also a special contraband item. The 5% makes sense as a risk factor.

Raeschen:
> No. This doesn't make sense to make it perfect because it is a hand teleporter, not like a full teleportation station or a big quantum teleporter. Hand tele is also a special contraband item. The 5% makes sense as a risk factor.

changed it to instead do the same action as the "None (Dangerous)" option instead of teleporting you to a random pos on the space Z layer.

None (dangerous) places you in a random space 10 tiles from origin. The '10 tile' can be increased for more risk if you'd like, or i can yet make it even more riskier.

Darlantanis:
Not sure why the hand tele even has a fail chance, it's an annoyance feature not present in other servers. At the very least it shouldn't proc when the destination is a vore belly.

Raeschen:
> 

That's... why i wanted to remove it but Raz said no!

PR #4041, opened 2022-04-09
[MIRROR] Map tweaks!

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/12681
--------------------
Purge CO2 and edit mime door access


Razgriz1032:
Map tweaks that don't actually affect us, but maybe this code will come in handy (or break things. We'll see)
image

PR #4047, opened 2022-04-10
No recoil for toy guns

Razgriz1032:
buh


BlackMajor:
you missed a couple of "recoil = null" on some guns

BlackMajor:
>:(

PR #4073, opened 2022-04-15
Fix slimes being unable to vent crawl

Razgriz1032:
Slimes got this new thing to do with lighting in their contents

Untested but I hope this works.


Razgriz1032:
It's this fucker
https://github.com/CHOMPStation2/CHOMPStation2/blob/b008ca0762ab774b6d0ad509620569e59bd6cf64/code/modules/mob/mob_planes_vr.dm#L13

Raeschen:
nudging this. Close if not relevant, or fix

PR #4074, opened 2022-04-16
attempting to re-ignite my desire to code by adding a funny poster

KeRSedChaplain:
Adds a meme of the funny clockwork cult deity as a poster decal. If anyone has a better description for it by all means tell me.


Darlantanis:
This will need to be added to a contraband_ch.dmi file or sent upstream. Any .dmi without a _ch.dmi at the end must match upstream, this type of file does not allow mismatching. 

If you move the poster to a _ch.dmi file, the actual code changes will need //CHOMPedit comments to distinguish it from upstream code.

Razgriz1032:
You have a missing pixel in the top left too in one of the frame animations. Also... animated poster? Really?

PR #4123, opened 2022-04-20
Protean rework + Modular folder and some (WIP)

BlackMajor:
Posting this PR early while I work on it so any curious onlookers can give comment and feedback if I'm doing something real stupid.

I may have also gone a bit hogwild and started porting obj traits from tg out of frustration. I'm electing to leave it in but unused to remind me to do that one day.

The following is what I plan to do, and my reasonings.
Proteans are broke af, hot damn. The fetish race is both less effective than normal synths, and less fun to play because of all the jank and crippling design. All this sacrifice so you can.. uh.. goo rp?
Not worth.
A forewarning, that I'm probably going to end up making them "better" than regular synths, minus of course the fact they can't use positive traits. But hey, whitelist race, so. Your discretion, headmins.

First off, the modularity.
Since we're a downstream, it would be wise of us to have a dedicated folder for chomp exclusive content, for the sake of differentiating it from upstream more easily what's ours, and what is theirs.
This folder should only be relevant for solely unique content, and complete file rewrites, like this one.
Most of my future PR's are probably going to end up in these folders, or move parts of our existing code into them.

Now about the Proteans themselves, here's what I'm doing/done/are going to do:
- [x] Prot defenses back to a baseline 1. No burn weak, no brute strong, no shock weak.
- [x] Prot rad immune. Why? Cause I can't be assed fandangling the jank necessary to have them be rad proofed while in some of their special states. (The humanform can get irradiated irrespective of the blob form and has no feedback to the player they're dying) If I were a stronger man, I'd port the entirety of cit's flag system so I can just have the protean blob form be the same human, flagged to look and act like a simple mob.
- [x] Made them pressure and heat immune, for consistency with their blobform also being this way.
- [x] Bump up the protean limb break threshold to 100 damage, instead of 20. Seriously who thought 20 was fair when most mobs do that in one hit. We're trying to do away with having to focus on hyper lethality and OHKO's.
- [x] Probably going to write code to make it so Proteans don't have a crit threshold. Cause you can't blobform while unconscious.. but you'd need to blobform to heal. Wacky.
- [ ] Miiiiight make the steel healing work while not blobformed? Strong balance concern there, might not do it.
- [x] Make blobform not drop all your gear. This was only necessary back when they could vent crawl to prevent them taking stuff with them.
- [x] Make prots unable to blobform while handcuffed, cause that'd be abuseable as shit.

-**Key point here**, make the protean rigsuit a focal point of the species.
- [ ] Ideally, the end goal here is for Proteans to be encouraged to perform symbiosis with another crew member via their rigsuit functionality. (Yes I'm encouraging goosuiting fetish, mechanically)
- [x] Make proteans spawn with their rigsuit already affixed to their back, and be unable to remove it, with a toggle to make its back sprite invisible if you value your fashion.
- [x] Allow the hardsuit transform verb to work both in blobform and human form
- [x] ~~Protean rigsuit starts with a storage module, so you're not losing your backpack by having this.~~ Just made it integrate your backpack instead.
- [ ] Protean rigsuit has an irremovable power cell, that charges off the nutrition of the associated protean.
- [x] Fix the protean rigsuit's bugs and inconsistency. Right now you can't upgrade them because it dels itself.
- [ ] Make a worn prot rig be able to utilise its own equipment while worn. So you can, infact, install a gun into the prot rig, and fire independently of your wearer.
- [ ] On the death of a protean, instead of destroying the body, instead, shunt the body inside of their rigsuit and become unable to reform until repaired. Repair via feeding 5 plasteel to the rigsuit itself.
- [ ] Maaaaaybe make the rigsuit heavier, the more modules it has.. but only if worn by its own protean? That way it's discouraged to wear yourself, but encouraged to have someone else wear you. Maybe not, sounds a bit unfun.
- [ ] Add a rigsuit defib module and map a couple onto the med desk next to the NIF disks. Cause otherwise how's a protean med going to defib someone.

-**Fetish stuff, hee hoo**
- [ ] Give protean blob a verb to latch on to a victim (with timer and warning), forcing the rigsuit onto their back and become irremoveable.
- [ ] Proteans have full access to their hardsuit, and can lock the user out at will.
- [x] A worn protean can click their blobform button to spontaneous vore their wearer, with pref alignment, of course.

- [ ] **The funny**. ~~Low key, I've considered making prots double function as an autolathe. Yep. Probably a module locked behind science, however. Need ammo? Shove this stack of steel into the nearest prot and demand they vomit bullets out.~~
No


LordSaladin:
Lol can we get the ability to shit bricks and use alcohol for energy for no reason other than funny bender references

PR #4129, opened 2022-04-20
adds .45 LC to the autolathes (webedit)

LordSaladin:
adds .45 long colt( in both normal and rubber types and as speedloaders) to the autolathe.
Reason: the colt Single Action Army revolver currently only spawns with 6 rounds meaning it's useless after you fire them this should fix that so anyone can be a space cowboy if they obtain one.


SilencedMP5A5:
shitpost status

LordSaladin:
> shitpost status
silence revolver ocelot /s

PR #4130, opened 2022-04-21
Explorer Carrier Powernet Fix

VictorZisthus:
Fixed a couple of constant power issues on the Southern Cross's Exploration carrier. APCs should start off deactivated with a full charge, and doors should be unbolted to allow entry in said unpowered state. No more power alarms a couple of hours into every shift. Yay!

Crimes McJohnson has left his mark **and** his favorite pair of brass tools. Don't think that his copper stripping days are over though. He will be back.

... Or will he?


VictorZisthus:
Hold up, something went wrong. This should NOT have been that massive of a PR.

VictorZisthus:
I know this is ugly, but check the commits.
image

Darlantanis:
That's why we have map diffs.

PR #4176, opened 2022-04-27
Protean rework part 2 (WIP)

BlackMajor:
Continued from the last PR seen here: https://github.com/CHOMPStation2/CHOMPStation2/pull/4123/

Done so far:
-Fixed a bug with Protean blob hands when rig transforming
-Added species scream sounds
-Add a verb for Proteans to Latch/Unlatch a victim ala forced wearing.
-Add the new death handler for Proteans (dying force RIG transforms them, and they can't come out until fed 5 plasteel). But they'll still be "alive" and able to talk.

TODO:
-Do all the relevant species sprites for protean RIG suits so tesh/vox etc can actually wear them. (Pain)
-Add more Protean blob appearance options, because someone suggested to me I let them look like animals and that sounds fun so I'm going to do that. With associated worn sprites, of course.
-Figure out a way to let them be digested considering the above means they never die in such a way that can be churned in a gut.
-Rewrite the ref full body verb to pull from your current save slot. With checks that species/name are the same.
-Rewrite rig code so Protean blobs can have full control of their suit while worn, as well as control the wearer's movement.
-Create several snowflake Protean rig exclusive modules that can only be utilised by a symbiotic Protean rig. As a means of encouraging performing duality with a host.
-Create a non-exclusive rig defib module, map them into med. Cause right now Proteans can't defib people, lmao.

And if I'm feeling ballsy enough:
-Make a Protean able to speak/emote through its own removed head. So you can do dulluhan RP.
-Figure out why humans don't process radiation/weakened/confuse etc while in nullspace. Cause that's the reason why temporary forms stay perma confuse etc.


LordSaladin:
It's latching time

Cameron653:
-Figure out why humans don't process radiation/weakened/confuse etc while in nullspace. Cause that's the reason why temporary forms stay perma confuse etc.

I believe this is because if you don't have a loc, then life() doesn't go through. 

These two lines in particular: https://github.com/CHOMPStation2/CHOMPStation2/blob/494be6adefe9a364473ce3cedc37b5ccb3dd62dd/code/modules/mob/living/life.dm#L10-L11

I mention this as it's the only place handle_mutations_and_radiation is called (which handles radiation) and handle_random_events (which handles confusion)

If you removed those two lines those two things (and everything else handed in life) would work as normal.

BlackMajor:
> If you removed those two lines, those two things (and everything else handed in life) would work as normal. Would come with it's own set of issues, however.

Not worth.

Cameron653:
Unfortunately so. It'd take a lot of fiddling to get it to properly work. 
Could move things around as well to try and fix the problem, but then you'd risk adding extra things to life() which would more than likely negatively impact performance and cause additional bugs.

So unfortunately it's one of those bugs that will be stuck around.

Razgriz1032:
Conflicts

Razgriz1032:
Conflicts

PR #4186, opened 2022-04-29
Additional gateway mission update

harry3124:
- Map has been extended to North
- Gateway mobs are no more able to be devoured when they're alive (no more vore cheesing)
- Vodka stash has been hidden in somewhere


VictorZisthus:
If you could do me a favor and make a small addition to your PR to save me from having to make one that would be awesome!

at X:113 Y:090 in the snowfield security checkpoint we have this:
image

When it needs to be this:
image

Power won't flow to the checkpoint otherwise!

VictorZisthus:
Cheers~

PR #4203, opened 2022-05-06
Making Gravekeeper Drone Viable

VictorZisthus:
Prior to this patch the poor thing wasn't able to do construction. This should fix that. Expand that gravesite until every simple mob is laid to rest properly!

Added some construction drone and clerical drone equipment. A zippo lighter too. Maximum gravekeeping.


VictorZisthus:
Hmm.

I need to throw one more change in here to meet guideline standards when I get home.

Don't code on no sleep folks.

PR #4228, opened 2022-05-09
Adds import and export style verbs to big dragons

TheGreatKitsune:
Allows keeping a consistent style for these mobs and getting to it quickly. Should be helpful for anyone who plays these things. (Could probably be put upstream as well I suppose?)

Input sanitization on the import is fairly robust, but currently doesn't tell you what you messed up if you did. This might... technically allow transparency, but... if abuse of that becomes a problem, it could probably be fixed. If it can be abused.

### DIFF
--- a/code/modules/mob/living/simple_mob/subtypes/vore/bigdragon.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/vore/bigdragon.dm
@@ -241,6 +241,10 @@ I think I covered everything.
 	verbs |= /mob/living/simple_mob/vore/bigdragon/proc/sprite_toggle
 	verbs |= /mob/living/simple_mob/vore/bigdragon/proc/flame_toggle
 	verbs |= /mob/living/simple_mob/vore/bigdragon/proc/special_toggle
+	//CHOMPEdit start - Add export/import style verbs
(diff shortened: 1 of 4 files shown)


Raeschen:
this should, if possible, be ported in extra files in the modular_chomp folder

Raeschen:
bigdragon_ch is not used. You appear to of left your edits in it after moving your edits to the modular folder.

TheGreatKitsune:
I mean... why do we have that file still then?

TheGreatKitsune:
And also... if it isn't used, then... it should be safe to leave those edits in, especially since it is a _ch file.

Raeschen:
> I mean... why do we have that file still then?

We made it first. Upstream then backported it. So we unchecked our local version and use the upstream version.

> And also... if it isn't used, then... it should be safe to leave those edits in, especially since it is a _ch file.

True, but these things should be kept clean. there's no need for it, so it should be kept out.

TheGreatKitsune:
Fair enough! Changes removed from that file.

Raeschen:
Looks good! If it's locally tested (which it should be) for success cases and any potential abuse cases, seems fine to me.

Raeschen (approved):

PR #4291, opened 2022-05-23
Fix Teleccomms machine browser and log browser.

Drulikar:
Previously, probe/refresh buttons would have no effect on the interfaces because the local list variables for tgui_data were named the same as the machine's. Also fixes usage of temp message and colorizes it.

TGUI rebuild required.

### DIFF
--- a/code/game/machinery/telecomms/logbrowser.dm
+++ b/code/game/machinery/telecomms/logbrowser.dm
@@ -26,13 +26,13 @@
 	data["network"] = network
 	data["temp"] = temp
 
-	var/list/servers = list()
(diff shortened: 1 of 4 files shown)


Darlantanis:
Ideally should be sent upstream given bugfixes, tgui conflicts are a pain to fix when changed upstream (especially if they take a different approach to bugfixing).

Cannot be merged without rebuilding the bundle either; somebody else can PR the bundle fix to your branch if you do not have tgui setup. That said, wait until #4288 is merged then rebase onto master before building the bundle.

Drulikar:
Re-created in a vorestation fork.

PR #4294, opened 2022-05-27
Adds two new tracks to the jukebox

LordSaladin:
from that hit game with funny skeletons and stuff.


Razgriz1032:
This actually won't change anything because the server is currently set to git ignore. I am going to steal the code and add it to the jukebox manually and then reset the gitignore to no longer ignore the jukebox

PR #4296, opened 2022-05-28
Fixes for two communication bugs and feature added for ghosts in belly

TheGreatKitsune:
Bug fixed: Sign language and other non-verbal communication couldn't be seen while either the speaker or viewer were being held.
Bug fixed: Blind people can see messages sent with the 'Me' verb now, as long as they are adjacent or being held. (Much like with subtles.)
Feature added: If you are a ghost inside a belly, and the pred sends a subtle while having ghost privacy on, then the ghost inside the belly should be able to see the message. I was unable to test if this works myself, however.

### DIFF
--- a/code/modules/emotes/emote_mob.dm
+++ b/code/modules/emotes/emote_mob.dm
@@ -204,7 +204,12 @@
 				if(M)
 					if(isobserver(M))
 						message = "<span class='emote'><B>[src]</B> ([ghost_follow_link(src, M)]) [input]</span>"
-					M.show_message(message, m_type)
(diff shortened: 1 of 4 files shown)


TheGreatKitsune:
This is waiting on the ghost subtle feature to be tested, I should say. I wasn't able to do this by myself because the ghost privacy pref either wouldn't be checked due to pred not having a ckey, or my ghost wouldn't be in the belly to see if this actually worked.

TheGreatKitsune:
That final feature is tested and working now, so... this is ready for review.

PR #4310, opened 2022-05-29
Un-nerfs the Prommie blobs

Cyrelius:
List of changes:
They don't drop things except for things in their hands.  (Why was this a thing to begin with?  They can't vent crawl, there's no valid balance reason for this to be a thing that I can think of.)
They actually have hands in their blob form, and do a small amount of damage, exactly like protean blobs.
They heal slightly faster in blob form than otherwise, making it useful to transform into when wounded.
Increases their speed to match protean blobs' own.
Removes the weakened state that you get when you transform into blob form.  (why was this even a thing to begin with?)
Grabs the headsets and ID from the carbon form to use in blob form, and puts them back when done.

Known bugs: 
ID slot and headsets are invisible after shifting back from blob form, but can still be interacted with and used in carbon form, and will still be there if you shift back into blob form.  They're also show up immediately if clicked on in carbon form.  This one's beyond my technical skill to fix.

Unsure if this is a bug or not, but the carbon form takes hits and will gib limbs when the blob form gets hit hard enough.  I'd rather this weren't the case, and that the prommie would reform with all its limbs like proteans do (or perhaps did before?), but I don't know how to fix it.


LordSaladin:
Nice.

PR #4334, opened 2022-06-06
Add preference for variable pitch emotes

TheGreatKitsune:
As much as I appreciate that it is an option, I do think it could use a preference attached to it.


Darlantanis:
Bless

PR #4353, opened 2022-06-13
Girders now render above catwalks.

TheGreatKitsune:
And probably also a few other things like dirt/blood. But girders now render the same way that Grilles do as far as layers/planes are concerned.


VictorZisthus:
Wait hold on, I don't understand.

You're just moving the girders "up" then, right?

Darlantanis:
Missing CHOMPedit comment

TheGreatKitsune:
Ack, comment added. But yes, this effectively does move them 'up', though given they are objects you can't move through anyways...

PR #4394, opened 2022-06-19
Robotics Brute Repair Surgery Fix

VictorZisthus:
Added BASIC FEEDBACK in the form of a damage check and notice message for the surgery step so people stop attacking their patients on the operating table mid-surgery.


VictorZisthus:
Gottim. Finally.

PR #4395, opened 2022-06-20
Gateway snowfield alt version (empty)

harry3124:
Made upon request and suggestion of the peoples on the past.

Completely empty version of the Snowfield base, for anyone who may use it.

All the doors, buttons and lock down blast doors are functional and linked.

Access are already on pre-set by the Snowfield-ID cards.

Do not merge, for someone wants to make a use of the map.
This is a raw version of the map; not intended for live game without adjustment of the others, or within moderation of someone's.
(It is actually fully functional to play on as station crews, however.)


Kazzc:
Good map, could likely make it into a gateway mission easily and easily use it on events in the future. 
Thanks for the content. :o

PR #4411, opened 2022-06-24
Allow carbonkin to see other phased entities while phased

TheGreatKitsune:
Rather then touch the code that controls carbons ability to see invisible things, I decided to do this instead.


Kazzc:
Sure, why not. This seems pretty useful if it doesn't bug anything out. ✔️

PR #4480, opened 2022-07-03
[MIRROR] GitHub Workflows

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/13233
--------------------
~~Adds workflow to test all of our live maps at once.~~
Adds a workflow to test the tgui bundle (if the PR has the TGUI Bundle label).

Another template, having the [MAPPING] prefix and additionally adding the Type: Map label automatically.


Razgriz1032:
Missing TGUI bundle

PR #4493, opened 2022-07-03
[MIRROR] Communicator Reply Messages

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/13260
--------------------
This enables the reply button if the message was sent inside of the communicator UI.
Also moves over communicator UI to TypeScript.

[issue link stripped]


Razgriz1032:
Missing TGUI bundle

PR #4501, opened 2022-07-04
Remirrors 4497

Razgriz1032:
... that was closed here https://github.com/CHOMPStation2/CHOMPStation2/pull/4497

Because fuck you conflicts. I don't give a damn. And why do you tell me I gotta resolve it on the command line. Gimme my damn UI because I am lazy.


Razgriz1032:
Missing two tether files that we don't care about

PR #4510, opened 2022-07-07
Earlyports fix for minimum examine size

TheGreatKitsune:
Earlyports fix made in https://github.com/VOREStation/VOREStation/pull/13292


TheGreatKitsune:
No longer needed with #4521

PR #4511, opened 2022-07-07
Fix stepping on folk laying down.

TheGreatKitsune:
A repeat of a fix I did a while back, but moved to a proper location and updated for the combat step mechanics change.


Razgriz1032:
Moved to a proper location? Where was it moved from? I don't see any deletions

TheGreatKitsune:
It was copied from `code/modules/vore/eating/slipvore_ch.dm`, which got disabled a few months back.

PR #4580, opened 2022-07-20
Say voice sound changes

BlackMajor:
Some stuff our devs preferred.

-Default sound to one of the less jarring goon ones
-One preference toggle for voice sounds
-Me/Subtle use voice sounds instead of the other sounds
-Probably something else I forgot


Darlantanis:
conflicts

PR #4587, opened 2022-07-20
Fixes my jank code involving teshari hat sprites.

tacoguy7765093:
The jank code I had merged here doesn't even work because I forgot to INCLUDE IT in the DME.

This should work.


tacoguy7765093:
WRONG ONE

PR #4591, opened 2022-07-21
Wild POIs (fixed)

FluffMedic:
Removed the line referencing the POI I stopped working on.
Also, sorry if I am not doing this correctly.


FluffMedic:
Okay, sorry, I think I screwed something up there with trying to add onto the old one. 
Sorry for all the trouble.

Kazzc:
Good stuff, I'd say it needs a lot more decor but I'll likely get around updating these maps.

PR #4608, opened 2022-07-25
Nerfing Vore

VictorZisthus:
I've come to nerf vore in the vorecode, and only the headmins and maintainers can stop me!


CMDRYouth:
Finally, Vore is getting nerfed

PR #4625, opened 2022-07-31
Nerf synth-hound pounces

TheGreatKitsune:
Apparently there are two types of pounce tools that dogborgs have. One is balanced, the other... isn't. This swaps all instances of the latter for the former.


User-U-U:
Does this switch them to the one that opens a menu and is impossible to use?

PR #4683, opened 2022-08-12
Fixes Demon Tongues

VictorZisthus:
They now speak daemon as well as cultish hivemind. They could speak cult, so it only makes sense.

I will bonk anyone with a Durasteel bat if I have to explain why Demons should know Daemon.


Kazzc:
Looks about right, surprised I didn't do this when I coded them in. Good stuff!

PR #4722, opened 2022-08-17
 hide_body_part flags for markings, with supplied abnormal bodytype markings.

AnymousUnknow:
Added bacteriophage markings in both full virus and anthropomorphic types, as well as a normal and taur tail.

'hide_body_part' now applies to body markings if set.


Darlantanis:
Any .dmi file edits will need to be sent upstream or implemented in a _ch.dmi file. Due to the way byond handles sprites, .dmi files here which exist on vorestation must match those on vorestation; otherwise any edits upstream will fully overwrite our edits here.

AnymousUnknow:
Aye, will get on it soon as I can

AnymousUnknow:
it is done, i hope

PR #4733, opened 2022-08-19
Desert Planet Tiles

Kazzc:
Little project I am still working on, expect a lot more sprite art but for now.
-Desert tiles
-Sand tiles
-Water tiles
-Gravel tiles
-Mud tiles


Echopsychotic:
Curious, could one of these be allowed to be made via sandstone?

Darlantanis:
You changed the .dme inappropriately again.

Kazzc:
> Curious, could one of these be allowed to be made via sandstone?

Sure? If it can be coded as a cosmetic tile that can be ordered from Cargo or made with sandstone in to sands tiles of sort.
If you can code it I will endorse it.

PR #4759, opened 2022-08-21
Shadekin eject belly contents when dying

TheGreatKitsune:
Prevents people/things inside the bellies from getting qdel'd.


CMDRYouth:
Shadekin can die???

PR #4761, opened 2022-08-22
[MIRROR] Adds Toxins & Cloneloss digestion damage.

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/13546
--------------------
- Adds Toxins as a digestion damage type.
- Adds Clone damage as a digestion damage type.

- Cloneloss gives 2x nutrition than brute/burn since it is exceptionally harder to heal from. (not that it matters)

So now slime characters can actually RP out sucking out your DNA and leaving you a DNAless husk while slamming you with toxins like xenobio slimes can.


Cameron653:
Also, since someone brought this up:
This also makes it so prey don't puke inside of pred's guts anymore. So you can healbelly someone without them being forced to puke every 30 seconds.

PR #4764, opened 2022-08-23
Conveyor belts can be sped up.

tacoguy7765093:
Hey look at this, cutting a wire in the lever makes connected conveyor belts go from a safe 3 mph to one's running speed!

Conveyors here were slow as balls and impractical with how expensive they are. Now you can make them a bit faster.

https://user-images.githubusercontent.com/10555869/186051140-7a71d01b-749b-496a-abdd-dd605250169b.mp4

### DIFF
--- a/code/modules/recycling/conveyor2.dm
+++ b/code/modules/recycling/conveyor2.dm
@@ -319,7 +319,7 @@
 	if(istype(I, /obj/item/weapon/tool/wirecutters))
 		if(panel_open)
 			toggle_speed()
-			to_chat(user, "You cut the safety wire of the conveyor switch.")
+			to_chat(user, "You adjust the speed of the conveyor switch.")
 			return
 //CHOMPedit End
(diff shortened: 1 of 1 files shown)


code/modules/recycling/conveyor2.dm
```diff
@@ -315,6 +315,13 @@
 					conveyors += C
 			return
 
+//CHOMPedit: Conveyor belts can be fast :)
+	if(istype(I, /obj/item/weapon/tool/wirecutters))
+		if(panel_open)
+			toggle_speed()
+			to_chat(user, "You cut the safety wire of the conveyor switch.")
```
Darlantanis, line 322: If I use a wirecutter on a sped up conveyor, does this mean I cut a second safety wire to slow it down again?

tacoguy7765093: fuck

PR #4804, opened 2022-08-29
Fixing Wilderness Atmos

Kazzc:
-southern_cross-10.dmm map size changed from 255x255 to 256x256
-Filled in the missing gap of terrain.


Verkister:
southern_cross-5 and southern_cross-6 are still broken.

PR #4818, opened 2022-09-01
POI - Abandoned Lab

thebestcodertoevercode:
### Adds a new POI in the deep wilderness

 An abandoned lab. 

Abandoned Lab

Featuring just over 16 hivebots, the challenge should be substantial for the reward of a clothing replicator.

In order to get in, a pickaxe nearby is required. The security barriers present are active and hinder movement while also telling a gruesome story about a mercenary experiment gone awry! There's also a weird door in what used to be xenobiology.. Do you dare venture deeper?


Kazzc:
Ey, good stuff. I really like what you got here. :D

thebestcodertoevercode:
Only thing that might be of issue is that the hivebots always duke it out with the mercs placed about inside, and in the xeno room the merc sometimes wins. That hivebot shoots frag grenades and -in my testing-, only about 1/3 times actually hit the target when they hit dead center. 2/3 of the times, the shrapnel shoots outwards of the target, leaving them completely unscathed, and the hivebot usually kills itself with the fragmentation. 
image
 
Other than that, the ensuing fight each time the POI spawns also makes some parts of the POI broken, which I was told is a good thing since it makes "every copy personalized" :p

PR #4834, opened 2022-09-03
Update custom_sprites.txt to actually be updated

Astra-Quark:
When the RUSS sprites were ported over, the text file to actually give RUSS's player permission to use them wasn't. This corrects that.


Darlantanis:
Custom sprites (and items, for that matter) are not permitted here. If you wish to use a custom sprite from upstream, it must be made available to everybody.

PR #4850, opened 2022-09-04
More supply packs and small teppi tweak

UniquaSa:
Add several supply pack that cargo can order, and reinclude the otie supply packs that were removed in 2017 or something.


LordSaladin (approved):
Haha bug

PR #4874, opened 2022-09-08
[TGUI] Newscaster `Create Feed Message` screen FIXES #4833

### LINKED ISSUE #4833: Newscaster Buttons become inaccessible if message body is lengthy.
When accessing a newscater's `+ Create Feed Message` screen, if you fill in a long message for the `Message Body` field the buttons for editing that field along with `Message Title` just disappear off to the right of the screen becoming inaccessible.

*See Below*
image

xnqm8h:
Tested with big ole blocks of text and it doesn't push the buttons off the screen now.


xnqm8h:
image

Darlantanis:
Probably worthwhile sending upstream.

This will need to include an updated tgui.bundle.js. This is the compilation file for tgui and thus necessary for any tgui changes to be reflected in-game.

PR #4878, opened 2022-09-08
POI - Engi Cruiser (Critique and testing are very welcome)

thebestcodertoevercode:
Layout name is from FTL - Faster than Light, a repurposed engi cruiser long abandoned by its former crew, overrun by the mechanical menace. Contains rewards not seen in any other POI's. Challenging and hard, but the loot is worth it.
 
engicruiser


Kazzc:
It's very good, I honestly can't see anything wrong with this one. Impressive stuff, keep at it! <3

thebestcodertoevercode:
Is there a way to prevent the " !! FAILURE !! [MAP: Active edges (all maps)]: Maps contained 6 active edges at round-start." error? As far as I know it shouldn't be related to my POI since the error log mentions turquoise carpet and ignore_mapgen/caves rocks, both of which my POI doesn't have.
It should be safe to merge this, afaik.

Darlantanis:
This is due to POI random gen + errors in pre-existing POI's and thus unrelated to your PR.

Cannot merge currently due to conflicts.

FluffMedic:
Is there a way to make a POI, and PR this without getting conflicts with this PR open?

Example. Leaving empty space in the area code where this POI's code would be?

Thank you for your time. The POI looks cool.

PR #4885, opened 2022-09-09
Defib Buff

VictorZisthus:
Removes the mental interface error on defibs.

Nothing is more annoying for players, medical and patient alike, for an OOC issue to bleed into an in-character one.

A brief bout of afk frustration or loss of internet leading medical to try and revive you using many chems, then giving up an leaving you for dead... A Search and Rescue going out to rescue someone in such a state... Why punish players of medical for doing their job with a no-sale at the absolute last step?

Bonus points, monkey's are now revivable. Yay!

### DIFF
--- a/code/game/objects/items/devices/defib.dm
+++ b/code/game/objects/items/devices/defib.dm
@@ -308,8 +308,8 @@
 		return bad_vital_organ
 
 	//this needs to be last since if any of the 'other conditions are met their messages take precedence
-	if(!H.client && !H.teleop)
-		return "buzzes, \"Resuscitation failed - Mental interface error. Further attempts may be successful.\""
+	//if(!H.client && !H.teleop)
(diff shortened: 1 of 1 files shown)


tacoguy7765093:
I support this, but how will this work with dead players if the ckey is online but their ghost isnt in the body, to they get forced back in or does it still not revive?

Cause I know some people also intentionally do not want to get revived rarely.

Razgriz1032:
Sounds like a good PR! Thanks guys, I hope this will help yall. LMK if it doesn't work as intended but it sounds good for live.

VictorZisthus:
> I support this, but how will this work with dead players if the ckey is online but their ghost isnt in the body, to they get forced back in or does it still not revive?
> 
> Cause I know some people also intentionally do not want to get revived rarely.

The body will be revived if the other checks pass no matter what, but it's still up to the player ghost to re-enter the comatose body manually. If they, for any reason, quit the round or return to menu, the body will remain comatose and an admin would have to get involved to put their mind back.

PR #4886, opened 2022-09-10
[WIP] Updated Promethean Revive

VictorZisthus:
Promethean revival reaction will now leave a solid chunk of phoron in the slime core on failure, allowing the Promethean to revive themselves in the future when/if they reconnect for that session.

WIP TO-DO LIST:
Add verb to allow post-failure phoron crystal check and revival.

Allow examining the core to determine if it's got a crystal in it already, so we don't use up even more phoron. Possible spritework?

Fix bug where gibbed prommie's -genesis networks splatter with default human blood instead of inheriting the prommie's blood.


VictorZisthus:
I need to get back to this one...

PR #4887, opened 2022-09-10
Give medihound back modules that they should have

TheGreatKitsune:
Considering we don't have surgery hound enabled on our codebase, normal medihounds should be capable of doing surgery instead.


CMDRYouth:
Why don't we just enable Surgery hound?

PR #4915, opened 2022-09-11
Allows interaction with mounted circuit assemblies, FIXES #4647

### LINKED ISSUE #4647: Interactions with mounted Custom Circuit Assemblies
#### Brief overview of the feature
Being able to interact with a mounted custom circuit assembly.

#### What you want to happen
When a CCA is wrenched down, clicking on it should open the interaction menu, instead of telling you it's wrenched down and you can't move it. It'd be a lot more useful as stationary devices then. Since it doesn't make sense for it to have buttons but then you can button the buttons.

xnqm8h:


Darlantanis:
Needs CHOMPedit comments if it's not being sent upstream.

PR #4925, opened 2022-09-12
updated voracious lizard sprites

cookiemans:
updated the voracious lizard sprites to make them look less wonky and more like dino

image


Razgriz1032:
Please do not modify base icon files, it is a nightmare to merge in any changes or additions from upstream.

What you should do is instead make a new file in the modular_chomp folder following a similar folder structure to base game folders, and then change the code of the mob to point the icon variable to your newly created file in modular_chomp

PR #4940, opened 2022-09-16
Allow all mobs and punpun to be spontaneously vored

Razgriz1032:
All mobs and punpun can be spontaneously vored. That includes drop, tele, and phase noms.

Remember that vore is an RP tool, not a mechanics tool. If you telenom a boss monster, you will be bwoinked, possibly gibbed (especially because that's funny).


CMDRYouth:
People are going to abuse this in spite of the danger of being bwoinked or gibbed.

PR #4970, opened 2022-09-21
Buffs muffin monster

Verkister:
Oh no.
Also adds sprites for the thing and a framework for simplemobs (via init_vore) or ingame varedits to set custom vorgan entry sounds.


Razgriz1032:
Omfg this PR is hilarious

PR #4986, opened 2022-09-23
[MIRROR] Fixes recursive explosions

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/13772
--------------------
Makes them not bug out and not have infinite loops or make L shaped explosions.

Not enabled in the config, so doesn't technically effect us unless it's enabled down the line for whatever reason. 
(It makes explosions in open spaces larger and explosions in enclosed spaces smaller. Takes cover into account. Etc. More destructive booms but more realistic. Also is more resource intensive but I can't confirm how much.)

Recursive fix


Razgriz1032:
Code status: Lifted

PR #5004, opened 2022-09-23
[MIRROR] [MIRROR] conk rete (ABLE TO BE MERGED) 

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/13467
--------------------
Original PR: https://github.com/PolarisSS13/Polaris/pull/8671
--------------------
Adds concrete as a material, somewhat ported from bay/nebula (at least, in terms of the colour value and the texture overlay for it), obtainable via cargo or the stacks found in engineering, or by crafting it with 2 water, 2 silicate and 2 calcium per brick.
Cynosure and a number of POIs have been edited to have concrete walls, except in some special locations such as the toxins burn chamber and the supermatter core.
There is also a new type of flooring for mapping and POI use.


Razgriz1032:
Virgo just had to include map edits too

Razgriz1032:
I am scared. I will attempt to merge. I think it's just many new and additional maps that we'll probably never use.

PR #5017, opened 2022-09-23
[MIRROR] Some (smaller) .js files moved over to .tsx for interfaces

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/13263
--------------------
These will be broken down into seperate PRs, to at least not make a giant nightmare of conflict resolving for downstreams.

TypeScript does have advantages over regular Javascript (though in it's core it works the same: Javascript is valid inside of TypeScript).

TypeScript does give developers better debugging tools, especially since it can validate the Javascript (variable types and if variables even exist) and tell if there is an error ahead of time (aka: before it's sent to the browser and the interface gets an error).


Razgriz1032:
Missing TGUI bundle

PR #5038, opened 2022-09-24
Adds in a clockwork cult RIG suit

KeRSedChaplain:
As it says on the tin. It's pretty much just a port and alteration of the clockwork cult armor from citadel. 
Obviously, since we lack the mode, this is intended to be seen via adminbus for the moment. 
I suck at balancing things so yell at me if this thing is too overblown.

image
alt naga sprite doesn't show due to a bug regarding the alt naga sprite itself


KeRSedChaplain:
>CI map error
???????

Darlantanis:
Yes by me, though I'd like to give Raz a chance to peek at the armor values first before merging. 

Map checks may fail if something goes wrong with the bot before it actually checks the PR. CI may also fail from currently-existing active edges triggered by POI spawning.

KeRSedChaplain:
I tried to balance the armor values to be around the same "tier" as merc suits since funny eldritch steampunk cultists, but if it needs taken down a bit it's understandable

PR #5050, opened 2022-09-27
Programmable Automated Chemical Synthesizer 

Darlantanis:
Awhile back I thought "chemistry is even more tedious now than when CHOMPstation started," and wondered how I might be able to might be able to address that. After somewhere over a year of "boy did I sure decide to be overambitious," I bring to you my attempt at a fix: the chemical synthesizer. The chemical synthesizer is a user-programmable machine capable of converting CSV formatted strings into byond lists which instruct the machine which chemicals to mix together to make the desired product. Simply: you add your recipes, fill the queue, press play, and come back later for your chems. As the machine is programmable, you may make a recipe for (almost) any mix you can create in a chemical dispenser.

Details:
-The machine starts in Tutorial mode; attempting to add a recipe in this mode will prompt the user to select each chemical from a list followed by the volume of chemical to add. Adding recipes in Production mode instead prompts the user for a CSV string formatted as "Chem1,Vol1,Chem2,Vol2,...". Recipes added in either mode may be exported to the chat window. 
-The ability to store recipes as CSV strings in notepad to reuse across rounds is intentional.
-Catalysts (or anything not from a cartridge) should NOT be included in the recipes. Instead, catalysts are placed in a catalyst bottle. 
-There are no checks on the chemicals you mix or the recipes you enter. It is your job as a chemist to ensure your recipes are correct and no unwanted side reactions occur. Should the machine encounter an error, it will clear its queue and require you to manually purge the reagents. 
-The time for a recipe to complete depends on the total volume of chems to be added (e.g. 120u of dylo takes twice as long as 60u).
-Instruction napkin included. 

Balancing Factors:
-The machine is slower than making chems by hand. At delay_modifier 4 (the "slowness" I consider fair), I am a bit more than twice as fast as the machine after not doing chems for two years or so. Should this be deemed too fast, the delay_modifier can be increased.
-The machine is limited to the same cartridge volumes as chemical dispensers and recharges chems at the same rate. This limits the maximum volume of chems able to be synthesized to the same limit as a chemical dispenser, as the machine is currently map-in only.
-Recipes are capped to 20 and number of steps in a recipe is capped at 16. These may be adjusted in accordance with balancing concerns.

image
image

Sprites courtesy of @Kazzc 
CALLBACK advice courtesy of @BlackMajor 
TGUI troubleshooting courtsey of @xnqm8h 
Sound courtesy of my job as a chemist


Razgriz1032:
Unit test is ok.

The instructions in this PR would also make good on a wiki page

xnqm8h:
> Unit test is ok.
> 
> The instructions in this PR would also make good on a wiki page

I can add it to the chemistry wiki page when it's merged in.

Darlantanis:
I do intend to add a more proper manual after it's gotten enough use to convince me I haven't left super glaring errors requiring major tweaks (assuming I can figure out tgui for that as well).

PR #5051, opened 2022-09-27
Adds fox taur bodies

cookiemans:
Now you can gecker as a biggol fat foxtaur

image


cookiemans:
I have plans to add fat foxo, synth foxo, and fat synth foxo at a later point in time, right now just want normal taur foxxie

PR #5065, opened 2022-10-02
[MIRROR] Sanitization

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/13825
--------------------
- Makes RCON tags less exploitable.
- Sanitizes some more things so you can't put HTML into it.
- Utilizes TGUI better to show the user the max they can post.
- Limits some things like circuits so you can't insert 50000 characters.
- Allows badmins to send HTML in subtle messages but only if they do it properly.


Razgriz1032:
Seems to be a problem with recent POI Atmosphere fix, but I think that is because I merged that before this one, and I made that PR after this one.

PR #5074, opened 2022-10-03
Remove self-insert

Razgriz1032:
We don't do that here, not even with our own staff

### DIFF
--- a/code/modules/admin/verbs/buildmode.dm
+++ b/code/modules/admin/verbs/buildmode.dm
@@ -372,7 +372,7 @@
 				easter_egg.name = "coding teshari plushie"
 				easter_egg.desc = "A small purple teshari with a plush keyboard attached to it. Where did this come from?"
 				easter_egg.color = "#a418c7"
-//CHOMP Remove end */
+CHOMP Remove end */
 
 		if(BUILDMODE_ADVANCED)
 			if(pa.Find("left") && !pa.Find("ctrl"))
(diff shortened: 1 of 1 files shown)


code/modules/admin/verbs/buildmode.dm
```diff
@@ -365,13 +365,14 @@
 					if(NORTHWEST)
 						var/obj/structure/window/reinforced/WIN = new/obj/structure/window/reinforced(get_turf(object))
 						WIN.set_dir(NORTHWEST)
+//CHOMP Remove start self-insert
```
Cameron653, line 368: Replace with /*

code/modules/admin/verbs/buildmode.dm
```diff
@@ -365,13 +365,14 @@
...
 						WIN.set_dir(NORTHWEST)
+//CHOMP Remove start self-insert
 			else if(istype(object,/turf) && pa.Find("ctrl") && pa.Find("alt") && pa.Find("middle"))
 				var/turf/T = object
 				var/obj/item/toy/plushie/teshari/easter_egg = new /obj/item/toy/plushie/teshari(T)
 				easter_egg.name = "coding teshari plushie"
 				easter_egg.desc = "A small purple teshari with a plush keyboard attached to it. Where did this come from?"
 				easter_egg.color = "#a418c7"
-
+//CHOMP Remove end
```
Cameron653, line 375: Replace with */

Razgriz1032: 🧠

PR #5091, opened 2022-10-08
Tiny Incorporeal fix

BlackMajor:
tl;dr Carbon Shadekin's are coded in a way that employ's the use of spawn(), such that if they phase twice between those spawns, the incorporeal_move var gets set twice. Allowing them to have incorporeal_move = 1 while not having the phase shift flag.

is_incorporeal() only checked for the phase shift flag.
My lazy fix is making it check for incorporeal_move too.


Verkister:
This doesn't work. This is on base mob level while the shadekin check is on humanmob level, and already returns TRUE there without ever getting a chance to ..() into this part. Also I think this should be `return ..()` anyway and might be the cause why ghosts can no longer pass through the sif transition zones

PR #5092, opened 2022-10-08
Lets capture crystal occupant see belly texts

Verkister:
Capture crystal occupants can now see the belly desc printed in chat whenever the crystal enters a vorgan.


Verkister:
Linter a fuck

PR #5127, opened 2022-10-10
Adds pAI and simplemob directory visibility

Verkister:
Everything tested working. Maybe this might also remind the deployment system that this file had been updated.


Darlantanis:
Going to hold off for the moment. I'm not seeing deployment succeeding yet despite many merges so maybe the last one never went live.

PR #5128, opened 2022-10-11
Science has a smart sheet storage machine now

tacoguy7765093:
No more crates and piles of metal just laying around all messily! Now you can storage the stuff in a convenient machine!
Note: This is **not** a ore silo,
This is **not** persistent.
This is just a machine for storage and to make things clean.
Tested and it works.

StrongDMM_0edVAkGHG2

QmS4Nnp6Yl


Darlantanis:
RIP plant

PR #5133, opened 2022-10-14
Adds a toggle for vorgan entrance chatlogs

Verkister:
Only in ingame preferences tab as a non-persistent setting for now. Might make a proper pref later if folks want that, but currently it's just a niche thing I wanted to try out for some extra level of mystery with rng autotransfer setups. Basically just makes it so that things entering a vorgan don't print out "thing slides into your vorgan" into the chat every time it happens.


Razgriz1032:
Conflict :(

PR #5159, opened 2022-10-20
Reverts an old YW nerf on teshari

Verkister:
It's time to sit on faces again.

WIP for now til i'm done with scoopy fix


ZVB334:
Please don't merge unless if a workaround fixes the fact this will break teshari scooping

tacoguy7765093:
Wonder if there is a way to keep the scooping but allow the pinning down.

PR #5161, opened 2022-10-22
Adds gradual corpse digestion

Verkister:
Adds a belly setting option to enable gradual corpse digestion. With this feature, a pred can digest their prey for longer even after death instead of instantly deleting the body upon expiring. Prey's body will finish digesting when the dead body has been given another dose of their maxhealth's worth of glorpin (total nutri gain halved due to double round). Also comes with an extra option for humanmobs to digest individual body parts one by one before the grand finish. (Why? Simply because I had a brainfart and a mood to chase it for no reason other than personal flex to see if i can code it so the code simply exists now. Why not? That's up for you to decide lol)

Currently lacking vorepanel buttons though. Haven't got the TGUI baking equipments set up on this compy yet and got maybe offered some assistance with it. (Per-belly settings for "slow_digestion" var (main feature toggle button) and "slow_brutal" var (that humanmob extra option toggle button)


Kazzc:
About time we got this, surprise it took this long for someone to make it. Good stuff, Verky. <3

PR #5186, opened 2022-10-25
Revert "Reverts an old YW nerf on teshari"

TheGreatKitsune:
Reverts CHOMPStation2/CHOMPStation2#5159 due to it being WIP and causing unintended changes at the moment.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/teshari_vr.dm
+++ b/code/modules/mob/living/carbon/human/species/station/teshari_vr.dm
@@ -1,5 +1,5 @@
 /datum/species/teshari
-	mob_size = MOB_MEDIUM
+	mob_size = MOB_SMALL //YW Edit: changed from MOB_MEDIUM to MOB_SMALL
 	spawn_flags = SPECIES_CAN_JOIN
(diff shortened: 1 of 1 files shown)


Darlantanis:
What unintended changes?

tacoguy7765093:
Teshari cant pass plastic flaps no more, teshari can no longer be scooped like a small animal.

Darlantanis:
Plastic flaps should not be automatically passable by crew. This is intentional.
Teshari scooping is old Polaris code which should have never been activated in the first place. Removing this is intentional; if you want it, recode it in-line with vorecode holders and set to respect vore pnael pick-up prefs.

tacoguy7765093:
I'm sure Azul here will find out a way to fix it

TheGreatKitsune:
Huh. I always assumed both of these were intended traits for Teshari to have. (I mean, the teshari den uses a 'blocker' of plastic flaps so that only teshari have easy access to it.)
Plastic flaps are also always passable in a sense, you just have to lay down and drag yourself to the tile to do it normally. (Thus why I saw teshari just being able to run through them given their greater agility as intentional.)

PR #5207, opened 2022-10-30
Makes bluespace toilets react to bluespace content

Verkister:
Since I had ended up fluffing them to work with Bluespace Magic:tm: I figured they should do what other bs tp stuff does, so now flushing something containing BoH kinda things will scramble the destination coordinates and land it somewhere random instead :v

Just another unimportant boredom brainfart moment tbh


Darlantanis:
When do we hire bluespace plumbers?

PR #5243, opened 2022-11-08
prometheans QoLs and fixes

Seris02:
Adds the ability to change the promethean's bodytype in preferences (but only to the bodytypes you can normally change to)

promethean eyes now actually change when the bodytype is changed so tesh (and others but I'm concerned about tesh) slimes don't have floating eyes or wrong eyes

also, digitigrade because vulp slimes.

Screenshot 2022-11-08 183533

deprecated because I was working on the wrong version from however long ago when I created my repo and positive negative traits were locked to a species, but it works when I tested it on the current code
Screenshot 2022-11-08 183609
Screenshot 2022-11-08 184138
Screenshot 2022-11-08 192849


Seris02:
pr'ing this upstream because I'm a doofus

Seris02:
pr'd upstream

PR #5310, opened 2022-11-23
Send message to pred when prey goes into hard crit

TheGreatKitsune:
Probably not especially accurate, but it should get the job done most of the time.


code/modules/vore/eating/bellymodes_datum_vr.dm
```diff
@@ -57,6 +57,7 @@ GLOBAL_LIST_INIT(digest_modes, list())
  		//CHOMPedit end
 
 	// Deal digestion damage (and feed the pred)
+	var/old_health = L.health;
```
Darlantanis, line 60: missing comment

PR #5311, opened 2022-11-23
Add location of where you are when held by someone

TheGreatKitsune:
A new entry in Status that lets you see how you're being held by someone, in their hands, pockets, on their head, in a bag...


TheGreatKitsune:
For example:
Location

TheGreatKitsune:
(This is also distinct from the admin-viewable `Location:` status entry.)

PR #5314, opened 2022-11-25
New mob variants and cyber horror AI tweaks and sounds

Razgriz1032:
New assets for upcoming Kara project. A few new additions to modular_chomp 
-New gygax variant for Kara with sounds and text
-New drone variant for Kara with sounds and text
-Cyber horror mobs now have sounds
-Regular horror mob variant adds AI behavior and sounds. 
-New weapon sounds under modular_chomp


Razgriz1032:
Should be good now

PR #5317, opened 2022-11-26
Raptor Borgs !

Valekin4:
If I did everything right, this pull request should add raptor borgs !
Current departments are Medical, Security, and Science !
Clown and Cargo and still in the works, and will be added in a future pull request.
borga


Raeschen:
Chomp specific code edits do need to be split or displayed that they are specific to us. See: https://github.com/CHOMPStation2/CHOMPStation2/blob/master/.github/CONTRIBUTING.md

Preferably, if you can, place edits in modular/chomp in new files (or existing files in there)
If not possible, put 
//CHOMPEdit - description
after single line edits, or encapsulate blocks in
//CHOMPEdit BEGIN
//CHOMPEdit END

PR #5330, opened 2022-11-28
[MIRROR] promethean fixes and QoLs

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/14056
--------------------
Adds the ability to change the promethean's bodytype in preferences (but only to the bodytypes you can normally change to)

promethean eyes now actually change when the bodytype is changed so tesh (and others but I'm concerned about tesh) slimes don't have floating eyes or wrong eyes (fixes the same issue for custom species too)

(shhhh, these screenshots are from downstream but I made sure to test it here, just couldn't be bothered making more screenshots)

Screenshot 2022-11-08 183533

below screenshots are technically deprecated because I was working on the wrong version from however long ago when I created my repo and positive negative traits were locked to a species, but it works when I tested it on the current code
just imagine it looks like it should :>
Screenshot 2022-11-08 183609

Screenshot 2022-11-08 192849


Darlantanis:
Currently breaks things upstream

Darlantanis:
#5334 added manually

Darlantanis:
Maybe wait until we're confident it's functional upstream before merging

Darlantanis:
fuck it

PR #5332, opened 2022-11-30
MuchSpiders and Minour Changes

FluffMedic:
Major Changes to
-Manor (Aka spider manor)
--Tldr: Slightly adjusted some room layouts, added more loot, and spiders.

---The main hall is now the center of the nest, giant spiders everywhere. With egg and spiderling creating spiders being rather common.
---Added two secrets rooms, one has a goofy item, pill bottle, and a lore note. The other has some culty stuff, a frontier marksmen rifle, and a Tech Priest robes var edited to have higher resistance to lasers and energy but their low melee and bullet values remain the same.
---There is now gold and black winter coats elsewhere, and in the storage room, more gold, and fancy minerals, to try and give the sense of fancy rich folk home infested with spiders.
---Main/obvious loot: Several Revolvers, two goofy guns, some weird frontier weapons, and a sword.

--The intention with the spiders is to create an overwhelming swarm that needs to be rushed, otherwise the spiderlings or eggs created would just replenish their numbers.

-Also fixed the untextured wood tiles by using the normal wood tiles. 

Then minor adjustments
-Changed Cult Bar windows to Plastiatnium to reduce the times and scenarios where the cultists break out.
-Changed the armory of Doom Peninsula merc base to include armor, to hopefully entice some folks to come to it. 
-Changed the armory of the Ambush/Sniper Base, because one of, if not the only criticism I can recall I got about it was the loot being lacking.

### DIFF
--- a/maps/submaps/surface_submaps/wilderness/AmbushBase.dmm
+++ b/maps/submaps/surface_submaps/wilderness/AmbushBase.dmm
@@ -93,11 +93,26 @@
 /area/submap/DugOutMercs)
 "dI" = (
 /obj/structure/table/rack,
+/obj/item/weapon/gun/energy/vepr,
+/obj/item/weapon/gun/energy/laser,
 /obj/random/energy,
 /obj/random/energy,
-/obj/random/energy,
-/obj/random/energy,
+/obj/item/weapon/cell/device/weapon,
+/obj/item/weapon/cell/device/weapon,
+/obj/item/weapon/cell/device/weapon,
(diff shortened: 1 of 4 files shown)


Cyrelius:
I was advising on these changes, and they do make sense as far as I'm concerned.

Kazzc:
Please provide screenshots if you make any map PR's, thankfully gitcheck worked for once so I can see what you edited. xp
But besides that, I like what you have done here.  Good stuff.

PR #5358, opened 2022-12-08
Let dead folk see/hear stuff like ghosts

TheGreatKitsune:
This could perhaps be behind a preference. This fixes a slight oversight with gradual corpse digestion though, in that once in that state you won't hear messages from the pred without becoming a ghost and leaving the vore belly.


Razgriz1032:
Sure I guess, but hopefully the dead will know not to use metaknowledge, lol

PR #5366, opened 2022-12-09
Quantum Checkpoint Part 2: The Wrath of Medical

SlaysInCode:
- Fixed The Atmos pipes, specifically a missed pipe I didn't realize I missed due to the full tile decal covering the pipes under it visually.

- Fixed the wiring using the correct wire instead of the duplicate that has the right visual but not the correct icon_state.

- Readded public weapons, located in the small room outside Hangar 2 and the checkpoint.

- Replaced the bathroom door with a solid one instead of glass so the Security Officer outside isn't staring you down while you're trying to wash your hands.

- Added 2 new buttons to the Checkpoint Control Console. Door Controls for both the inner and outer door so they can open and close the door from inside and lock it immediately.

- Replaced the single normal wall in the checkpoint with reinforced. OCD havers rejoice.

- Named the doors accordingly so people no longer sit on the bathroom toilet thinking it's the lounge.

### DIFF
--- a/maps/southern_cross/southern_cross-1.dmm
+++ b/maps/southern_cross/southern_cross-1.dmm
@@ -114,8 +114,8 @@
 /turf/simulated/floor/tiled,
 /area/hallway/primary/firstdeck/port)
 "aam" = (
-/obj/machinery/atmospherics/pipe/simple/hidden/scrubbers{
-	dir = 10
+/obj/machinery/atmospherics/unary/vent_scrubber/on{
+	dir = 8
 	},
 /turf/simulated/floor/tiled,
 /area/teleporter/firstdeck)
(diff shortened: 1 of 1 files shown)


SlaysInCode:
FORGOT ONE

-Added a maint room in the maints by the checkpoint

SlaysInCode:
image_2022-12-09_155930640
Screenshot 2022-12-09 155824
Screenshot 2022-12-09 155807
Screenshot 2022-12-09 155747
Screenshot 2022-12-09 155720
Screenshot 2022-12-09 155616
 
image_2022-12-09_160051534

Kazzc:
Most of this was discussed and agreed on DM's, good stuff as always. 👍

PR #5370, opened 2022-12-10
Engineering Vox Rigsuit cost adjustment 

crabcube:
The modules it was anticipated to work with do not work so have been stripped and the cost is now aligned with the other. No need for it to bee that expensive and not have anything in it.


Cyrelius:
You removed the equipment from the /equipped version.  No.

crabcube:
> You removed the equipment from the /equipped version. No.

This requires more work than I can quick fix while at work
Gonna do a WIP build here for the weekend

PR #5373, opened 2022-12-10
Makes audible emotes use voice freq setting

Verkister:
*belch [BASS BOOSTED]


Razgriz1032:
I AM GONNA LAUGH MY ASS OFF IF IT IS BASS BOOSTED

PR #5395, opened 2022-12-13
Add verb to toggle species speech sounds

TheGreatKitsune:
Added to both carbons and simplemobs (Used by Vox and xenomorphs at the moment.)

Also moves the hide nutrition verb over into modular_chomp while I was at it.


Razgriz1032:
Why is hide nutrition getting removed or some such?

TheGreatKitsune:
It is not being removed, it's being moved into the modular_chomp folder instead of being chompedits in virgo files.

PR #5398, opened 2022-12-14
[Security Department] Autolathe

SlaysInCode:
- Adds in a hacked autolathe to the armory so that Sec no longer needs to use their rifle as a baseball bat once they use up their 2-3 magazines.

- Shelf added along with it with 1 stack of steel and 1 stack of glass to start with.

- Moved the table up one tile to give room to walk and moved the holopad into its place.

- Wall cell charges both on the wall opposite the autolathe now.


SlaysInCode:
image

SlaysInCode:
On hold while I fix an error

PR #5404, opened 2022-12-17
Doom Peninsula Overhaul

FluffMedic:
Expanded the merc base upon the peninsula, giving it a purpose of being a research base. Added some more goodies, and more foes. Turrets might be a tad overboard.

Oh and added 8? gyropistol clips, and a cell battery to the manor. Cell battery is in the storage room, and 3 clips with the gyropistol, while 5 are in a closest in another room. Note gyropistol ammo cannot be made. Those 8 clips are all you're getting.

In depthish stuff
-Armory has some armored hardsuits and tactical suits. Pick your poison, Slow down or less defense. Also has some random weapons.
-The new xenoarch area has an auto-cloner but it won't activate until removed from the container.
-The xenoarch vault has a new fun foe to fight, along with a vault full of randomized goodies. I am sure that the big artifact will only cause issues very, very rarely.
-There is a secret room with neat goodies.
-Hopefully folks stop ignoring this POI....actually they probably still will, since the stigma is still there, and many turrets, but hopefully not.

### DIFF
--- a/maps/submaps/surface_submaps/wilderness/DoomP.dmm
+++ b/maps/submaps/surface_submaps/wilderness/DoomP.dmm
@@ -904,13 +904,15 @@
 /obj/item/weapon/archaeological_find,
 /obj/effect/floor_decal/corner/grey/diagonal,
 /obj/item/capture_crystal,
-/obj/item/stack/material/morphium{
-	amount = "5"
-	},
 /obj/item/weapon/circuitboard/mecha/imperion/main,
 /obj/item/weapon/circuitboard/mecha/imperion/targeting,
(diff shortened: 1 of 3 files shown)


FluffMedic:
Well then. That's going to be problematic. Time to have fun trying to figure out if I can fix it.

FluffMedic:
Hopefully that fixes thing.
Did add some further lighting.
And a beacon at Xenomoprhs because there is a transmitter there. (And I wasn't sure if I could do the commit with the merc thing alone)

Razgriz1032:
Very cool, thank you

PR #5405, opened 2022-12-17
sect drone l o a f + sect drone icon improvements (in my opinion)

Seris02:
I felt like the sect drone needed an update to its sprites (because I decided to make a sect drone character).

Also, haven't put this upstream because they don't have taur bellies and their loafing code is different and I'd have to seperate this into two different pull requests, when it can easily be done (basically) fully modularly.

So:
Sect drone taur tail back sprite now actually shows the back of the drone, instead of a 3/4 perspective of the back sprite.
Sect drone taur tail side sprites now correctly show the belly in front of the legs behind them, even if the original sect drone mob sprite doesn't.
Sect drone taur tail has loaf sprites now.
Sect drone taur tail has belly sprites now. (but no struggle animation because I'm not that good of a spriter)
Sect drone bodytype marking (head) is more accurate to the mob sprite, and has borders on it now.
Sect drone eyes marking is more accurate to the mob sprite now.

Taur bellies can now have a loaf variant, of which the sect drone taur tail now makes use of.

See below for example.

droonnnnnnesannotated


Darlantanis:
icons/mob/vore/Taur_Bellies.dmi may be overwritten in the future if upstream modifies it, github does not support conflict resolution for .dmi files. safest way to avoid this is using a _ch or modular chomp file

PR #5409, opened 2022-12-17
Makes Glucose a craftable reagent and uncomments VOREstation removal from sweatmaxes.

tacoguy7765093:
**EDIT: We made a nerfed version of glucose crafted via sacid+milk, a lactose based sugar called Galactose which has an OD and is much less effective.**

A med station can make chemicals that literally unhurt your bones but cant make sweet IV fluid, so now you will be able to make an injectable version of those nutriment+iron pills.

Also uncomments a 2 year old removal from the sweatmax machines for glucose, lunchboxes already have it!

**Why? Is it op?**

- Liquid rations can be already eaten through a hard suit restoring blood
- It is basically the same thing as shoving a 60u pill of iron, protein and nutriment in you but now IV/injectable
- You get shot more often here meaning blood is punched out more often out of you
- This might make hardsuit medical treatment a bit powerful
- ~~Science RIG suits have a glucose slot for refill but no way to make it grrrr~~ Oops snowflake med wont work on it. oh well
Brought this up with Nadyr and others regarding balance, if its too OP could make a lesser version if requested.

https://user-images.githubusercontent.com/10555869/208333734-e2b13af0-2b52-4aed-8803-1957c29786fd.mp4


tacoguy7765093:
Oh yeah, chem recipe is 1 part Salt, 1 part Water, and 1 part Sugar making 3 parts Glucose.

Cyrelius:
I think we may need to nerf its blood replenishment a fair bit, as it is very powerful.

tacoguy7765093:
Did testing, yeah its pretty powerful, 0.5% a tick is pretty op, might make ghetto ass mid version, maybe sneak in OD to it

tacoguy7765093:
From request of Cyrus, I made a different version of glucose with 1/6th the nutritional value, has an OD to prevent someone from just pumping a shit ton of sugar in them and never bleeding out, and a different recipe being sulfuric acid + milk.

Normal glucose is untouched and left as limited medipens.

PR #5517, opened 2023-01-07
Makes round vore stats little more honest

Verkister:
Kind of silly to report 30 individual people getting eaten if a 3 person crew simply consisted of two preds playing belly tennis with a single preything :v


Seris02:
BELLY TENNIS

PR #5519, opened 2023-01-08
You can eat through synth facemask

tacoguy7765093:
Finally. Protogen tesh can commit normal consumption.


Darlantanis:
Missing comments

PR #5554, opened 2023-01-17
Removes all pre-mapped dirt

tacoguy7765093:
No more magic oil stains appearing after cleaning the station aggressively

This only affects trash spawners and random filth spawners. I intentionally left some items on the ground and the 'strange goo' that hints towards the abandoned genetics lab.


Verkister:
For the next round janitors this is kinda like removing the ore gen from miners imo.

tacoguy7765093:
My goal was to have the persistence persistent, I was told maint as a area code already generates trash but that may not be true 
Either way I can't be assed to resolve conflicts, good enough for low pop server come to think of it

PR #5555, opened 2023-01-18
Add a mode to muffle items.

lbnesquik:
# About this pull request
This PR adds a 'muffling' mode for items, for bellies. 
In clearer terms, putting on or removing items, opening bags, no longer makes any noise. 

I did not extend this to every noise making options, as this would require a freaking ton of chances, and I'm too lazy for that. But that should still be a nice improvement.

# Why is it good for the game
It allows you to make your belly appear more muffled than before, while allowing prey people to do things inside of it. Like eating food for their NIF, or silently pulling out their tablet computer.


Darlantanis:
items.dm requires //CHOMPedit comments, otherwise everything looks good

PR #5556, opened 2023-01-18
Shuttle Zooming

FluffMedic:
Swapping around the 3 shuttle POIs a bit.

Made the alien shuttle more difficult, and rearranged the loot a bit. It's still probably overpowered as all hell.
image

Made the blackshuttledown a bit more packed with loots and threat.
image

And made the blueshuttledown not a downgrade of blackshuttle down and hopefully worth the 50 points.
image


Razgriz1032:
Is it bad that I look at the last one and think "horny land"

FluffMedic:
Not at all considering the server.
Personally don't have any desire in them, but the server loves them.

PR #5565, opened 2023-01-19
more digitigrade legs and better handling of said legs' icons

Seris02:
so- the code in modular_chomp/.../organ_icon.dm hasn't changed from the original for... a long time now, I just condensed it down to be more readable, and less effort to edit and work with in general
can undo this and rework the code in a different way if wanted

also
- lets proteans have digitigrade legs
- lets prometheans have digitigrade legs (why could they not)
- allows for the use of digitigrade prosthetics (right now, literally only in DSI stuff (except DSI teshari), because that's the exact same as the normal icons, and I can't be stuffed spriting actual prosthetic digitigrade stuff)
- lets proteans use the DSI - Teshari prosthetics

### DIFF
--- a/code/modules/organs/organ_external.dm
+++ b/code/modules/organs/organ_external.dm
@@ -81,6 +81,7 @@
 
 	// HUD element variable, see organ_icon.dm get_damage_hud_image()
 	var/image/hud_damage_image
+	var/prosthetic_digi = FALSE //CHOMPStation edit - when it's prosthetic, can it be a digitigrade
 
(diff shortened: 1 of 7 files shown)


Raeschen:
The edits on the upstream files could easily be made into modular_chomp directory edits instead. putting "//ChompEDIT" tags is okay but putting it in the modular folder is preferable

Seris02:
will do in a bit, working on another pr to make proteans a small bit better

PR #5567, opened 2023-01-19
makes proteans store their bodytype like custom species/prometheans, also fixing their eyes in the process

Seris02:
title

it also means they can pick their bodytype in the preferences menu like the custom species and prometheans
actually planned on doing this with that PR, but it was done upstream and so didn't have the modular protean files

unfortunately, I'm not willing to go through and code up a proc to verify all their prosthetics to make sure they fit the bodytype / change prosthetics to match it, but it didn't do that before this PR either, so...


Seris02:
closed in favour of PR #5589

PR #5573, opened 2023-01-20
Makes slimes in xenobio split to 5.

Cyrelius:
This should make it faster, while also making it more dangerous.


FluffMedic:
Sounds fun.

PR #5574, opened 2023-01-20
Making clothing protect against water like it should.

Cyrelius:
Especially for those species that are vulnerable to it (mostly Prometheans).
Also adds water_protection to a few test variants of the cohesion suits, may fold that in later, this should be test merged for now rather than fully merged.


Cyrelius:
Also adjusting the phase weapons due to Raz's changes to the phasers.

Cyrelius:
Ready to merge, assuming tests are successful.

PR #5580, opened 2023-01-21
Frontier Weapon Removal

VictorZisthus:
Limited ammo time, oh no.

**ALSO CONTAINS:**
Bugfix for Checkpoint flasher and doors.

Bugfix for cash registers.

Added privacy shutters and tint controls for warden, moves APC to accommodate new buttons.

PR #5579, opened 2023-01-21
Frontier phasers charge nerf, explo vendor changes, RD printer addition

Razgriz1032:
-Added frontier phasers to research console.
--Costs and requirements reflect rechargeable battery
-Adjusted printing costs for standard phaser weapons.
-Added the frontier marksman rifle to explo vendor
-Increased frontier phaser costs
-Nerfed charge time of all phasers by dividing phase_power by 8
--It will take two minutes to charge the carbine to full power.
-Change out cargo frontier phasers crates to standard phaseguns

-----
Map changes to follow - Removal of frontier phasers from station and explo carrier gun cabinets - to be replaced by standard phasers with extra weapons battery. POIs will also be swept for frontier phasers.

Intent is to increase challenge to exploration by simply not granting people the most powerful gun in the game considering it does considerable bonus damage to mobs vs most weapons. It is also ridiculously quick to recharge (faster than recharging alien batteries even). So we want to encourage players to find and use the other powerful loot weapons available within the POIs.

The frontier phasers are still available as a reward in the explo vendor and also for Reasearch to print on request. No, frontier phasers will not available for cargo - explo's gotta put in work cataloguing or tech retrieval for these.

PR #5583, opened 2023-01-21
Minor Wilderness Shelter Buff

VictorZisthus:
The wilderness shelter will still be susceptible to power outages without upgrades, but will no longer completely shut down if the doors are left open.

Buffed the starting generator's output to account for Air Alarm heating, ensuring the passive heating will not drain the entire shelter's power supply. Active charging of cells still will, rapidly.

### DIFF
--- a/maps/southern_cross/southern_cross-10.dmm
+++ b/maps/southern_cross/southern_cross-10.dmm
@@ -839,9 +839,6 @@
 /turf/simulated/floor/tiled/white,
 /area/surface/outpost/shelter)
 "SI" = (
-/obj/machinery/power/port_gen/pacman{
-	anchored = 1
-	},
 /obj/structure/cable{
 	d2 = 4;
 	icon_state = "0-4"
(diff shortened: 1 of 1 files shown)


CMDRYouth:
Other than myself, I can think of zero people who actually upgrade the power in the Shelter.

VictorZisthus:
**ACTIVE EDGES, MY NEMESIS! AAAAHHHH!**

VictorZisthus:
Also this should probably be put in at the same time as #5580 and #5579 so things are not too brutal.

VictorZisthus:
> Other than myself, I can think of zero people who actually upgrade the power in the Shelter. And “upgrading” is really just finishing the three generator frames after asking for Phoron and Uranium from Science. Stock parts can be got from the Autolathe upstairs.

Better parts make more power. Stock parts are not as great at taking care of that!

CMDRYouth:
> Better parts make more power. Stock parts are not as great at taking care of that!

Stock parts are a lot easier and quicker to get.  Usually by the average time better parts are available (+3-4hrs Shift time), people are done with the Wilderness entirely.

But aside of the point there, the only thing the Shelter really needs higher power generation for is recharging energy weapons and cells.

And doesn’t the Exploration Shuttle still provide infinite Recharger power?

CMDRYouth:
Something that’s easy to do is just scrounge a Hyper or Super Power Cell from the Station and courier it to the Shelter.  That usually sorts all power needs in the Shelter for a shift.

PR #5588, opened 2023-01-22
[MIRROR] Tilted Duffles

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/14353
--------------------
Switches dufflebags to use a tilted angle rather than being horizontal over the lower back. Includes all current non-fluff duffles. Obviously a subjective thing, but I think this looks nicer and better reflects their higher capacity compared to regular bags.

Also fixes one of the subtypes having no on-mob icons

Little preview of the north-facing on-mob icons (east/west facings are also included):
duffs

Black, engi, med, EMT
Away-team, drone*, sec, talon
Science

*The missing type!


CMDRYouth:
wooooo tilted duffle mirror

PR #5599, opened 2023-01-23
allows teshari shaped things to be picked always picked up along with teshari

Seris02:
exactly as the title says

because I want my tesh slime to be able to climb up and be picked up like a normal tesh


CMDRYouth:
Good code edits.

PR #5617, opened 2023-01-28
Altevian Clothing Fix

VictorZisthus:
Fixes the job based clothing of Altevians not having the minor stat buffs of their humanoid counterparts clothing options.

Hydroponics and Janitorial are the only other job suits that offer any bonuses, but Altevians don't have any specific special suits for those jobs; I can't fix what does not exist!

No CHOMPEdit / CHOMPFix comment because this really should be pushed downstream/upstream.


Darlantanis:
Waiting for upstream implementation.

Darlantanis:
Since it's been a year without comments added or this getting sent upstream...

VictorZisthus:
> This ever getting sent upstream?

I sent it upstream I thought.

I'll look into it.

PR #5642, opened 2023-02-03
Wires shield SMES to main grid instead of engine

Darlantanis:
Wires the shield SMES into the main grid instead of the engine, as this has repeatedly caused delaminations due to siphoning power away from the engine SMES.

If we're going to add unnecessary conveniences for more experienced engineers, let's implement it in a way that doesn't trip up new players. Engineering already has the steepest learning curve out of any department.


VictorZisthus:
Ah, the problem with this is the shields can pull the entire master grid's power supply, which will depower the entire station.

Unless we adjust master grid to be able to output 4mw, shields would need to pull from the engine grid, or be adjusted to consume far less power.

CMDRYouth:
Engineering has a learning curve?
Jokes aside, isn't it kind of dangerous to edit out every little problem that comes up?  Eventually people won't be able to learn things, or learn from their mistakes.

VictorZisthus:
Also this needs to be fixed before the merge. I think the file type is off.

image

PR #5684, opened 2023-02-07
witch hat and cyborg mask added

Echopsychotic:
Adds cyborg mask and witch hat to loadout

image
image


Darlantanis:
Flesh tail is haunting.

Echopsychotic:
> Flesh tail is haunting.

let the fur be shaved.

PR #5691, opened 2023-02-09
Hivebot and Home Update

FluffMedic:
Soo I love the hivebots but oh dear they're so freaking easy. They are so weak, and die so easily. The radiation and Ion bots feel like the biggest threat.

So added a new purple branch called Precursor.
All have a nice shield to try and protect their fellow hivebots, and hopefully pack some form off high offense. Machine Gun - Should be a stream of bullets. 
Laser - Fires a precursor beam.
Lobber - Lobs a special 'precursor' version of the energy artillery shot.

Also turned the tank class of thinks into blue, similar to ablative/durasteel although a darker color.

Might add loot tables to the precusor ones when I figure that out. And might have spelled precursor wrong somewhere.

And with the new hivebots, just plopped them into the derelict engine.

And well, I did adjust their home a tad.
-Added back the tank armor, swapped the alien translocator and muscle overclocker, and added one of each type of Precursor Hivebot. Shifted the armor to the armory, and have the alien techie stuff be in the teleporter room. If a map image is wanted is wanted feel free to ask. Or if the map updates should be removed also poke me, because I could certainly be jumping the gun with that aspect.


FluffMedic:
Ehh decided to send a picture anyway.
Not sure if you can even see the difference from it.

PR #5700, opened 2023-02-12
[MIRROR] Implements periodic tips for (new) players!

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/14456
--------------------
- ~~Using mob/living/life() to call the proc as we already send this only to mobs.~~
- ~~Chose mob/living over carbon to cover silicons and maintpred/other simplemobs.~~
- Uses a subsystem dedicated to periodic tips. Lowest fire priority, fires only during game
- By default, starts ON. Can be turned on/off at any time under preferences tab. The players are informed how to do this on the initial, 10 minute tip.
- ~~Only those mobs that have a client receive this~~
- Tries to ensure every new tip is different from the last.
- Should be easy to add new tips as time goes on, just need to balance the weights when adding a new one
- Could probably add custom per role tips, but I do not know how to check for player job/role so I went for the basics for now.
- Advice list needs significant expansion, but the current size should be good to go onto server already.
- Advice list picks from categories with separate weights to what's inside the category.
- Adds a new verb for players to request tips outside of cooldown. Players can specific tip subtype (none, general, gameplay, roleplay, lore), overriding the category randomization.
- Currently fires after 10 minutes of roundstart, and every rand(45,75) afterwards. Timer is global between players
- Everyone who did not opt out receives the same advice.

This is how it looks like:

image

(while testing, used 10 sec initial, 30-60 second range afterwards)


Razgriz1032:
I have hijacked this PR and made modular_chomp/code/modules/player_tips/player_tips_list.dm

PR #5721, opened 2023-02-14
This is what we call a difficulty tweak

Razgriz1032:
Station events rebalanced.
-Supply Demand moved from inactive to active in all three severity categories.
-Blob security assignment requirement reduced from 3 to 1.
-Meteor wave weight decreased from -110 to -90. Engineer assignment requirement reduced from 3 to 1. Added medical weight 10 and any weight 1.
-Metroid infestation weight changed from -105 to -100. Added any weight = 2. Security assignment requirement reduced from 3 to 2. If proc'd, it now has a 50% chance to fail. Spawn # of metroids increased. Spawn weights added and full bestiary (minus the queen) is spawnable. God help us all if the Omega evolves.
-Major spider infestation added. If proc'd, it now has a 25% chance to fail, but only for the major event. Spawn # of spiderlings increased. APPARENTLY I MISSED A VIRGO SNEAK WHERE NURSES WERE INDEED NOT SPAWNING, so they now spawn again. Number of possible spiderlings spawned from small egg clusters increased.

### DIFF
--- a/code/game/objects/effects/spiders.dm
+++ b/code/game/objects/effects/spiders.dm
@@ -132,15 +132,15 @@
 		qdel(src)
 
 /obj/effect/spider/eggcluster/small
-	spiders_min = 1
-	spiders_max = 3
+	spiders_min = 2 //CHOMP Edit
+	spiders_max = 6 //CHOMP Edit
 
 /obj/effect/spider/eggcluster/small/frost
 	spider_type = /obj/effect/spider/spiderling/frost
 
 /obj/effect/spider/eggcluster/royal
 	spiders_min = 2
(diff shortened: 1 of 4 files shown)


Rykka-Stormheart:
YES I LOVE THIS.
MORE DANGEROUS STATION HELL YEAH.

Rykka-Stormheart (approved):

Razgriz1032:
This is so sus that it is working on the first try wtf
(ignoring the missed ==)

CMDRYouth:
DANGER STATION WOOOOOO we can finally be robust AND IT WILL BE OKAY!

PR #5737, opened 2023-02-16
gargoyles

Seris02:
because yes

why it's good for the game:

....... ¯\\_(ツ)_/¯

someone put it in suggestions and I like the idea (and I got a gargoyle character), so, I made it happen


Sharkmare:
What if ability to adjust Statue color for more variety?

PR #5740, opened 2023-02-16
Gygaxes Mob Nerf

FluffMedic:
Tried to nerf the gygax mob's insane freaking speed. They're way too fast, and need curbing a tad. Well, so do player mechas but that is an issue for a more skilled coder or when I have less active projects and more frustration on their stronkness.
Not sure if the movement speed being increased by 1 is too little or too much, but if this gets accepted, we'll see.


FluffMedic:
Realized what may be an error. Fixing in a moment.

FluffMedic:
All should be well. AADG should still be at full strength while gygaxes are fast but not light speed.

PR #5745, opened 2023-02-17
[MIRROR] Alien slug ears and tail

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/14498
--------------------
Adds new markings ported over from Chompstation.
image
image
image
--------------------
CHOMP Notes
I noticed that there were a bunch of ckey snowflake code, so I just disabled them entirely.


Razgriz1032:
Congrats, it only took three years :P
They're good. They're knockoff goodra parts.

Razgriz1032:
All good. And that is a fine example of how modular code works out for us

PR #5755, opened 2023-02-18
[MIRROR] Adds new machine that mass-produces autoinjectors

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/14481
--------------------
**What it does**
Adds a new machine, which Science can make and also starts mapped in on all 3 maps' triage areas.

This machine takes plastic sheets OR autoinjectors and a glass reagent container.

- ~~It uses the plastic sheets to make either a normal or large autoinjector, then fills them with chemicals.~~
- It takes autolathe produced autoinjectors, then fills them with chemicals
- Alternatively, you can make autoinjectors on the spot at 3x the cost of making in an autolathe (500->1500 for large, 10->30 for small)
- It can make multiple autoinjectors.
- It can give the autoinjectors custom names.
- It warns you if the autoinjectors will contain less than maximum chemicals, letting you abort.
- It can recycle spent autoinjectors (prevents doing so while they contain chems)
- It is a wall mounted machine
- It was mapped in to chemistry & triage of each map (triage, to encourage EMTs to make them as they need them without invading chemistry)
- Alt-Clicking removes beakers, for ease of use!

**Why we need this**

- Making autoinjectors, at this time, is a massive chore I would not wish on anyone. You need to manually fill them all with syringes.
- Autoinjectors are useful for crew that are not trained paramedicine, allowing them to self-treat in space without having to break character, or go untreated.
- The reason a more ready access to auto-injectors is required is that the presence of medical staff can make or break whether people can participate in an away mission
- - Medbay staff may only have 1 person and cannot join with clear conscience
- - Medbay staff may not be interested
- - On breathable away missions, this isn't an issue as medical crate/first aid kits do everything you need and are usable by anyone.
- - However, on space-borne missions, lack of injections can almost guarantee death for the intended rag-tag groups.

**Why not just add it to ChemMaster**

TGUI scares me. 

**Mapping Locations:**

Tether:
image

SD:
image

RP:
image

**The Icon**

image
---------------------
CHOMP Notes
Not actually mapped in on our end yet.


Razgriz1032:
It's not actually mapped in, so too bad. Someone's gotta do that

PR #5772, opened 2023-02-23
Shadekin Hideout, Dark Tunneling, Shadekin death changes

TheGreatKitsune:
Trio of major changes!
First is a small Shadekin Hideout made by Kassc mapped onto the centcom z-level, which includes one of the hub Dark Portals.
- Hub Dark Portals can be set to travel to any hub or minion Dark Portal by a shadekin.
- A crewmember can touch a hub Dark Portal to set it to do a one-time trip to the station
- An outsider or simplemob can touch a hub Dark Portal to set it to do a one-time trip to Sif
image

Second is Dark Tunneling, a new ability for carbonkin and simplemob kin that lets them spawn a minion Dark Portal, which a shadekin can then open up to any hub Dark Portal for five minutes (Outside of events, this would be the hub portal in the Shadekin Hideout)
- Creating the portal requires a shadekin to have maximum Energy, and can only be done once per shadekin in a round.
image

Third is a change to Shadekin death. Now, instead of 'warping somewhere far away' and qdeling themselves on death, carbonkin or simplemob kin will instead warp to the Dark (Outside of events this would be the Shadekin Hideout) and gain a modifier called Dark Respite.
- Shadekin cannot travel through Dark Portals for 15 minutes, multiplied for each time Dark Respite has triggered this round.
- Shadekin cannot use any of their abilities or gain nutrition or energy for 25 minutes, multiplied for each time Dark Respite has triggered.
- Shadekin get enhanced healing for the longer duration as well, but only if they are in The Dark (Outside of events, this would be the Shadekin Hideout.)
- Shadekin heal half of the damage they had taken when they warp to The Dark. If this would still cause them to die afterwards, then they will do so, but in The Dark.
- Shadekin will die and leave a body if they are currently under the effects of Dark Respite, or if they are in The Dark at the time of their death.


Rykka-Stormheart:
One bug already found - Shadekin can teleport out of VR to the Dark portal, allowing VR characters to escape VR.

TheGreatKitsune:
Removed the multiplier on waiting times for consecutive triggers of Dark Respite, and made the waiting time shorter for vore deaths.
Carbonkin can manually trigger Dark Respite while in the Dark for the enhanced healing, letting them fix IB and broken bones they are suffering without access to the station.

PR #5790, opened 2023-02-25
[MIRROR] Makes speech bubbles manually selectable, adds a whole bunch of new ones

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/14536
--------------------
Idk if this will be decided as 'generally desirable' but I felt like doing this so why not

New verb in OOC tab for that. Using 'default' option forces it to select a bubble based on current standards.

Will probably put it in the char setup in separate PR if this one is deemed okay.

image


Rykka-Stormheart:
SPEEDMERGE PLEASE. SO GOOD. FOX CHAT BUBBLE.

PR #5826, opened 2023-02-28
Remove random damage on straight disposal pipes

Sharkmare:


Kazzc:
This will give me a lot more options for POI's and some generic fun for players messing with pipe.

PR #5830, opened 2023-03-01
Adds a blacklist that can modified in mapping for Designs

Sharkmare:
This allows us to define a proto-lathe to have block EX a VR proto-lathe from making Teleporters


Sharkmare:
Code compiles, runtimes are unrelated to the changes.

PR #5839, opened 2023-03-03
Buff Promethean limbs

Cyrelius:
We made it so bones don't break as easily, but seemingly forgot to do the same to Promethean limbs, which splatter at 20 damage.  This increases that to 35, making them less likely to get crippled in a single lucky hit to a leg.


Kazzc:
Most guns can already deal more than 30 damage. Will be more punishing if you don't have armor. 35 sounds good.

PR #5857, opened 2023-03-05
Adds green fox plushie!

RinPin:
[Adds](image) a green fox plushie to the game and loadout. to go with the other colored fox plushies.


RinPin:
image

PR #5864, opened 2023-03-07
Raising point gain from blindness

Sharkmare:
Blindness without a doubt is the singular most crippling trait we have, to the point where you are arguably entirely incapable of combat. Arguably much worse of an effect on the player also because it limits who you can easily interact with.

Its nowhere comparable to other 8 point traits in limitations


Kazzc:
I'd say maybe 10. 16 is a little redonculous.

Kazzc:
16 is a little to much if you ask me, I'd say maybe 10-12.
I've not mess around with that trait so I have no idea how limiting it is, but this just feels much more of a power creep for points.
Would much prefer blindness was more of a roleplay tool than outright getting very strong traits from just being blind.

Sharkmare:
> 16 is a little to much if you ask me, I'd say maybe 10-12. I've not mess around with that trait so I have no idea how limiting it is, but this just feels much more of a power creep for points. Would much prefer blindness was more of a roleplay tool than outright getting very strong traits from just being blind.

Blinding is literally crippling. and unfixable.
If you are blind all you can do is RP really, so i don't think its unreasonable to allow a lot more leniency as power gaming while blind is impossible

Kazzc:
My stance still stands between 10-12 points. Traits at their current state are overtuned for my preference.

Sharkmare:
> My stance still stands between 10-12 points. Traits at their current state are overtuned for my preference.

12 it is then sure

PR #5866, opened 2023-03-08
Removed my bloat chem, replaces with inefficient glucose recipe 

tacoguy7765093:
**New recipe: Glucose**
**1 part water, 1 part sodiumcloride, 1 part sugar**
**Makes 1 unit instead of 3**

I don't think the 'reduced nutrition' even gave less blood regen in the first place. I didn't like the snowflake chemical I made because stuff like RIGsuits couldn't even take it cause its its own chem, and the OD didn't matter cause of how fast glucose metabolizes. 

Doesn't matter balance wise because you can get like 5 glucose pens in a sweatmax and as a loadout item, and someone can already succ protein rations through a space suit, and iron pills give you more mileage per unit, this is just injectable.

### DIFF
--- a/code/modules/reagents/reactions/instant/instant_ch.dm
+++ b/code/modules/reagents/reactions/instant/instant_ch.dm
@@ -58,12 +58,11 @@
 	required_reagents = list("unsorbitol" = 1, "phoron" = 1)
 	result_amount = 2
 
-/decl/chemical_reaction/instant/galactose
-	name = "Galactose"
-	id = "galactose"
-	result = "galactose"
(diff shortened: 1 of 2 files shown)


Cyrelius:
That's fair, we'll have to figure out some other way to make glucose less OP for blood regen.

tacoguy7765093:
Honestly, it's pretty neck on neck to iron, when I did my initial tests I thought the regen was more powerful because I was testing it on a teshari with less blood.
It takes 30u to recover from ~75% blood more or less

PR #5867, opened 2023-03-08
Adds TG Dragon Wings + new chomp_modular folder

Aroliacue:
First (technically second) Chomp PR. Please comment if anything's out of order


Seris02:
there should already be a wings_ch.dmi file, better to keep them all together I think. /icons/mob/vore/wings_ch.dmi

Aroliacue:
> 

I made it after BlackMajor asked me to make a corresponding file in the chomp modular file, I'll migrate all the vore/wings_ch over when I finish fixing up some more TG wings

PR #5868, opened 2023-03-08
Removes CKEYS from the pregame lobby

Sharkmare:
Anonymizes all waiting players as "Player" instead of identifying them by their Byond key in line with our general anonymization efforts. Ideally, someone who knows how should change this to use the selected character name instead if possible

This means you can queue up as EX CMO without everyone knowing who is playing the CMO on round start.

### DIFF
--- a/code/modules/mob/new_player/new_player.dm
+++ b/code/modules/mob/new_player/new_player.dm
@@ -117,7 +117,7 @@
 			var/datum/job/refJob = null
 			for(var/mob/new_player/player in player_list)
 				refJob = player.client.prefs.get_highest_job()
-				stat("[player.key]", (player.ready)?("(Playing as: [(refJob)?(refJob.title):("Unknown")])"):(null))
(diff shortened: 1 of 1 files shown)


Seris02:
just remove that section entirely I think, no reason to put it there if there's no way to tell who it is

Sharkmare:
> just remove that section entirely I think, no reason to put it there if there's no way to tell who it is

It tells you which jobs are gonna be immediately active. and once someone adds character names it will tell you what characters are gonna be in round immediately.

Sharkmare:
Also for sake of review sanity, the runtime is unrelated but I fix it in PR #5870

Darlantanis:
Will ckeys still appear properly for admins?

Sharkmare:
> Will ckeys still appear properly for admins?

Admins have the who verb to check connected ckeys, 
The Lobby in this case is anonymized in general since we dont really need it for moderation.

PR #5880, opened 2023-03-09
Shovels can shovel off sif grass.

tacoguy7765093:
Kinda nitty gritty but using a shovel directly on growth can tile it off to expose plating, and using floor tiles on it with a shovel in the offhand will lay them on the sif grass. This allows proper building on sif, though there can be better implementation of this.
dreamseeker_nltc9DKJlE

Oh yeah, this replaces the 'dig planter' to the cultivator now.


tacoguy7765093:
I got a better way

PR #5881, opened 2023-03-09
You can put floor tiles on sif grass

tacoguy7765093:
This was a feature that got somehow lost to time from some mirror PR a year ago. I finally added it back.  Finally you can build~

dreamseeker_9tlE0Yo6c2

### DIFF
--- a/code/game/turfs/flooring/flooring.dm
+++ b/code/game/turfs/flooring/flooring.dm
@@ -151,18 +151,6 @@ var/list/flooring_types
 	build_type = /obj/item/stack/tile/grass/sif
 	has_base_range = 1
 
-/turf/simulated/floor/outdoors/grass/sif/attackby(obj/item/C, mob/user)//CHOMPedit begin, other tiles have ways to build on them, sif grass doesnt. So I put this snowflake on just sif grass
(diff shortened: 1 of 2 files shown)


code/game/turfs/flooring/flooring.dm
```diff
@@ -151,6 +151,18 @@ var/list/flooring_types
 	build_type = /obj/item/stack/tile/grass/sif
 	has_base_range = 1
 	
+	/turf/simulated/floor/outdoors/attackby(obj/item/C, mob/user)//CHOMPedit begin
```
Darlantanis, line 154: More appropriate to add this chunk to the original outdoors/attackby override in outdoors.dm, doesn't make sense to split proc overrides across files when it isn't specific to a child. You've also put this inside of /decl/flooring/grass/sif, which... I have no idea what that will do.

Splitting this way may result in your proc overriding (or being overriden by) the code chunk below. Particularly bad if it's overriding the below code, since that calls the parent proc + parent return values.

image

tacoguy7765093:
There, I made the attackby only apply to sif grass and I moved the code to somewhere better as requested.

I also accidentally discovered this behavior is _somewhere_ for the snow turfs before this PR too.

PR #5892, opened 2023-03-09
Refactored talk_sound_map

Sharkmare:
This restructueres talk sound map entirely  into a different list format because auto key fetching was messing up. debugging this took a while.


Sharkmare:
02 should be equal to the existing one on our master but should double check that, probably before merging

Issue #39, opened 2019-12-29
Add a sound toggle for emotes with sounds


Darlantanis:
Unlike many other sounds, emote sounds cannot be toggled. This issue is more pertinent now that VOREstation added many more emote sounds.

Razgriz1032:
Resolved with https://github.com/CHOMPStation2/CHOMPStation2/pull/279

PR #5893, opened 2023-03-09
less strict sanitization on some of the text based character customization options

Seris02:
title

because I\&\#39;d really like to not have to deal with html entities, or in other words: normal looking text \&gt; html entity text

### DIFF
--- a/code/modules/client/preference_setup/general/01_basic.dm
+++ b/code/modules/client/preference_setup/general/01_basic.dm
@@ -140,7 +140,7 @@
 		return TOPIC_REFRESH
 
 	else if(href_list["metadata"])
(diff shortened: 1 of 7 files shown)


Seris02:
eeeeee
I forgot to push my fixes

KillianKirilenko:
Are there any particular dangers or exploits this might open up? If not, I might upstream this because the aggressive sanitization on the chardir also bugs the everliving hell out of me.

Seris02:
> Are there any particular dangers or exploits this might open up? If not, I might upstream this because the aggressive sanitization on the chardir also bugs the everliving hell out of me.

not that I can really think of. since it deletes html brackets, you can't really do anything (afaik)

PR #5902, opened 2023-03-11
Protean health nerfing

Razgriz1032:
Re-added default protean brute and burn mods:
-Brute mod from 1 to 0.8
-Burn mod from 1 to 1.5

Nerfed protean limb health. 100 max health was way more than literally everything. Trying a balance from prometheans, making them slightly stronger than baseline promethean.
https://github.com/CHOMPStation2/CHOMPStation2/blob/master/code/modules/organs/subtypes/slime.dm
-Head, groin, chest from 100 down to 70
-Arms, legs, hands, feet down from 65 to 40.

EDIT: Changes to the groin and chest doesn't actually affect anything since they are indestructible.


CMDRYouth:
This is it folks.
The end of Protean Gaming.

Which is a good thing, good job Raz. 👍

PR #5909, opened 2023-03-12
Stop the grand escapes Part.27

FluffMedic:
Fixes some stuff that was discussed in dev chat.


FluffMedic:
I was asked to keep this open in case it became an issue once more, since there is a fix proper as a pr.

PR #5910, opened 2023-03-13
Technomancer hand teleporters no longer work in VR

tacoguy7765093:
People playing with antag equipment in VR: Good
Using it to escape VR: Cringe
This disables the hand teleporter from working in the VR Z-level
Map specific check so no runtimes if any other map is loaded.


tacoguy7765093:
oversight detected will reupload

PR #5923, opened 2023-03-13
New capture crystal type

LpSuchtie:
Adds a new (cheap) capture crystal that only works on crew, not animals or carbon mobs with no client, so making it RP only.


Darlantanis:
Looks like you didn't enable this file in the .dme. You can do this by clicking the checkmark next to your file in dreammaker (or hoping you get a prompt for it in VSC). Without enabling this in the .dme, it will not be compiled and included in the game.

I'd recommend testing this locally if you haven't already.

LpSuchtie:
Ah, forgot that. I tested it already, just forgot to enable it when I copied the repo. I'll fix it in a few hours.

PR #5948, opened 2023-03-15
fixes char directory gender, sexuality and event tag buttons

Seris02:
somehow I missed that when I fixed it up from the last PR

### DIFF
--- a/tgui/packages/tgui/interfaces/CharacterDirectory.js
+++ b/tgui/packages/tgui/interfaces/CharacterDirectory.js
@@ -71,14 +71,14 @@ export const CharacterDirectory = (props, context) => {
                   <Button
                     fluid
                     content={personalGenderTag}
-                    onClick={() => act('setTag', { overwrite_prefs: overwritePrefs })}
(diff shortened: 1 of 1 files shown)


Darlantanis:
Does this need an updated bundle?

Seris02:
> Does this need an updated bundle?

uhhh yes but afaik from all the other times I've done prs involving tgui the deployment does it by itself

Darlantanis:
I'm not aware of anything here which will regenerate the bundle on master. TGUI .js tweaks typically need to come with an updated bundle.

Seris02:
could merge it without the bundle update just to check anyway, not like it's exactly an urgent thing, can be done with an updated bundle in whichever PR does it afterwards if it actually doesn't

...but again as far as I've seen, the prs I've done that touch tgui without touching the bundle work just fine the in the next deployment

Razgriz1032:
Just run the bundle update to be safe.

Seris02:
you guys are right btw, checked it all, there's nothing to auto-update the bundle here, unfortunately it's a tg thing

Razgriz1032:
Master will need to be pulled into this PR's branch and then the bundle will need to be generated.

PR #5953, opened 2023-03-16
Adds a new HOP alt title

Sharkmare:
Adds the "Performance Management Supervisor" Alt title as an analog to the Captain's "Chief Supervisor" themed role Just like we have Facility Director and Facility Steward


Darlantanis:
That's a mouthful.

PR #5957, opened 2023-03-16
No shoes shoe option

Sharkmare:
No clue if this works, time to test it


Sharkmare:
someone should probably implement this in a cleaner way but i tried finding the code how to cleaner and didnt find it, todo: make better implementation i guess

PR #5959, opened 2023-03-16
Buffing Suppressor

FluffMedic:
Merc suppressor's threat seemed based off when velocity damage was a thing. Damage from 10 to 30

Oh and also nerfed the base bullet from 60 to 15 because some bullets are using that damage value for some reason, and really, why the was it that high? 

Looking in to trying to fix some broke guns that default to that damage, but this is in case there is more and I am not smart enough to figure it out.


FluffMedic:
Closing because someone is fixing the smg and don't want to mess with that.

PR #5967, opened 2023-03-18
better gargoyle code

Seris02:
moves the gargoyle customization verbs to trait preferences, less verbs is better
changed the way gargoyles copy overlays, so they now no longer keep any overlays except clothes and body layers
said body layers are the only things that actually have their colors changed, because why would your clothes become stone as well??
fixed an issue where going out of the statue didn't auto unpetrify

also gets rid of the code that undefines all the layer defines in human/update_icons.dm and updates the modular file to stop duplicate defines


code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -1366,6 +1370,7 @@ var/global/list/damage_icon_parts = list() //see UpdateDamageIcon()
 		update_wing_showing()
 
 //Human Overlays Indexes/////////
+/* CHOMPEdit - why are these undefined??
```
Darlantanis, line 1373: image

PR #5988, opened 2023-03-25
Changing blizzards

FluffMedic:
Reduced the alpha of the blizzard's screen fill effect from 225 to 105. If any issues arise from the way I did this, let me know. Might need help finding files though. It's a bit heftier then fog, but hopefully folks can actually see through it now.

image


Darlantanis:
This implementation will need to move the icon to a modular .dmi and redirect the weather to use the modular .dmi. Github does not allow conflict resolution with .dmi files.

Alternatively, this can be sent upstream.

FluffMedic:
Understood. Will open up a PR later with the first one, or functioning snow goggles.

PR #6000, opened 2023-03-28
Behemoth Rework and Loot

FluffMedic:
Changed the behemoth and juggernaught's reflection to be more effective against weaker projectiles but less effective against stronger projectiles. Aka nothing stops the pulse rifle. 100 - p.damage is the reflection rate.

And to future proof, stole the speed change from that major speed change PR.

The varitant behemoth uses a projectile that is the cultist laser but not hit scan, and it also drops a rigsuit now, yay. The only reason normal one doesn't is because I don't want to play with the fire that is the giant PR more then I have. The rig has behemoth's stats and a nerfed version (80) of it's reflection but a slowdown of six.


modular_chomp/code/modules/mob/living/simple_mob/subtypes/occult/constructs/juggernaut.dm
```diff
@@ -3,13 +3,16 @@
 	real_name = "Behemoth"
 	desc = "The pinnacle of occult technology, Behemoths are nothing shy of both an Immovable Object, and Unstoppable Force."
 	melee_damage_lower = 45
-	melee_damage_upper = 55
+	melee_damage_upper = 65
 	ai_holder_type = /datum/ai_holder/simple_mob/intentional/adv_dark_gygax
-	projectiletype = /obj/item/projectile/icicle
+	/obj/item/projectile/energy/electrode/cult
```
Darlantanis, line 8: projectiletype = /obj/item/projectile/energy/electrode/cult

FluffMedic: Ahh thats what I forgot. Opps, should be fixed now.

PR #6011, opened 2023-03-31
Library Cafe Changes

FoxglovePurpur:
Adds librarian access to the cafe, cafe maint door, and cafe shutters.

- Also adds a sick ass rag (for cleaning) to the cafe.

### DIFF
--- a/maps/southern_cross/southern_cross-2.dmm
+++ b/maps/southern_cross/southern_cross-2.dmm
@@ -43998,7 +43998,7 @@
 	},
 /obj/machinery/door/airlock{
 	name = "Coffee Shop";
-	req_one_access = list(25,28)
+	req_one_access = list(25,28,37)
 	},
 /turf/simulated/floor/tiled/old_tile/yellow,
 /area/crew_quarters/coffee_shop)
@@ -49209,7 +49209,7 @@
 	id = "coffeeshop";
 	name = "Cafe Shutters";
 	pixel_y = -26;
-	req_one_access = list(25,28)
+	req_one_access = list(25,28,37)
 	},
 /obj/machinery/button/neonsign{
 	id = "cafeopen";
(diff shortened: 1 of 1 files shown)


Sharkmare:
but why?

This literally just gives the librarian a job-changeless second job for no reason other than "bored while playing lib, be barrista"?

FoxglovePurpur:
> but why?
> 
> This literally just gives the librarian a job-changeless second job for no reason other than "bored while playing lib, be barrista"?

If that's the case then I'll be sure to make bartender and barista different roles.
Because as is, cafe is never used. Why shouldn't the librarian have access to helping man the cafe? Wouldn't exactly be the worst thing I've seen.

Bartenders sometimes go into the kitchen to help cook.
Chefs sometimes do gardening when they should fax for a botanist.

I don't mind if you don't like it, but it was discussed on Discord awhile ago and mostly thought to be a fun addition.

CHOMP is supposed to be fun, and it's not like they gave Bar access. *shrug

Sharkmare:
> I don't mind if you don't like it, but it was discussed on Discord awhile ago and mostly thought to be a fun addition.

This has nothing to do with me not liking it?
The chef and Bartender are both Food service roles of the entertainment block, it makes sense if they help eachother out.
A Librarian also having access to the starbucks doesnt make much sense, comparably.

But yeah if Bartender and Barrista were made seperate (even if just by giving the barrista alt title different access from the bartender main title) I think that would make it make more sense since it would reassign the responsibilities there.
 
> CHOMP is supposed to be fun, and it's not like they gave Bar access. *shrug

Im asking for a reason no need to go "why do you hate fun" basically

Kazzc:
For the purposes to ease roleplay, giving librarian access to the cafe could cause more interactions.
Just to note that job stealing is against the rules, it is expected for the librarian to let a barista do its job.
Since the cafe itself is inside the library, only makes sense for the librarian to help out when applicable.

PR #6018, opened 2023-04-02
Updates Supermatter and R-UST Manuals

FoxglovePurpur:
> Updated the Supermatter Manual with information applicable to Southern Cross
> Updated the R-UST Manual with information applicable to Southern Cross

> Added a new book sprite for the R-UST Operation Manual


Darlantanis:
This will need to use a modular file for the .dmi changes. Github does not support conflict resolution for .dmi files, so any upstream changes to the .dmi will overwrite changes here.

FoxglovePurpur:
> This will need to use a modular file for the .dmi changes. Github does not support conflict resolution for .dmi files, so any upstream changes to the .dmi will overwrite changes here.

There we go, I couldn't find a chomp specific for this already so I just went ahead and created library_ch.dmi for chomp library sprites.

PR #6026, opened 2023-04-02
Fixes infinite stack exploit

Verkister:
People have been exploiting this to generate infinite diamonds and phoron by getting 5-10 sheets from recycling one sheet. Should just buff the crusher efficiency instead.

Also might be an indirect cause for the recent 18k/h "invalid stack amount" runtime spam incident possibly caused by excess stack exploit and byond being jank with big numbers and such.

### DIFF
--- a/code/modules/recycling/recycling.dm
+++ b/code/modules/recycling/recycling.dm
@@ -149,8 +149,8 @@
 
 /obj/machinery/recycling/sorter/proc/dispense_if_possible()
 	for(var/mat in materials)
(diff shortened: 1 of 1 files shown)


Verkister:
Hmmh actually unsure if it was this thing tho. The runtimes seem to be still spamming even after resetting the cargo machines.

Verkister:
yeah pretty sure it's not this.

Verkister:
Actually it might kind of be this. This certainly is an infinite resources exploit I've witnessed recently and byond has always been weird with some numbers

PR #6029, opened 2023-04-03
People can self-operate on robotic limbs

tacoguy7765093:
**EDIT: This applies to all robotic limbs now instead of just synths**

From what was me initially tinkering around out of sheer curiosity, giving synths the ability to self-operate on themselves added some interesting mechanics to the game while making synths that _tiny_ bit less of a pain to play. It is still way better to get a friend to do it for you and here's why. Also it felt more fun playing with prosthetic mechanics without the game aggressively punishing you for simply having them.

**How do I use it?**
Enable the verb in the object menu to allow yourself to attempt synth surgery. You still need to stand on a valid operating surface like a table to attempt it. There is a 3 second delay before you start the surgery. This respects table chances, (So being on a table will only allow it to work 50% of the time.

**Might be OP?**
I was doing a bit of testing and I find myself spending a good amount of time trying to mend damage even with nanopaste, because once you hit a certain amount of damage your limbs will begin to spark and you will cancel the surgery by dropping the tool or collapsing. You are basically **soft-capped** and at the mercy of RNG to not have a surgery interrupted, combined with the lower odds of doing it outside of a operating table **and** the fact it takes **longer** to even do it compared to if someone else did it, giving more time for a random malfunction to interrupt it. **No, sitting on a chair does not prevent you from collapsing**.
And if you're at the point where you're critically damaged and constantly getting sparks its a lost cause to even try.
 My testing was only 30 minutes of attempting solo exploration and it felt like I spent allot of time repairing myself but I might have missed a edge case or two.

**EDIT: Synths cant operate on their own head, they cant see it and will need outside help**

**Why?**
IMO I feel like it adds a bit of depth to synthetic players, giving them a small but not insignificant buff that isn't something boring like un-nerfing nanopaste (Which was nerfed for a very good reason) and also as a safety net of not being completely crippled cause the roboticist is in the shadow realm. Also it makes sense for synths to be able to tinker with themselves.

**What about organic self surgery?**
No.... Just no... I tested it and its way too powerful and low roleplay cause organics don't suffer the same way synths do at high limb damage due to chems/painkillers/splints. Also it tip toes the line of the 'no suicide rule' 

Extra notes: Might be way too op with #5978, may close that PR and think of another way.

dreamseeker_UMSGVQiwdz
*note final PR does not say surgery but operating on oneself instead.

### DIFF
--- a/code/_helpers/unsorted.dm
+++ b/code/_helpers/unsorted.dm
@@ -1141,12 +1141,19 @@ var/global/list/common_tools = list(
 
 // check if mob is lying down on something we can operate him on.
 // The RNG with table/rollerbeds comes into play in do_surgery() so that fail_step() can be used instead.
-/proc/can_operate(mob/living/carbon/M)
-	return M.lying
+//CHOMPEDIT START self surgery
+/proc/can_operate(mob/living/carbon/M, mob/living/user)
+	. = M.lying
+	if(user && M == user && user.allow_self_surgery && user.a_intent == I_HELP)	// You can, technically, always operate on yourself after standing still. Inadvised, but you can.
+
+		if(M.isSynthetic())
+			. = TRUE
+
+	return .
 
 // Returns an instance of a valid surgery surface.
-/mob/living/proc/get_surgery_surface()
-	if(!lying)
+/mob/living/proc/get_surgery_surface(mob/living/user)
+	if(!lying && user != src)
 		return null // Not lying down means no surface.
 	var/obj/surface = null
 	for(var/obj/O in loc) // Looks for the best surface.
--- a/code/_helpers/unsorted.dm
+++ b/code/_helpers/unsorted.dm
@@ -1146,11 +1146,13 @@ var/global/list/common_tools = list(
 	. = M.lying
 	if(user && M == user && user.allow_self_surgery && user.a_intent == I_HELP)	// You can, technically, always operate on yourself after standing still. Inadvised, but you can.
 
-		if(M.isSynthetic())//Beep Boops only.
-			. = TRUE
-			if(M.zone_sel.selecting == BP_HEAD)//Cant see your own head to operate on it
-				to_chat(M, "<span class='warning'>You cannot see your own head to operate on it.</span>")
-				. = FALSE
+		if(M.organs_by_name[user.zone_sel.selecting])//Living people with their prosthetics can tinker with them
+			var/obj/item/organ/external/S = M.organs_by_name[user.zone_sel.selecting]
+			if(S.robotic >= ORGAN_ROBOT)
+				. = TRUE
+				if(M.zone_sel.selecting == BP_HEAD)//Cant see your own head to operate on it, mostly applies to FBPs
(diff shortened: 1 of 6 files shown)


tacoguy7765093:
**Update: ** Synths cant self surgery their head, they cant see it. Also is a spot where you must get outside help now if damaged.

tacoguy7765093:
Ok now anyone can operate and fiddle with their own prosthetics.

Kazzc:
Self-surgery may be to strong of a buff for a synth, if the annoyance of being a Synth player is needing to backtrack every time they get nicked for more than 15 damage, then we should increase that threshold. 
25-35 health can be self fixed, beyond that a Synth player needs someone else to fix them. This should drastically reduce downtime.

If downtime is the killer here, I am willing to buff Synths. There is no documentation so I don't know any of the current stats Synth suffer from internal and/or limb damage and their health stats, what requires to get it fixed, conditions and debuffs a player experience when injured.

This is what I would consider ideal. 
-Have Synth's limbs health threshold increased that it can survive 2-3 hits before needing someone else to fix it.
-Have Synth's internal components health threshold increase that it can survive 1-2 hits before needing someone else to fix it.

4-6 hits to destroy limbs
3-5 hits to destroy internal components 
Damage values of mobs/guns/environment are all over the place, so is hard to estimate what health threshold is ideal before a synth player needs to deal with broken parts. aka, downtime.
Assuming. . . most mobs deal around 25-35 DPS, adding damage reduction from traits and armor. . . 25-35hp threshold before needing someone else to fix sounds alright.

tacoguy7765093:
The reason I left this for synths was because there is allot of neat mechanics associated with prosthetics and robotic limbs that you just do not see in general, and one of the reasons is because of how much the game punishes you just for existing as a robot. So my goal of this is to make being a robot just slightly less painful while still having team play as an option. If you'd like, I could increase the delay from 3 seconds to 5 seconds, and this is considering the fact the limbs are malfunctioning. You are likely not doing this at an operating table, and you're in a situation where you are alone and can't call for someone to assist.

tacoguy7765093:
And my second comment about damage values, honestly with such a powerful buff here, it's really more in retaliation due to the power creep of damage values in general, it's more of a plaster on a bigger problem as is a lot of other buffs as of recent. Because in normal ss13, 15 damage is respectable but not severe, but here its considered minimal with a lot of values lagging behind (Like the threshold)

KeRSedChaplain:
IIRC the threshold for synth limbs being irrepairable is 30, but most mobs deal 10 through 20 damage for a hit, projectile or otherwise. To explain a little for Kassc, the surgery itself isn't the main issue, it's the fact that synths can barely self-sustain themselves when compared to an organic character, where the only thing they have to worry about is bone breaks (which don't prevent organics from healing a limb to full, unlike synths) and IB. Self-surgery is very strong, yes, but the point at which synths would have to have a surgery to fix a limb is also when they'd drop things in their hands or fall over from a busted leg.

tacoguy7765093:
That was my other point I was going to bring, unlike organics which have some semblance of being able to help themselves, You are defenseless baby as a synth. At least with this PR you will be subject to RNG but have a safety net regardless.

PR #6040, opened 2023-04-05
Makes recycling more exciting

Verkister:
And much more dangerous with a new kind of garbage mulcher that basically takes in pretty much any item, even crates/lockers, and chews it the fun up, popping all its contents out to be swallowed next, ensuring everything within gets also recycled for optimal efficiency. Also applies oofy brute grinds on mobs and gibs dead bodies, slurping up the gibs and splatters right after. Complete with animations and noises asset remixed out of laziness to a decent enough result.

Has an emaggable safety switch that reacts to humanmobs with mind datum, shutting down the thing and sending an alert message on supply channel so that someone may come pull out whoever's stuck there and manually restart the machine.

Originally envisioned overhauling the entire recycling system, but kinda got lazy and ended up just making the new piece an addon to utilize the existing system for further processing.

Also added an adjustable var for disposal outlets to determine the range of the stuff it shoots out, and made the one at disposal chute launch only 1 tile, no longer blasting it all directly into the windoor with fat sound/message spam.

Added item digestion blacklist and a filter feature for this machine to release the trapped blacklist items contained within with a crowbar.

Removed the monkey animation from the default gib proc.


Verkister:
CI xenoarch moment as usual

PR #6047, opened 2023-04-06
Smacks the naughty artifact master

Verkister:
*moan

Liberates CI checks from that one runtime error we all love.


Darlantanis:
YES

PR #6060, opened 2023-04-09
Religion Defaults list changes

Sharkmare:
Removes some IRL religions from the Default list of religions, primarily to encourage the custom ones present and for peeps to just define worldbuild their own religions.

These can still be used if wanted via free text entry but personally don't see much point in having them in there much less at the top of the choices

Keeping in the vague ones that are technically based on IRL stuff but aren't actually religious movements ex Pleromanism etc

This change is just a suggestion with the above reasoning.

### DIFF
--- a/code/modules/client/preferences_factions.dm
+++ b/code/modules/client/preferences_factions.dm
@@ -95,13 +95,13 @@ var/global/list/antag_visiblity_choices = list(
 	)
 
 var/global/list/religion_choices = list(
-	"Unitarianism",
-	"Neopaganism",
-	"Islam",
-	"Christianity",
-	"Judaism",
-	"Hinduism",
-	"Buddhism",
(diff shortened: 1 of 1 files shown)


Darlantanis:
Unsure if these should be outright removed given irl is canon to this setting (just, 500+ years old at this point).

Do agree that they shouldn't top the list, at least not in our server.

Sharkmare:
> Unsure if these should be outright removed given irl is canon to this setting (just, 500+ years old at this point).

Thoughts behind removal are that anyone can easily just input any IRL religion of today so a drop-down with them isn't really needed, VS the non-real ones that peole have to even learn they exist first.

Actually by that path i can also remove Deism and Agnosticism since they should already be known things

Thoughts?

Razgriz1032:
There is 0 reason for this. Needless change that may just cause headache in future if upstream decides to change things too.

Raeschen:
closing this, a bit semantics and overly limiting

PR #6072, opened 2023-04-10
Couple more recycling tweaks

Verkister:
Added a separate button for garbage grinder that can be pressed without needing to open the windoor.
Fixed AI/borg interaction with the grinder.
Added a var to toggle the old recycling crusher's manual feeding, which the grinder will toggle off for it. (borg modules will no longer be eaten out of their slots)
Made the crusher start upgraded to full efficiency.
Also fixed the funned up poster near the entrance.


Darlantanis:
conflicts

PR #6073, opened 2023-04-10
Adds new mechanic for liquidbellies

Verkister:
Liquidbellies now splash prey/items with their reagents upon entering the belly, strength scaled to reagent amount and divided by amount of belly contents. Reagent amount affects slosh sound volume played when things enter the wet vorgan.
Also added a few new belly reagent options suitable for splashing or other effects, these being stomach acid, lube, biomass, and fertilizer. Felt pretty balanced on test runs what comes to usefulness or abusefulness concerns.
Also made reagent unacidable checks respect the item digestion blacklist.
And fixed digestion death moving prey's reagents directly into bloodstream instead of ingested metabolism.

### DIFF
--- a/code/modules/reagents/reagents/dispenser.dm
+++ b/code/modules/reagents/reagents/dispenser.dm
@@ -375,7 +375,7 @@
 	if(ishuman(M))
 		var/mob/living/carbon/human/H = M
 		if(H.head)
-			if(H.head.unacidable)
+			if(H.head.unacidable || is_type_in_list(H.head,item_digestion_blacklist)) //CHOMPEdit
 				to_chat(H, "<span class='danger'>Your [H.head] protects you from the acid.</span>")
 				remove_self(volume)
 				return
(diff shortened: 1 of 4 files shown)


Verkister:
Yeah pretty much. Decided to leave it non-specified to acid because extra spaghetti and prommies+water is also a thing it should protect from. It's just a splashing mechanic so non-digestable fols aren't missing out on too much. Gonna do one more tweak tho, which is making the acids not splashif the belly is not in item digest mode.

Darlantanis:
As much as I hate the bloat, wondering if gut splashing needs its own pref. Reagents are variable enough that I'm not sure any preexisting pref covers it adequately, and for whatever reason liquid belly mechanics seem to squik many players interested in trying out our server.

Verkister:
Sounds like those players might prefer playing on virgo anyway.

Verkister:
Thankfully I'm pretty good at reading between the lines and don't need to put much effort or spaghetti to rectify the concern.

PR #6080, opened 2023-04-13
Drone code restructure

Sharkmare:
Moves drones into the modular folder to allow for more violent changes without having to worry about upstream.

Restructures some of the drone code, merges our own changes as well as virgos extra file into the main files, makes swarm a sub folder in the drone folder to make it a bit nicer to see what belongs to where.

Is it perfect? no. did i modify EVERY file? no, some didnt really need adjustments (yet).

This is simply to get the ball rolling since this means we can now change drone code as violently as we want with no regard for upstream (not that anyone touches drone code anyway)


Sharkmare:
Untested for now gotta make sure everything still works since i heavily restructed the attackby proc

PR #6082, opened 2023-04-14
Actually unifies the Unified Drone Fabricator.

VictorZisthus:
Poor mining drone was forgotten. NO MORE! This re-adds them to the list for the unified style Dronefab.


Darlantanis:
conflicts

Sharkmare:
The mining drone is not part of the default list on purpose.

The list of available drone is added to and removed from beyond the defaults in mappings, no drone was forgotten

PR #6091, opened 2023-04-14
Fixes mob holders being a bit too slippery

Verkister:
Fixes mob holders dumping held mobs prematurely, and also fixes the dropped empty holders lingering for a tick before vanishing


Seris02:
pr upstream perhaps?

PR #6094, opened 2023-04-14
[MIRROR] Concussion Maul

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/14767
--------------------
I wanted to try making a spicy hammer, so I made a spicy hammer. Behold, the concussion maul!

The concussion maul is a slow but powerful melee weapon that must be manually charged up prior to every attack for maximum effect. Whilst it can deal ruinous damage and throw targets away from the wielder when charged, its very slow swing speed, manual charging between attacks, limited number of uses per cell (4 from a standard weapon cell), and sheer size make it difficult to use effectively. Uncharged, it has no knockback or stun effect.

It also gains increased knockback distance when used in disarm mode, with the standard disarm-mode drawback of halved damage.

Finally, it also has limited destructive and industrial applications, as it can easily smash through certain obstacles whilst charged, and can pulverize most rock terrain. It *will* destroy artefacts though, and this still consumes power from the cell.

Low-quality demo GIF:
ezgif-3-be08175925

TODO:
- [x] basic code/functionality. works as intended.
- [x] balance. should probably use 2h wielding mechanics.
- [x] sprites, could maybe do with something more here.
- [x] sfx, functional but could use some spicing up.
- [x] actual mining/destructive functionality?

~~In draft mode so it doesn't get accidentally merged before it's ready.~~

In summary: a slow, powerful melee weapon that allows you to absolutely body single targets in exchange for being weak against groups of enemies. Doesn't have the raw killing power of a fire axe or katana, but has slightly more utility so long as you have enough power cells to spare.


Darlantanis:
sure why not

PR #6113, opened 2023-04-17
Reworks belly fullscreen file storage

Verkister:
Divided the main vorgan fullscreen overlay file into separate dmis, and also added their extra overlays into those dmis.

The old file with all states moved to the same subfolder for archival purposes, while the file in its former place, which the system still uses to pull the iconstate options and vorepanel thumbnails from, reduced to a less volatile non-animated 120p resolution.

This should fix the bug with all the fullscreen options vanishing from vorepanel serverwide, caused by the large dmi filesize choking byond on host level during roundstart inits.

Also fixed a typo in the rodent maw's tongue color overlay, so that one should now also be available.


Darlantanis:
bless

PR #6114, opened 2023-04-18
Fixed double cyborg charger

FluffMedic:
Fixes the double cyborg charger.
image


FluffMedic:
image
I also made the button in this room work.

PR #6116, opened 2023-04-18
Reduces search time for trash piles

Aroliacue:
Quality of life - Makes searching trash piles faster.


Razgriz1032:
Impatient gamer PR

PR #6123, opened 2023-04-18
Makes liquidbelly acid splash use digest_act

Verkister:
Liquidbelly acids now utilize digest_act proc for in-belly item splashing/melting, and now also support partial digestion when acid amount is under meltdose threshold.
Instead of being a solid 30u meltdose, it's now 10u per w_class (30u for normal size item).
And since it now uses digest_act, storage items and such get to dump their contents instead of just taking it all with the old melt qdel.
This also means the in-belly melted items get to make their final gurglenoises as well.
Also lowered nutrition cost for generating stomacid, making it a little more in line with the prey damage potential and item meltdose reagent consumption.

### DIFF
--- a/code/modules/reagents/reagents/dispenser.dm
+++ b/code/modules/reagents/reagents/dispenser.dm
@@ -374,36 +374,38 @@
 		return
 	if(ishuman(M))
 		var/item_digestion = TRUE //CHOMPEdit Start
+		var/splash_mult = 1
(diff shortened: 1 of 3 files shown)


Verkister:
Probably gonna close this one in a bit because I kinda added it into the other active PR's test build before adding more stuff on top and I'm too lazy to snip it out of there to avoid conflict

PR #6125, opened 2023-04-19
Belly acid splash overhaul x2, Swoopie neck speed fix, TURBO MODE addon, dynamic liquidbelly visuals, and a bunch of other tweaks and fixes.

Verkister:
Liquidbelly acids now utilize digest_act proc for in-belly item splashing/melting, and now also support partial digestion when acid amount is under meltdose threshold.
In-belly acid splash melt effect changed to use 10u per w_class (30u for normal size item) instead of hardcoded 30u for every item.
And since it now uses digest_act, storage items and such get to dump their contents instead of just taking it all with the old melt qdel.
This also means the in-belly melted items get to make their final gurglenoises as well.

Lowered nutrition cost for generating stomacid, making it a little more in line with the prey damage potential and item meltdose reagent consumption.

Fixed "speedy" vorgans not getting removed from SSbellies processing list (the cause behind the inconsistent transfer times on swoopie neck), by simply making them init directly for SSobj without touching SSbellies subsystem in the first place, and also switched those from mob processing to obj processing to reduce lag.
This also means that the feature can be used by any mob type without needing extra code in their life proc.

Added TURBO MODE belly mode addon, utilizing the forementioned mechanic.

Optimized some mob processing as well, mainly by making simplemob carboncorpses, monkeys, brainmobs, and observers low priority processing for the mobs subsystem.

Added items_preserved list reset for brute/burn damage setting changes, fixing the issue of item digesting belly not doing anything when changing the damage up from 0/0 level.

Fixed liquidbelly entrance slosh sound runtimes caused by somehow ownerless vorgans and eaten non-copied lists.

In-belly acid splash no longer affects mobs when not in digest mode. And is now also removed from the mob to prevent the acid suddenly kicking in the moment they're released.

Also swapped swoopie's main gut fullscreen to a more appropriate one.

Also massively overhauled the acid mechanics, details below.

- Reagent splashing on mobs now prioritize splash over touch (since splash applies touch effect in it anyway)
- Initial reagent entrance splash amount reduced, because it now consumes the reagent, and now the additional reagents are applied passively over time while stewing inside the full/generating gut.
- Acid digestion now fully works with features such as nutrition gain percentage and digest reagent generation.
- Acid digestion potency now scales according to the amount of acid inside the gut, divided by the amount of other contents, and depletes it according to the applied effects.
- Acid reagent interactions with non-human mobs now function correctly.
- All things fully support digestion settings and prey prefs and item modes. (no non-dispersed digestion though)
- Acid gurgles are not affected by existing digestion damage settings though, and can be used together or without them. The new type of digestion basically works by the speed of your reagent generation and amount, and does dispersed damage effects on all contents, mobs included. (Lower liquid max capacity lowers max damage potential and higher max capacity increases it when filled)
- Stomacid generation cost lowered to water level due to now being purposed as a method of gaining nutrition via digestion.
- Stomacid can no longer be dispensed into containers however, fully eliminating the potential for external splash griefs.
- Liquidbelly digestion reagent generation now works and actually gives more than the old 0.00scraps for digestion time gains.
- Liquidbelly digestion reagent generation now returns the surplus gain as nutrition for belly owner when capacity is full.
- Items touched by belly reagents no longer fill "reagent containers" at all, and the acid can now digest them without contaminating the snack's own nutriment contents.

Also now there's a mechanic that applies dynamic liquid level indicator for the prey to see how wet their spot is. An animated bubbly overlay using the liquidbelly reagent color var that steadily climbs up from the bottom of the screen according to the current liquid level in relation to the custom max volume, and its opacity scales according to the size of custom max capacity. In other words, a smaller capacity belly will fill up faster while having much thinner overlay, while a large capacity gut will have a thicker overlay and less speed.
A toggle for the effect can currently be found in preferences tab ingame. Proper savefile pref and pred-side toggle will be added to vorepanel later after the current tgui conflict queue has been cleared.

Added diluted digestive acid as reagent option for slower scenes (-80% power) as the old stomacid was quite potent for its job.

Made new visual overlay for the liquids and adjusted the values accordingly. Also now includes a non-bubbly version for hold/hold mode.


Verkister:
fat fuck pr title

Verkister:
Testing one more thing i might add to this...

Verkister:
that'll do

Verkister:
about to shove in another chunky commit

PR #6132, opened 2023-04-20
Feedable belly preference

Fracshun:
Adds a new belly specific option to disable it from appearing for others during feeding prompts.


Fracshun:
Let me know if anything needs to be done about that tgui.bundle.js conflict. It was generated when running tgui-prettybuild.bat and the new tgui option doesn't appear without the file change, but it could be excluded from the PR if it'll be regenerated later. Just need to know what the process is for tgui edits.

Darlantanis:
Bundle is always necessary with tgui changes; it is the tgui compilation file which runs tgui during station rounds, so it must be updated for tgui changes to take effect. 

The one important part with generating the bundle is making sure your branch is up-to-date with master. Otherwise new tgui changes will be overwritten.

PR #6141, opened 2023-04-24
Grand Library SP-pod

Kazzc:
-Shutters and button use ID:"glcafe"
Credits go to "FreaRose"

Better way to abate winter storm but to cozy inside a wooden library, a lot of books to read to pass the time!

image
image


Kazzc:
More SP-pods!

PR #6144, opened 2023-04-25
Removing Hitscan, and curving some things

FluffMedic:
Made a new projectile/mob folder, for mob projectiles mimicing real ones. Hit scan weaponry is not fun to play against.
Pirates, mercs, drones, mechs, and hivebots all affected.
Nerfed the hivebot laser to use the same thing as a drone, since the hivebot bullets deal 10 damage as well.

Part two of the PR, curving the health of mobs.
Dire/Sec wolf now have 100 HP. Why did they have 300 before?
Instead of a bunch of variance by 10 and 15, a lot of spiders have been put into two categories, tanky and disruption. The first has 130 and the other has 70.

I also sort of reverted the nerf of the gygax now that mob movement works plus mob slowdown
image


Kazzc:
I like what you are doing here, very promising changes.

Darlantanis:
Good by me. Mob hitscan projectiles are nasty.

PR #6146, opened 2023-04-26
Tiny css tweak

BlackMajor:
Kassc asked. Colourblind can't see this purple in dark mode.

Should probably add colourblind compatibility at some point.


Kazzc:
>.>"

PR #6150, opened 2023-04-27
Add a confirmation with guidelines on emergency shuttle usage

TheGreatKitsune:
Big message but hopefully people actually read it before hitting yes
image

### DIFF
--- a/code/modules/tgui/modules/communications.dm
+++ b/code/modules/tgui/modules/communications.dm
@@ -273,9 +273,10 @@
 			if(!is_authenticated(usr))
 				return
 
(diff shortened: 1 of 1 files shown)


VictorZisthus:
I would change the I to "we."

Otherwise looks good!

TheGreatKitsune:
Updated the message based on headmin request:
```
OOC: You are required to Ahelp first before calling the shuttle. Please obtain confirmation from staff before calling the shuttle.

Are you sure you want to call the shuttle?```

Kazzc:
Good stuff

PR #6153, opened 2023-04-28
Fixing some minor things #Lizzard

FluffMedic:
Removed the double toilet in in the 3rd deck giant bathroom. As in the two toilets in the same space. Swapped the escape pod nanomeds for emergency medical wall lockers, because of the malfunction machine event.

And removed the random space tile in outsider pod restaurationbar.dmm
image


Kazzc:
Good stuff!

PR #6155, opened 2023-04-29
Secondary Auto-Transfer

Fracshun:
Adds in a secondary location and chance for auto transfers. When objects transfer they will individually attempt the primary location first, and failing that it will attempt the second. If both fail the object's transfer cooldown starts again.


Darlantanis:
Needs rebased to master and bundle regenerated, otherwise good to merge

PR #6169, opened 2023-05-01
Terrible. Adds a handheld way to matrix colour anything.

BlackMajor:
Test merge only for a bit while I figure out what a good limitation is for this if it's added to autolathes.

Title.
Adding a matrix painter item that can paint stuff with the new matrix colouration.

I'll probably add a verb to simple mobs in this same PR to let them recolour themselves


Raeschen:
you want tgui/packages/tgui_ch, not tgui/packages/tgui. We disabled tgui and use tgui_ch now. and also a rebase and bundle rebuild

PR #6187, opened 2023-05-11
A bunch of vacpack and vorgan juice fixes

Verkister:
Vacpack now checks for spont prey pref.
Fixed vacpack direct targeted sucks missing their after-click doublechecks. (the cancel stuff for fleeing distance and things leaving/yoinked from the tile)
Fixed liquidbelly acid item interaction order (now does digest_act stuff first if in belly)
Fixed open containers eating up all liquids.
Fixed vorgan vac destination not checking if vorgan owner is present or adjacent.
Fixed massive piles bypassing any capacity checks on trashbag mode.

Also added new auto-setting for vacpack, which makes the vacpack use the smallest level required to suck the target, making it nice and quiet for floor dirt while still able to suck up bigger things if needed.


Verkister:
Also adding an auto-setting feature for the vac, pushing once tested.

PR #6193, opened 2023-05-11
[MIRROR] That's a thing people will probably do

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/14850
--------------------
Makes it so mob forgiveness only works if the mob isn't in a belly


Darlantanis:
Relies on #6189

PR #6206, opened 2023-05-19
Mapping icon fixes

Kazzc:
Some minor icon duplication.
``/obj/structure/marker_beacon`` ``/obj/item/weapon/reagent_containers/hypospray/autoinjector`` ``/obj/item/seeds/vanilla`` Should no longer have invisible/missing sprites on map format.
image


Darlantanis:
Base .dmi files cannot be modified without said modifications being sent upstream.

PR #6229, opened 2023-05-22
Minor map fixes

VictorZisthus:
**Z2**

Fixes missing atmos vents and scrubbers in Research server room.

Fixes Atmos custom mix input and output.

Adds ShieldGen bypass breaker boxes.

Actually gives our brave first responders (Atmos Techs and Engineers) some gas masks... And some more inflatables.

### DIFF
--- a/maps/southern_cross/southern_cross-2.dmm
+++ b/maps/southern_cross/southern_cross-2.dmm
@@ -115,7 +115,7 @@
 /obj/effect/floor_decal/industrial/warning{
 	dir = 1
 	},
-/obj/structure/cable{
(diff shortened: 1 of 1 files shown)


VictorZisthus:
image

"This stuff is pretty simple. The red wire is the Master grid. The yellow wire is direct engine output. The blue wire heads to the shield generator's SMES. By default, the shields will charge from master grid, but this breaker station is here for those folks that like to power the shields straight from the engine. One of these stays on, the other stays off. DO NOT HAVE BOTH ON AT ONCE, OR BAD THINGS WILL HAPPEN."

- The note

PR #6234, opened 2023-05-24
VR tweaks and mob transformation

Darlantanis:
Tweaks virtual reality code and adds some new features.
-Rewrites vr console code to not rely on an aheal to spawn into VR
-Adds a gross proc to delete any mobs which escape from VR
-Fixes a bug where dying in VR while TFed into a mob makes you die in real life
-Adds the ability to join VR as a mob through the VR console
-Adds the ability to join VR as a ghost via a shiny new button
-Grants VR players a resize verb
-Grants VR players a mob TF proc to transform into mobs from a maybe-curated list

Less VR-relevant
-Moves living_ch.dm code to modular_chomp folder
-Adds a general purpose transformation proc for possible non-VR future use

image


Kazzc:
I been asking for all these features, lets hope nothing breaks!

PR #6235, opened 2023-05-24
Fixes saddlebag slowdown.

Verkister:
Fixes saddlebags having twice the slowdown of duffles despite being same size.


Darlantanis:
Implemented in #6245

PR #6246, opened 2023-05-25
Fixes wolftaur 3-colour loafing sprite

Aroliacue:
Removes two pixels in the down-facing position.


Darlantanis:
Any tweaks to upstream .dmi files must be submitted upstream; github does not allow conflict resolution for .dmi files.

PR #6257, opened 2023-05-27
Caps Downside of Biofuel Processor Traits

VictorZisthus:
If species.synthetic_food_coeff is = 1, then charging is efficiency is zero while still draining power. Likewise values greater than 1 will result in power LOSS while charging. This fixes both issues.

Minimum charge efficiency is now 10%.


VictorZisthus:
https://github.com/VOREStation/VOREStation/pull/14905

PR #6262, opened 2023-05-28
[MIRROR] Adds new human ability: "Explain Custom Traits"

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/14887
--------------------
**What this does:**
Gives all human mobs (ergo: anyone that's not a simple mob or a silicon) a new ability. 
Unless they do not have any traits (at which it returns and notifies them they have an empty trait list),
Pressing this ability brings up a dialogue box listing choices of TGUI, print to chat or cancel.

Unless user cancels, a list of keys is generated from the name vars of each custom trait their species has (confirmed working for custom and non-custom species alike), which is based completely on the character creation window choices. A number of associative lists are likewise populated using the name var of the given trait we are iterating over as the key for easier retrieval. These lists are used for both to_chat and tgui functionalities

**If the user picks the to chat option**, a SPAN_NOTICE formatted message is sent to their chat with line breaks. If the dev did not define tutorial for their trait, the user is recommended to go to #dev-suggestions and offer their own guide.

**If the user picks the TGUI option,** they are presented with a screen similar to the following:

image

Not shown: a scrollbar appears if the text or number of traits are longer than the window.

Furthermore, the text in the right element wraps around automatically, removing need to worry about line lengths for developers.
The only thing developers must keep in mind is to use <br> for their traits over \n. This is reinforced twice - once in the trait_tutorial.dm comment at the top, and once for the var/tutorial itself.

**Why we need this:**

It has often come up before as mentorhelps, #cadet-academy questions and during addition of more complexly behaving traits that just the trait's desc var does not provide sufficient information for proper usage.
While the desc var could be made more descriptive, it is not accessible by users during gameplay easily AND it should be short and and optimized to guide user choices, rather than a detailed description of the respective mechanics. 
While we do have a wiki page for traits, downstreams may lack it causing potential confusion for their players, and our own players often forget about the wiki's own existence. Plus, large walls of text that some traits may require may scare users off of the wiki.

**Commit Contents**
[Adds new human ability: "Explain Custom Traits"]
* New verb of category "Abilities"
* New var for /datum/traits: tutorial
* Prints to_chat: Name, Desc, Category, Tutorial
* var/tutorial defined as a string by default
* Developers are to use <br> over \n when defining subtypes of tutorial
* bloodsucker obligate/freeform have a tutorial defined as example.

[Adds TGUI functionality to Trait Tutorial](https://github.com/VOREStation/VOREStation/pull/14887/commits/cd602eb20e5394f3ad069396b2fedcdb5bd12549) 
[cd602eb](https://github.com/VOREStation/VOREStation/pull/14887/commits/cd602eb20e5394f3ad069396b2fedcdb5bd12549)
* Creates trait_tutorial_tgui.dm to act as backend, adds to vorestation.dme
* Creates TraitTutorial.tsx to act as frontend
* Modifies traits_tutorial.dm for TGUI compatibility
* Adds comment to var/tutorial definition on how to make it pretty in TGUI
* Modifies logic in trait_tutorial to catch empty trait lists.
* Edits bloodsucker obligate/freeform to add another <br>


Darlantanis:
Need to merge #6231 first

PR #6265, opened 2023-05-29
Add inshoe steppies

purplefoxy27:
Based on one of the `ss13-suggestions-and-feedback-forum` posts, I attempted to bring the _steppies_ mechanics to in-shoe interactions. These will also only be enabled if both prey and pred have stepping mechanics enabled in the vore panel, else it will revert to old in-shoe messages, with its awful `prob(1)`.

All damage and effects are based on https://github.com/CHOMPStation2/CHOMPStation2/blob/master/code/modules/vore/resizing/resize_vr.dm#L260, but with an increased weaken timer to account for the time needed to take your boots off and shake the prey from shoes.

Additionally, I added a few tips to remind people that the vore panel has preferences for the steppies too!

Finally, this is my first PR with additional features! Let me know if I missed anything.

Here is how it looks: 
- **Recording**: [OneDrive cloud link](https://1drv.ms/v/s!AkF92os5E5O3kKEZ3xRrf-GX4Xy8bg?e=1y1tal)
- **Pred Side**: 
image
image

- **Prey Side**: 
image

### DIFF
--- a/code/game/turfs/simulated.dm
+++ b/code/game/turfs/simulated.dm
@@ -108,7 +108,7 @@
 			if(H.shoes)
 				var/obj/item/clothing/shoes/S = H.shoes
 				if(istype(S))
-					S.handle_movement(src,(H.m_intent == "run" ? 1 : 0))
+					S.handle_movement(src,(H.m_intent == "run" ? 1 : 0), H) // CHOMPEdit handle_movement now needs to know who is moving, for inshoe steppies
 					if(S.track_blood && S.blood_DNA)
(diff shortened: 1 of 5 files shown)


Kazzc:
Can a player OOC-escape from this?

purplefoxy27:
@Kazzc, **HECK**. It turns out we couldn't, not even in the current `master` branch. I assumed it had when I started the PR, my bad and thanks for the reminder!

But I think I got it now! I followed the other OOC escape situations and works locally (Added a VV link to help investigate if needed, too).

PR #6295, opened 2023-06-02
Alt Version Of Deep Wilds POIs

FluffMedic:
Adds some alternate versions of some Deep Wilds POIs. Whether some mobs are swaped out, locations swapped, terrain altered, intended path changed, etc.


KeRSedChaplain:
won't this mean multiple of the same PoI would spawn

FluffMedic:
No

FluffMedic:
There is a variable that puts POIs in a group and only one of those group can spawn.
Examples
-The radio shacks
-Blueshuttle down and blackshuttle down

Kazzc:
Good stuff! Can always use more POI's!

PR #6296, opened 2023-06-03
Trying to make amends.

FluffMedic:
I tried to created boss mobs and difficult foes.
Made them too lethal.
I'll try to properly write stuff down in the morning but basically

Eclipse -> All shootslower, and deal less damage. Less deadly grenade
Phazon -> Missle destructive ranges deceased
Boss Hivebot -> Singulairty attack changed priortity, new hivebot summon attack Precursor hivebot -> All can drop minor things, health decreased Phazon - Base version no longer does rad attacks, instead does basic explosiove/fuel rod shots...actually I will likly change this in the morning.


FluffMedic:
Back to amendments of previous work.

Eclipse mobs
-Yeeted a higher delay for melee mobs.
-The hound and wheel got their range attack removed.
-All special projectiles took at least a 10 Point hit in AP or/and damage.

Phazon
Boss-Explosion radiuses nerfed.
Non:boss-uses the large toxic bolt instead of my idiotic idea of a fuel rod cannon.

Chrono Hivebot
-It's loot pool has chances icnreased, and slightly increased with the special translocator.

Precursor Hivebot
-Now have chances to drop alien junk, and/or a cell

PR #6305, opened 2023-06-03
[MIRROR] Little Cleanup

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/14937
--------------------
This only is a little code cleanup in the code/ directory.

To prevent accidental running or compiling into 515, a if-statement with an error message was built into the __byond_version_compat.dm file.

It also offers new defines that become relevant when testing or moving on to version 515 (or maybe for Downstreams if they´d like). 

Additionally, __513_compatibility.dm was removed and the relevant contents put into other files, since the if statement is not reachable as the minimum version is defined as 514.


Raeschen:
#6330 replaces this

PR #6318, opened 2023-06-04
stops vox npcs from showing up on sensors

Seris02:
title


Seris02:
eeeee, that failure was not a fault of mine

PR #6329, opened 2023-06-05
[MIRROR] More 515 Compatibility

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/14976
--------------------


code/modules/events/infestation.dm
```diff
@@ -87,26 +87,70 @@
...
+/datum/event/infestation/proc/count_spawned_vermin()
+	. = 0
+	for(var/mob/living/simple_mob/animal/M as anything in spawned_vermin)
+		if(!QDELETED(M) && M.stat != DEAD)
+			. += 1
 
+// If vermin is kill, remove it from the list.
+/datum/event/infestation/proc/on_vermin_destruction(var/mob/M)
+	spawned_vermin -= M
+	GLOB.destroyed_event.unregister(M, src, PROC_REF(on_vermin_destruction))
```
ItsSelis, line 153: This block of code seems to be on upstream, while here these lines are missing.
They were not apart of the 515 Compat Update.
(The update merely only should update single lines, mostly replacing stuff like ./proc to PROC_REF etc.)

```suggestion
/* CHOMPedit - Upstream Code, not implmeneted here
/datum/event/infestation/tick()
	if(activeFor % 5 != 0)
		return // Only process every 10 seconds.
	if(count_spawned_vermin() < vermin_cap)
		spawn_vermin(rand(4,10), prep_size_min, prep_size_max)

/datum/event/infestation/proc/spawn_vermin(var/num_groups, var/group_size_min, var/group_size_max)
	if(spawn_locations.len) // Okay we've got landmarks, lets use those!
		shuffle_inplace(spawn_locations)
		num_groups = min(num_groups, spawn_locations.len)
		for (var/i = 1, i <= num_groups, i++)
			var/group_size = rand(group_size_min, group_size_max)
			for (var/j = 0, j < group_size, j++)
				spawn_one_vermin(spawn_locations[i])
		return

// Spawn a single vermin at given location.
/datum/event/infestation/proc/spawn_one_vermin(var/loc)
	var/mob/living/simple_mob/animal/M = new spawn_types(loc)
	GLOB.destroyed_event.register(M, src, PROC_REF(on_vermin_destruction))
	spawned_vermin.Add(M)
	return M

// Counts living vermin spawned by this event.
/datum/event/infestation/proc/count_spawned_vermin()
	. = 0
	for(var/mob/living/simple_mob/animal/M as anything in spawned_vermin)
		if(!QDELETED(M) && M.stat != DEAD)
			. += 1

// If vermin is kill, remove it from the list.
/datum/event/infestation/proc/on_vermin_destruction(var/mob/M)
	spawned_vermin -= M
	GLOB.destroyed_event.unregister(M, src, PROC_REF(on_vermin_destruction))
*/ // CHOMPedit End
```

PR #6330, opened 2023-06-05
[MIRROR-FIX] Little Cleanup

ItsSelis:
Closes #6305


Darlantanis:
Probably more appropriate for world.name to be "CHOMPstation" rather than "VOREstation"

PR #6369, opened 2023-06-15
Adding DB Capability for Mentors

ItsSelis:
This enables Database capability for getting mentors when the admin legacy system is not enabled.

This will require a little database edit. The DBSchema.sql file has been also edited accordingly.

Tests:
- I was able to use mentor tickets as just a mentor, no other permissions. Admin tickets were still invisible like regular players.
- Using "Special Verbs > Unmake-Mentor", I was able to remove my ckey from the database

Sadly I could not test out adding myself because me having permissions to make someone mentor already makes an if statement true. But as the other test work, this should work fine too since its a very similar logic.


Raeschen:
@Razgriz1032 i'll nudge you to run the DB command and then merge this

PR #6383, opened 2023-06-16
Minor armor update on Snowfield

harry3124:
Adjusts armor value of lootable armor in armory


FluffMedic:
Whyyyyyy.

harry3124:
Oh. I originally thought it would be invincible, but I see that someone has already tweaked its armor value. Also, there's a bug, so I have to close this anyways, even if I wanted to push. Sorry!

Aroliacue:
These values are already incredibly excessive and on-par with gear like the Unathi Breacher suit. Considering the current state of 'balance' here, making more meta exploration gear isn't ideal in the least.

PR #6388, opened 2023-06-16
adds a paw and foot vore overlay

kcin2001:
what it does on the tin, heres a pic of each

VBO_foot
foot
 
VBO_paw (with example coloring)
paw


FoxglovePurpur:
Never knew I need this in my life. *nod

PR #6394, opened 2023-06-17
Finding a Fix for Active Edge failures

Aroliacue:
First it was stairs, now it's this.


Aroliacue:
Issue identified, it's the valley POIs causing it. Passed on to Jasper.

PR #6396, opened 2023-06-17
Fixes some improper turf things

FluffMedic:
Does what the title says, or more apptatly makes more tiles match temp, nitro and oxy.


FluffMedic:
Those aren't mine. . .will likly fix in a bit

PR #6418, opened 2023-06-20
General's Chicken/Wing Basket fix

radiantAurora:
Fixes the recipes for General's Chicken and Wing Basket through removal of redundant code.


Raeschen:
comment out unwanted upstream code please, don't just delete it. Use //Chomp REMOVE tags around the area.

Otherwise this makes merge conflicts

PR #6440, opened 2023-06-23
Makes Winged Flight a Neutral Trait

FoxglovePurpur:
We aren't DELTA airlines, let us fly for free. /j

- Commented out Positive Trait Winged Flight 0 point 1 trait
- Added in Neutral Trait Winged Flight 0 point 0 trait

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/traits_vr/neutral_ch.dm
+++ b/code/modules/mob/living/carbon/human/species/station/traits_vr/neutral_ch.dm
@@ -80,6 +80,20 @@
 	autohiss_exempt = list("Vespinae"))
 	excludes = list(/datum/trait/neutral/autohiss_tajaran, /datum/trait/neutral/autohiss_unathi)
 
+//CHOMP Note from purdev, we're MRP and having wings that you can't use is kinda stinky. :3
+/datum/trait/neutral/winged_flight
+	name = "Winged Flight"
+	desc = "Allows you to fly by using your wings. Don't forget to bring them!"
+	cost = 0
+	has_preferences = list("flight_vore" = list(TRAIT_PREF_TYPE_BOOLEAN, "Flight Vore enabled on spawn", TRAIT_VAREDIT_TARGET_MOB, FALSE))
+//	custom_only =
+
+/datum/trait/neutral/winged_flight/apply(var/datum/species/S,var/mob/living/carbon/human/H)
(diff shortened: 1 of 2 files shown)


Aroliacue:
Sexy update

Raeschen:
maybe do the same to the xenochimera flight trait too, that's the same i think

Kazzc:
Flight has mechanical advantages, don't change the cost.

FluffMedic:
The mechanical advantages are so slim though.
You fly over like 5 things, and all five of those things are rare to some extent.

Aroliacue:
Most locations don't get much use out of flight as-is, and the benefits of using it for an advantage also come with negatives (flying into any solid object will knock you over, more nutrition use, can't cross water or lava tiles any faster than usual unless there's an open z-level above you). There's also vore options for flight, which is handy for scening or accidental vore as it triggers on bump with no extra steps. It's also restricted to wings, and while you can equip any accessory in the 'wings' tab to gain those effects, it's still easy to tell who really shouldn't be flying, and who should.

I think that in its current state, flight is absolutely just a flavour trait and should be relegated to the neutral trait category. If it could allow you to pass lava and water tiles or didn't cause you to get stunned by moving into any dense turf, I'd consider it a positive.

FoxglovePurpur:
Flight has a technical positive impact in the same a lot of the neutrals do I feel. Seeing as how flight is something rarely used for "robust gameplay", and is often more used as an excuse for funny scenes and such, I feel as though it should be an option of "if you have a character with wings, you should be able to use them".
This is an MRP vore server and a lot of folk seemed to be excited about this chance on Discord (and the few here), is there any way I could add a limitation to the neutral version that'd make you okay with having it for the fun of players?

Raeschen:
i think i agree with kassc. I see it being used in explo for certain advantages at times. i'd err on the side of keep posi/one-cost, buut i'm not holding a huge opinion on it.

SpadesNeil:
I can nyoom over all danger in the wilderness because there is a little known flight level above it.

But even so, I think it would be nice to have it be made free. I can't get _indoors_ to anyplace using flight trait, although it would be super nice if POIs had a second Z level to play with.

Raeschen:
negative feedback from heads, so this one's a no i guess

SpadesNeil:
Eh it's probably for the best.

PR #6450, opened 2023-06-23
Adding belly sprites to the Rift Magnus

FluffMedic:
I just yoinked the player belly sprites since that made sense for a humanoid mob.
image


FluffMedic:
Well then...that was unexcepted.

PR #6451, opened 2023-06-23
Event buffs: Meteors and Spiderlings

Razgriz1032:
-Meteor spawn random multiplier maximum increased from 3 to 4
-In addition to spiderlings, there is now a 10% chance of spawning an egg cluster.


Aroliacue:
I hate this, please speedmerge

PR #6453, opened 2023-06-24
[WIP] Airhack

Razgriz1032:
Yeah webedit. Ported from old chomp. Original code by jfmherokiller
This is the air mechanics inflation code.

This works as a neutral trait *I THINK* but this is untested at the moment. This used to be integrated into the vore panel at one point, but that was before big scary TGUI.


Seris02:
please
air_attackby

Kashargul:
Will be redone as component in the future https://github.com/VOREStation/VOREStation/pull/18727

PR #6470, opened 2023-06-28
MORE RANDOM Events

FluffMedic:
-An alt version of the spider event that can gib a larger spider then normal 
-An event where some dangerous intruders have snuck onto station, some in secure places.


Aroliacue:
I really wouldn't recommend adding new sec-focused random events, at least while the discussions around Southern Cross being a 'safe zone' for that sorta stuff (particularly spider infestations)

FluffMedic:
Oh someone commented on this.
Anyway closed this myself out of personal reasons.

PR #6471, opened 2023-06-28
Modernizes the nuke code

tacoguy7765093:
By request from Kassc for event rounds, the nuke now behaves more like the other codebases.
Changes:
New sprite, ported from /tg/
The nuke makes a beeping sound if it counts down
The alert level will auto set to Delta if activated, lowering down to Red if disarmed.
The nuke alert sound is a global sound, if you hear it, the round is ending in 10 seconds anyway.

I would have made it so if the nuke explodes OFF station, it wouldn't end the round, but I the code that handles the cutscene kinda scares me so uhhhh... it ends the round regardless.
dreammaker_hiGiJZ6OHO


VictorZisthus:
That works. An admin can always delay the round end until a code fix can be done. Cheers!

PR #6472, opened 2023-06-28
Southern Cross: The Armory Update

SlaysInCode:
Hello! Local Security Dragon here with yet another Security update!

This one involves out lovely, and at this moment, rather poorly defended Vault! This update gives not only a joyful layer of protection, but also grants us more RP chances and a new area.

What has been done:

- The Vault is now equipped with a Cycler, much like the one leading into Xenobiology, or Virology.

- Turrets have been installed for an extra layer of protection.

- A Security Checkpoint located just outside the Vault for better protection. Security Camera Console and Records on-site, making it another good post for our lovely Security! The Checkpoint is also equipped with two lockdown switches. One for the Vault itself, and one for the checkpoint.

- A Lounge just outside the checkpoint for those with business there or those who just want a new place to relax.

- Two new \area\'s have been added. Vault Checkpoint, and Vault Lounge.

- The escape pods that were once across from the vault were relocated to where the old gym was.

Images:

Vault1
Vault2
Vault3
Vault4

### DIFF
--- a/maps/southern_cross/southern_cross-1.dmm
+++ b/maps/southern_cross/southern_cross-1.dmm
@@ -455,9 +455,6 @@
 "abT" = (
 /turf/simulated/wall/r_wall,
 /area/maintenance/firstdeck/forestarboard)
-"abU" = (
-/turf/simulated/wall/r_wall,
-/area/construction/firstdeck)
 "abW" = (
 /obj/structure/closet/wardrobe/grey,
 /obj/item/weapon/storage/backpack,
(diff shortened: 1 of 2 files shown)


SlaysInCode:
Ignore the fact it says Armory, it's Vault, lemme see if I can edit that

CMDRYouth:
Well that's fantastic, now Evacuees from the Medbay will have to cross potentially hazardous sections of the Station in the event of a major emergency to reach an Escape Pod.
You really don't need the Checkpoint, but the turreted cycler looks great.

SlaysInCode:
I don't know what route you're taking from Medbay to get to Escape pods, I can think of about 4-5 other escape pods that'd be better for medbay. such as the large one directly under medbay. As for the checkpoint, it provides another Security post for use. As there's no real logical patrol pattern for that area, and it is a vital protection area.

TheGreatKitsune:
I might already have plans that would conflict with the security checkpoint part of this. But we will see what happens.

Razgriz1032:
You can drill into the vault from the ceiling and no one's the wiser.

Razgriz1032:
It's kinda cool, but also kinda roundabout. You might be better off just putting turrets inside the vault which would address all issues immediately.

Razgriz1032:
Actually no, putting turrets inside the vault would invalidate the spiders. Hmmm

Kazzc:
I supervised this in DM's with ``SlaysInDutch``, I got no complaints but curious about any feedback.
Likely merge it soon, regards of other mapwork going on the station. DM me directly and we can get around merging map works.

PR #6497, opened 2023-07-04
Public Circuitry MAP CHANGE

FoxglovePurpur:
Honestly, this is a QOL change.
I've gotten a tad sick n' tired of SCI staff taking the public circuitry kit in maint instead of just using the one in the Circuitry lab.
This moves it over to cargo's area, above disposals there's a new circuitry kit, a few things laying around, etc.
Also, I've deleted and replaced the electronics stuff above medbay where the kit used to lay with some stuffs to peak at for you with grubby hands. :3

### DIFF
--- a/maps/southern_cross/southern_cross-2.dmm
+++ b/maps/southern_cross/southern_cross-2.dmm
@@ -17990,11 +17990,12 @@
 /turf/simulated/floor/tiled,
 /area/quartermaster/office)
 "bdR" = (
-/obj/machinery/alarm{
-	dir = 4;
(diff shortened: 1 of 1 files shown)


FoxglovePurpur:
I'm not sure what ...
> !! FAILURE !! [MAP: Active edges (all maps)]: Maps contained 2 active edges at round-start.
... refers to? I mainly just moved around pre-existing things. I **did** move one atmospherics alarm. So perhaps that's causing an issue? But it seemed to work fine in my test build.

PR #6502, opened 2023-07-05
Trying to make wilds POIs modular

FluffMedic:
Disabled the normal wilderness, and wilderness_area files, copied and pasted it, and the POIs into a modular folder.
it complied just fine, and doubt it would break anything.
I may have misunderstood the intial issue though, I need more sleep.


FluffMedic:
Ehh reopening this because most of the POIs have been touched and altered, and having wilderness POIs in two diffrent places would cause headaches.

Raeschen:
i don't know if this is the best way to do it. If code changes, i have to manually check every PoI. best to just modularise stuff we actually change, or maps only for us

Razgriz1032:
Dunno if necessary. We should be rejecting ALL upstream map changes anyway. Like, all of them.

Issue #4931, opened 2022-09-14
[Integrated Circuits] No button in assembly to actually remove the device cell.


xnqm8h:
#### Brief description of the bug
No button in assembly to actually remove the device cell.

#### What you expected to happen
There to be a button for this.
https://github.com/CHOMPStation2/CHOMPStation2/blob/207203850e5008c457fcc714e763b1866cb50736/code/modules/integrated_electronics/core/assemblies.dm#L112-L121

#### What actually happened
There just isn't a TGUI button for the action so you can't actually remove the device cell to insert a better one.

![image](https://user-images.githubusercontent.com/91218429/190091569-79460c13-b15f-4f5e-b0be-d9be45a9d463.png)

PR #6535, opened 2023-07-08
Adds interface options for cell removal and renaming assemblies.

xnqm8h:
Adds the buttons to hook up to the actions for removing cells and renaming circuit assemblies.


xnqm8h:
This FIXES #4931

Raeschen:
this won't apply till the tgui bundle is rebuilt, but i'll do that later

PR #6536, opened 2023-07-08
Changing too_much_tox (phoron) to too_much_tox (toxin).

FoxglovePurpur:
Changing the too_much_tox alert to be actually "too much tox" and not "too much phoron".
> I do plan to come back to this and make it so phoron and nitrogen breathers will get a custom too_much_tox_*oxy* alert. But for now I was just tired of the general "too_much_tox" alert assuming it's phoron. With human crew that's fine since phoron is what causes tox.  But when there are phoron/nitrogen breathers and oxygen is what's causing tox, seeing a warning saying it's phoron makes no sense. (Especially if a phoron breather is choking and their alert says it's because they're breathing phoron and they need to get an oxygen tank)

### DIFF
--- a/code/_onclick/hud/alert.dm
+++ b/code/_onclick/hud/alert.dm
@@ -135,10 +135,10 @@ The box in your backpack has an oxygen tank and breath mask in it."
 	desc = "You're not getting enough phoron. Find some good air before you pass out!"
 	icon_state = "not_enough_tox"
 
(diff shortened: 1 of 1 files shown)


FoxglovePurpur:
Github seems to have an error about something not ending with a ".
I left the code as written with the /
denoting the line break or whatever. If you'd rather I can just remove that and leave it as a single wrapped line instead.

PR #6570, opened 2023-07-13
Active edge banishment

Raeschen:
gets rid of many active edges to clean the integration checker up.

Also fixes a runtime in holder2.dm


Aroliacue:
Sexy, I like

PR #6579, opened 2023-07-15
[MIRROR] Giant Bow (Also fluff item for verysoft)

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/15144
--------------------
It's a giant bow, you can equip it on your head or on your ear

I made this mostly as a fluff item for me, because of my character's ears, but I included a public version that anyone can use that doesn't have the ear cuts! The public version also doesn't have a south facing sprite, just because it would 100% interfere with basically every kind of hair and ears except bald.

image

https://forum.vore-station.net/viewtopic.php?f=26&t=2341


Raeschen:
don't worry, the ckey locked thing is in a file not included

PR #6593, opened 2023-07-17
Secretary Alt-Title: Bridge Officer

Aroliacue:
A new alt title for Command Liaisons, inspired by one of Rykka's older Virgo PRs. Purely fluff with a unique outfit, all files modularized.

:cl:
add: Adds alt job title for secretary: Bridge Officer
add: Adds some start gear for Bridge Officer title
/:cl:


Raeschen:
Is giving an alt job title their own specific gear a thing really done anywhere else?
This is the only thing that seems dubious if so.

Aroliacue:
> 

It's just the clothing outfit you spawn with. The main thing there is the outfit_job_name which both sets the job alt-name and also lets us change what they spawn with, if anything, for a little more redundancy. There's two roles that use something similar, both of which are service-civilian, but as it gives you no extra items and is completely overwritten by loadouts, it should be fine.

PR #6610, opened 2023-07-18
Markings and tail.

Jesseman1:
## About The Pull Request

Updating female orca markings and tail.

## Changelog

:cl:
add: Added new orca tail and markings.
fix: Updated female orca markings to display as intended.
/:cl:


Raeschen (changes requested):
~~Your icon_states are wrong. The four markings in sprite_accessories_extra_ch are all set to "orca", they should be set to the corresponding icon_states in the icon file: 

variable "icon_state" is for the state of each icon in the file.
variable "icon" is for the icon dmi path (e.g "icons/mob/human_races/markings_ch.dmi")~~

Nevermind i looked into it more. this is correct! apologies

Raeschen:
also these icons appear to be for a body (i.e. a species), not markings.

PR #6612, opened 2023-07-18
Stunbatons Disrupt Solargrub Larvae

Aroliacue:
🆑
 balance: Stun Batons can now disrupt solargrubs infesting machinery, similar to a multitool.
/🆑


VictorZisthus:
I like the premise, but we could also just allow other modules to perform the same function!

Heh... Stun baton the light...

Raeschen (changes requested):
Make cyborg batongs zap solargrubs from objects similar to the multitool instead seems to be the preferred method

Raeschen (approved):

PR #6658, opened 2023-07-26
Remove some named content

Raeschen:
## About The Pull Request
Enforcing our rule of no user-specific content more.


Zeldazackman:
It should be ended on line 183 instead of 201 due to andrews not being a personalized mob 59WKuiV

PR #6677, opened 2023-07-30
[DNM] Tweaks whitelist code

Darlantanis:
DO NOT MERGE, other non-code staff work must be completed first.

Re-adds functionality to jobwhitelist.txt by commenting out a return value which forced whitelist checks to always fail. Creates job_whitelist_overrides.dm to control whitelisting rather than the whitelisted_positions check (nice one Jon, you fixed hardcoded whitelists by hardcoding the whitelist). Said return value is used for guest jobbans... which isn't needed here as guest accounts are banned. If you want to re-enable that and also have the whitelist work, have fun untangling the spaghetti. 

@Raeschen Many modified files here are missing // CHOMPedit comments, seemingly carried over from CHOMP1. I added the ones I spotted but very well could be more.

Documentation thing
Whitelist works by checking the whitelisted job "title" var (datum/job/jobposition) against the job title in jobwhitelist.txt. Changing the title var without updating the whitelist WILL break the whitelist. Futureproof fix would be replacing the title var with defines + however many hardcoded uses of the title var with said defines, then converting the whitelist to use a list of ckeys for each whitelisted position. This would make a mess of whitelist files, but no more whitelist breaking if/when somebody decides to change the title var. Until then, I guess just remember to update the whitelist if the var changes.

## Changelog
:cl:
-Replaces booleans in configuration_ch.dm with TRUE/FALSE defines
-Tweaks whitelisted positions
-Adds whitelist_only vars in new job_whitelist_overrides.dm file
-Comments out return value of guest_jobbans proc which broke the jobwhitelist system.
-Adds numerous missing CHOMPedit comments
-Adds jobwhitelist.txt to gitignore
/:cl:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,5 @@ temp.dmi
 
 node_modules/
 package-lock.json
+
+config/jobwhitelist.txt
--- a/code/controllers/configuration_ch.dm
+++ b/code/controllers/configuration_ch.dm
@@ -3,9 +3,9 @@
 
 
 /datum/configuration
-	var/discord_restriction = 0
-	var/use_jobwhitelist = 1
-	var/emojis = 1
+	var/discord_restriction = FALSE
+	var/use_jobwhitelist = TRUE
+	var/emojis = FALSE
 
 	var/vorefootstep_volume = 75	//In future see about making a function to adjust volume serverside in config.txt, easy to do with reenable values. - Jack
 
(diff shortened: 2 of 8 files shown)


Raeschen:
when it comes to chompedit comments, 'accidently' adding one where not necessary is better than missing them. It's easy to see in future updates if said code was upstream anyway, but less easy to see if uncommented code SHOULD stay and be ours. 

This appears to be well commented enough.

PR #6700, opened 2023-08-01
Adds support for simple taur bellysprites

Verkister:
## About The Pull Request
Taur bellysprite system no longer breaks if a taurtype doesn't have sprites for every size state or struggle anims. Simple bellysprites given to Naga, Naga Alt, Horse, Cow, Cow Alt (new addition i found in my old files and decided to include), Lizard, Synthlizard, Feline, Slug, Drake, Otie, and Deer.
Some of them come with loaf variants too.
Also new dual-color cow alt taur ported from old cit files I found while getting the old alt naga bulges.

## Changelog

:cl:
add: Added support for simple taur bellysprites.
add: Added a bunch of those simple bellysprites.
/:cl:


icons/mob/vore/Taur_Bellies.dmi
Raeschen, line 1: editing upstream file not preferred, better to _ch our own (if it IS upstream. Even if it isn't, it violates the "_ch or modular_chomp" rule)

Verkister: I'm pretty sure this is a chomp file even though I have no idea why it's not named as such. Yeah just checked and its ours

PR #6757, opened 2023-08-09
Fixes Parry Shield, Buffs Energy Sword

Aroliacue:
:cl:
balance: Parry shield belt to last twice as long, but takes an extra second to recharge.
fix: Cell can no longer be removed from the parry shield belt.
/:cl:
:cl:Upstream
add: Zone based marking customisation
/:cl:


Razgriz1032:
No sword damage buff.

Aroliacue:
> No sword damage buff.

Changed, I will look into other ways of making melee more suitable for exploration content

PR #6834, opened 2023-08-18
Misc loadout clothing additions

RinPin:
## About The Pull Request
adds a few new clothes to the loadout
A new cyberpunk jacket reskin
a new pair of jeans
two new duty gloves reskins

## Changelog

:cl:

add: adds a few new clothes to the loadout.

/:cl:


Raeschen:
Clothing with armor values in the loadout?

No.

Raeschen (changes requested):
Loadout items cannot have armor values

RinPin:
But we have items in the loadout that have armor? These are a child of said items, the duty gloves and the cyberpunk vest both have armor and both are avaliable in the loadout for 2 - 3 points each, though the armor itself is negligible

Raeschen:
Are they job-specific? It can be ok if job-specific. If not, we need to review them and possibly change them.

You also haven't //ChompEDIT commented chomp-specific code in upstream files.

RinPin:
The items in question aren't job specific, anyone can grab the cyberpunk jacket and duty gloves from the loadout, which makes me confused as to what the issue is, since my PR is effectively just reskins of thoses.
as for the chomp specific edits, i'm unaware of what those would be, these are just a child of the original base item.

Raeschen:
See https://github.com/CHOMPStation2/CHOMPStation2/blob/master/.github/CONTRIBUTING.md

Raeschen:
> The items in question aren't job specific, anyone can grab the cyberpunk jacket and duty gloves from the loadout, which makes me confused as to what the issue is, since my PR is effectively just reskins of thoses. as for the chomp specific edits, i'm unaware of what those would be, these are just a child of the original base item.

Just because items currently exist that have armor, doesn't make it ok for you to add more. We'll deal with these in time. For now, this won't be accepted until the items are made non-mechanically advantageous, and that Chomp-EDIT specific code tagging practices are used.

RinPin:
Does any of that applies to this? the chomp edit things,  could you explain to me what exactly here warrants the comment, since the base items already have it?

RinPin:
they should be cosmetic only now, and i hope the comments are what was asked for.

Raeschen (changes requested):
Editing upstream sprite files is also not really preferred. This kind of change is probably something that should be done in modular_chomp, since edits of the same already exist there.

PR #6855, opened 2023-08-22
Fixing a POI

FluffMedic:
## Changelog

:cl:
remap: the hadzard zone POI areas have better placed smoke clouds, and better loot.
/:cl:


FluffMedic:
...why is it no longer saving my titles in gitkraken?

PR #6867, opened 2023-08-25
Fixes planetary lighting

Darlantanis:
Fixes planetary lighting bug from #6852 . Previous if() checked numbers against planet datums, so it always failed and thus Sif lost its sun.

:cl:
-Fixes planetary lighting.
/:cl:


Raeschen:
obsolete due to #6868 ?

PR #6902, opened 2023-09-01
Adds event/shenanigans vareditable shapeshifter mechanic

Verkister:
## About The Pull Request
Basic HowTo: Staff can spawn a mob to be the shapeshift form, and then varedit the target user to add verb "shapeshift_form" (shows up in abilities tab) and then edit the target user's "tf_form" var into a mob reference of the spawned mob. Once done, the user can swap between forms kind of blobform style. TF rays will revert the shapeshift, but won't delete the shapeshifted form so the user can still shapeshift back.
Tested to work with regular TF mechanics on top of this stuff with no conflicts or issues so far.
Now also supports ckey swapping with an occupied tf_form mob when applicable.

## Changelog

:cl:
add: Added event/shenanigans vareditable shapeshifter mechanic
/:cl:

### DIFF
--- a/code/modules/mob/mob_movement.dm
+++ b/code/modules/mob/mob_movement.dm
@@ -302,6 +302,10 @@
 	if((direct & (direct - 1)) && mob.loc == n)
 		my_mob.setMoveCooldown(total_delay * SQRT_2) //CHOMPEDIT
 
+	if(!isliving(my_mob)) //CHOMPAdd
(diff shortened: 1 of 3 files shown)


Verkister:
adding another little feature to enable ckey swapping with an occupied tf_form mob when applicable

Verkister:
done

PR #6934, opened 2023-09-06
A few more shapeshifter tweaks

Verkister:
## About The Pull Request
Shapeshifting and other mob transformations (vr/metamouseguns/etc) now have a quick visual effect pop between the forms.
Mob TF now also carries shared nutrition pool between forms. (with the exception of multiplayer shapeshifter stuff)
Fixed new mob TF happening without init_vore proc, causing the user's main vorgan to be overwritten by the voremob's default vorgan settings.
Also removes mind from reverted non-multiplayer tf_form mobs, preventing cryo/gateway from double-announcing the user and their stored form upon leaving.

## Changelog

:cl:
add: Added visual effect and shared nutrition to mob TF mechanics.
fix: Fixed mob TF with uninitialized voremobs overwriting user's main vorgan with the mob's own settings.
fix: Stored reverted non-multiplayer tf_forms no longer cause double cryo announcements when user leaves round.
/:cl:

### DIFF
--- a/code/modules/vore/mouseray.dm
+++ b/code/modules/vore/mouseray.dm
@@ -93,6 +93,7 @@
 		if(!M.allow_spontaneous_tf && !tf_admin_pref_override)
 			return
 	if(M.tf_mob_holder)
+		new /obj/effect/effect/teleport_greyscale(M.loc) //CHOMPAdd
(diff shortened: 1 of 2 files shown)


Verkister:
one more coming up

Verkister:
done

PR #6935, opened 2023-09-06
Holster and Vest Availability

Aroliacue:
:cl:
add: Vests and the bluespace badge no longer have specific role/job requirements, to facilitate more character customization.
add: Added in a new holster option: left leg holsters. These just visibly sit on your left leg.
balance: Increased the loadout cost of vests and drop pouches from 1 to 2, as they carry more than the simple webbing.
fix: Fixed black holster sprites not showing up on mob correctly, and also not showing up in the loadout.
/:cl:


Razgriz1032:
No on letting all roles take holsters and vests. They should remain job specufic which is realistically within reason.

I honestly do not know what a bluespace badge is tho.

The rest is great.

Aroliacue:
> No on letting all roles take holsters and vests. They should remain job specific which is realistically within reason.
> 
> I honestly do not know what a bluespace badge is tho.
> 
> The rest is great.

Bluespace badge is what the name suggests, a badge that has bluespace storage (same total space as the vests)

By vests I mean accessory vests, i.e. the vests with more space. All of these can be readily found around the station, at pre-spawn locations and also as maint loot and explo loot. For the most part they're nothing special. If we're being realistic, they should probably be more difficult to find, but going through every loot option and changing probabilities is a lot more work than just turning the problem on its head and opening up.

Nothing besides the holsters is restricted, and even the holsters aren't 'restricted' per se as you can still see anything in them, so it shouldn't infringe on open carry rules.

If you do wish for them to remain locked to jobs, I'll reduce the price back down as all the departments can find these vests in their lockers anyway, the only reason they'd choose to spawn with it is laziness or forgetting to stop by security (events, for example)

Darlantanis:
I don't mind the vests, that's one extra small item in storage space compared to the currently available-to-everybody webbing.

No on holsters. Gun storage available to every role implies every role should have a gun to store, which is very much not the case.

Unsure about bluespace badge. I assumed this was intended for security, but code doesn't seem to support that.

Aroliacue:
I'll re-add restrictions to the holster.

I'm mainly basing this entirely on gameplay rather than roleplay, if people wish to pay those extra points for storage, or save the points and go find it as maintenance loot (or getting it for their respective), they now have that opportunities. Medical, security and exploration have the benefit of getting the vest storage for free rather than paying a loadout cost for it.

PR #6938, opened 2023-09-06
Wake up babe new markings

radiantAurora:
Adds a set of three new markings aimed toward glass-bellied characters:
Internal Organs - Digestive
Internal Organs - Heart & Lungs
Internal Organs - Ribcage


Kazzc:
Finally, organs!

PR #6941, opened 2023-09-07
Updates organ markings

radiantAurora:
Updates markings added by #6938 with a brighter greyscale to improve colouration.
Also fixes the markings so they actually work.


radiantAurora:
Drafted - There appear to be other problems with the markings so I'll use this PR to fix those issues too.

radiantAurora:
> Drafted - There appear to be other problems with the markings so I'll use this PR to fix those issues too.

thanks to the power of actually testing things this time around all issues with the markings have been fixed.

PR #6986, opened 2023-09-14
Small synth buff, as a treat

tacoguy7765093:
## About The Pull Request

Increases the repair cap from 30 to 50, id do 15 but '45' is a weird number.
Synthetics needed a small little help-me-up especially in a codebase that's mechanically intensive if you choose to be intensive. Giving them an extra hit to play with so far seems fair, especially since you already have the disadvantage of no chems.

Edit: I also realized we already had our "broken limb" threshold set to 60, but a 7 year old upstream PR made a variable that overrode it for synths only. If requested, I can remove that var and make it 60, but that might be too much buff now IMO

## Changelog

:cl:
balance: Repair threshold for synthetic repairs bumped up by 20
/:cl:

### DIFF
--- a/code/__defines/damage_organs.dm
+++ b/code/__defines/damage_organs.dm
@@ -47,7 +47,7 @@
 #define DROPLIMB_BURN 2
 
 // Damage above this value must be repaired with surgery.
-#define ROBOLIMB_REPAIR_CAP 30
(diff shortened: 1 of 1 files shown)


tacoguy7765093:
I am gonna close and resubmit this in a new PR, mostly cause of something else I discovered and I want it to appear in the discord.

PR #7056, opened 2023-09-28
Funny Breathing Traits

FluffMedic:
Adds two new traits.
One lowers how much air you need, and the other rises it. Normal - 16 pressure needed
Light - 12 pressure needed
Heavy - 22

They're all worth 1 point of their positive/negative type. The reason heavy breather is much more of a change is because I rather reduce the number of 'Hahah funny trait never comes up and is free' mindset traits.

Also messes with traits that don't work properly with synthetics.

## Changelog

:cl:
add: Two minor traits that mess with how much air your character needs to breath.
/:cl:

### DIFF
--- a/modular_chomp/code/modules/mob/living/carbon/human/species/station/traits/negative.dm
+++ b/modular_chomp/code/modules/mob/living/carbon/human/species/station/traits/negative.dm
@@ -1,2 +1,13 @@
 /datum/trait/negative/deep_sleeper
 	custom_only = FALSE
+
+/datum/trait/positive/deep_breather
+	name ="Deep Breather"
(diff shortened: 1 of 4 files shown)


modular_chomp/code/modules/mob/living/carbon/human/species/station/traits/negative.dm
```diff
@@ -1,2 +1,13 @@
 /datum/trait/negative/deep_sleeper
 	custom_only = FALSE
+
+/datum/trait/positive/deep_breather
```
Darlantanis, line 4: Is this intended to be trait/positive?

PR #7059, opened 2023-09-28
Mail fixes, loot table updates and less junk mail

TheCaramelion:
## About The Pull Request
Makes it easier for Cargo to sort through all the incoming mail by (this time properly) giving colours for the departments to the envelopes and making the junk ones white.

## Changelog

:cl:
add: Colours for each department
qol: Removed random colours for junk mail
qol: Bigger money quantities are a little bit more rare now
/:cl:


TheCaramelion:
I've got some feedback about Chef/Bartender, I'll be adding them as soon as I can instead of making another PR just for them

code/game/objects/items/mail_ch.dm
```diff
@@ -145,8 +134,7 @@
 
 	var/datum/job/this_job = SSjob.name_occupations[new_recipient.job]
 	if(this_job)
-		if(SSjob.get_primary_department_of_job(this_job) && department_colors[SSjob.get_primary_department_of_job(this_job)])
-			color = department_colors[SSjob.get_primary_department_of_job(this_job)]
+		color = SSjob.get_primary_department_of_job(this_job).color
```
Darlantanis, line 137: SSjob.get_primary_department_of_job(this_job) runs a proc, so adding .color to the end causes dreamchecker to fail. You'll probably want something like var/relevant_typepath = SSjob.get_primary_department_of_job(this_job) then take color from that var (assuming the proc works that way).

TheCaramelion: Should be solved now

PR #7091, opened 2023-10-10
Lowering Projectile Speed

FluffMedic:
Travel time increased from 0.8 to 3.
Why?
Simple, I want combat to be more skilled base and more open to folks. No need to make uber tank char #300, just practice reaction times and such, and you'll be seeing medical a lot less...or do build uber tank #300 if that's your play style.

Granted, unsure how much of a difference this number shall make in letting folks take the dodge route more casually, but it should help.

It also makes it so you do need to lead your shots a tad more.

Oh and whilst I was at it, nerfed the speed further of one mob that is based off this entire slower moving projectile thing.

## Changelog

:cl:
balance: changes projectile speed from 0.8 to 3.
/:cl:


Darlantanis:
May have unintended consequences due to things like emitters using projectiles. Not aware of any non-combat cases where projectile speed matters, however.

PR #7101, opened 2023-10-13
Let Teppi charaters use teppi

FluffMedic:
Fun fact, we have teppi.
Also fun fact we have several teppi charaters.
I am only half surprised that teppi was unseable.
Although I may have missed a memo, if I did oh well, here is me trying to let non-staff teppi use teppi.

## Changelog

:cl:
add: Teppi is public use
/:cl:


Darlantanis:
Would be preferable to remove this from teppi mobs if we want this to be a player-speakable language

PR #7144, opened 2023-10-27
Changes Swoopies to only swoop on grab, along with QOL and bugfixes

ReoDaProtovali:
## About The Pull Request

Started out as a quick change to make swoopies' integrated vac-pack only activate when they're on grab intent, but changed to also do some QOL and bugfixes that I wanted to do a while ago but never got to.
Also Holy shit there's more changes than I thought there were after going over them again _WHEW_

## Changelog

:cl:
add: Added spawn message for player swoopies that let them know how to control the suck birb
add: Moved the description part informing the player how to use swoopie's vac pack to a description_info section as well as adding information about how to use the swoopie like a normal vaccum
add: also gave swoopies a muffling belly by default
del: Removed contaminates from swoopie's neck bellies. Their belly is the messy place
qol: made swoopies' integrated vac-pack only trigger while they're on grab intent
balance: rebalanced something
qol: made swoopie's neck bellies not have entrance logs by default. This should reduce message spam when sucking down stuff.
fix: Fixed swoopies being able to suck while dead
fix: Fixed swoopies AI returning to the mob and being able to vaccum the ground while the player was dead and ghosted (also presumably fixing pure AI swoopies from doing this too)
code_imp: changed swoopies from using speedy_mob_processing to using TURBO MODE belly flags. It does the same thing, but TURBO MODE is ingame accessable without varedit, so I think that's the proper implementation that should be used going forward
/:cl:

### DIFF
--- a/modular_chomp/code/modules/mob/living/simple_mob/subtypes/vore/swoopie.dm
+++ b/modular_chomp/code/modules/mob/living/simple_mob/subtypes/vore/swoopie.dm
@@ -1,6 +1,7 @@
 /mob/living/simple_mob/vore/aggressive/corrupthound/swoopie
 	name = "SWOOPIE XL"
-	desc = "A large birdlike robot with thick assets, plump belly, and a long elastic vacuum hose of a neck. Somehow still a cleanbot, even if just for its duties. Use DISARM intent to access Vac-Pack settings."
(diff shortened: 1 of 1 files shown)


ReoDaProtovali:
the branch name is funny

PR #7151, opened 2023-10-30
Reverts a bit of #6587

tacoguy7765093:
## About The Pull Request

The PR that makes oxygen tanks anchored doesn't even work properly, it doesn't prevent the lockers from flying when pressure gets changed. It served more as a borg/body dragging nerf and anti QoL more than anything. I am leaving the secure closets anchored though.

## Changelog

:cl:
del: Removed those rusty bolts off oxygen lockers.
/:cl:


Skysteak:
Given that this didn't even stop lockers from flying off (or the locker being yanked open by pressure, causing their contents to fly out), this is a welcome change.

CMDRYouth:
Are you sure you have this information right?  Bolted Lockers stay bolted and in place on the floor during a pressure change, and they do not open automagically, that's not a thing.  Every open locker/crate was opened and left open by a player.  Also by "oxygen tanks", you mean Emergency Lockers, don't you?

PR #7171, opened 2023-11-03
Fixes resize server hitches

BlackMajor:
Disables aura_animation on resizes.
We actually had this disabled a while back but it's been re-enabled for some reason.

If you look extremely closely, it animates a faint, blue-ish hue around the sprite of someone being resized for the 0.25 seconds the animation plays.

And it chugs the server, not worth it. It gets significantly more expensive when that mob has multiple overlays that all inherit the animation.


BlackMajor:
Self merge sin

PR #7175, opened 2023-11-05
A fix, and a major change - Trait addition

FluffMedic:
Hello hello, I return. It's 2 AM.

## About The Pull Request

Hello hello, I return. It's 2 AM.
I am here with two probably agreeable things, and one that is...certain to spark debate. Which is why I am going to sleep when I hit create pull request.

The Fix
-Lowered the vaule of deep breather. Is useable normally now.

Nerf is being removed for another discussion/PR

NOW, the biggest change.
There is a funny variable in the trait section.

varchange_type
And this can override the base values of what the trait is being applied to.
Soo, I set haste, slowdown, weakling, brute and burn to being non custom and yeeted this on it.
ICly, I think haste, slowdown, and weakling can be explained via really in or out of shape.
Brute and burn....yeah I don't got much outside of gene modding, or some intense physical training. But I just want to give the folks the chance to make this characters.
And hamper custom species being the main choice.

## Changelog

:cl:
add: More traits to use for non-customs
fix: Deep breather is more useable
/:cl:


Aroliacue:
I don't think we should be nerfing the damage resistance traits without some prior discussion both internally and community-wide to decide where we stand on balance. 

These values have been as they are for a very long time now, suddenly coming out and nerfing them without at least addressing some of the larger issues present to reduce the burn would be ideal.

The other changes are great, I see no issue with them.

PR #7194, opened 2023-11-09
Altering Outsider Pods

FluffMedic:
-Made Demon Pool v2 less painful to use via changing doors. -Removed the mechas on the fab ship.
-Removed 2/3s of the mechas for mech storage fab
-Nerfed the trading ship by removing the extra pods, and the charged slime core

I want to make some more outsider pods, but unsure where area code is.

## Changelog

:cl:
remap: tweaked some outsider pods
/:cl:


S0LiD-ST4T3:
i do not appreciate removing the extra pods, but the slime cores have to go.

FluffMedic:
Outsiders should not begin with things stronger then station start, much less the ability to make a small town.

Kazzc:
This is pretty good, thanks for looking into this.

PR #7230, opened 2023-11-17
More Xenomorph Markings

Aroliacue:
:cl:
add: Added in two new marking specific Xenomorph Queen heads and a new Xenomorph backplate.
/:cl:


Aroliacue:
image
Image for reference (100% fully colourable, will do another set of markings for the teeth and maybe a glossy body another time as different heads have different teeth positions)

PR #7285, opened 2023-11-29
Event stuff, but also Squeak component

BlackMajor:
Ok this is good to merge.

I added a bunch of things that are mostly relevant to events, but also the squeak component, which could very much be refactored into all instances of mob step sounds.

Borgs could use it, if they wanted move sounds, for example.

More *explicitly*, however, what was tweaked was..
-Event preferences now include Be Prey(Endo/Digestion/Absorption) and Be Pred(Endo/Digestion/Absorption)
-pick_weight added, for picking items from a list with a weighting.
-Removed duplicate COMSIG_MOVABLE_MOVED in atom/Move()
-step_trigger/message had its span removed, so as to not conflict with mapped in spans
-step_trigger/teleport_fancy fixed
-squeak component ported from tg, with an adjustment to act as a footstep controller for mechs and mobs
-mech_shoes tweaked to use squeak component instead of hard coded footsteps
-added a headset/event radio that applies an overlay, spells, light, speed, etc. and removes them when removed.
-check_shields now also checks ears
-slight tweak to footstep sound math
-autostripping map effect that removes a target's entire equipment list/genetics powers and stores them in a marked locker before re-equipping them with some basic gear
-pen item added that just directly applies a stun
-unrestricted spell datum type added, for unrestricted spells

🆑
qol: Expanded event preference selection to include specific vore options
/🆑


BlackMajor:
Self merge sin cause I want to get other PR's into the TM phase.

PR #7303, opened 2023-12-02
[MIRROR] PDA Ringtone Pref

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/15550
--------------------
Adds a persistent PDA ringtone pref to char setup, along with a selection of new custom PDA ringtones. Removed one disabled ringtone and another ***obnoxiously*** loud one.

Almost accidentally included some unrelated body/surgery changes I was testing, but I've shelved those for now!


ReoDaProtovali:
Removes the dust PDA sound- probably should remove that part of the PR when this gets merged

PR #7324, opened 2023-12-03
Adjustment to "Brings back pref ignoring micro stomps"

Morsel-LP:
## About The Pull Request

Currently, if a player is 75% or below in size, they have 'step mechanics' disabled, are in harm/grab/disarm intent, and are bumped into by another player in harm/grab/disarm intent, they will be knocked to the ground and briefly inflicted with weaken.

What this basically means is that a character below 75% cannot turn off step mechanics, regardless of prefs. All they can do is disable the fetishy description of the act.

This is also inconsistent with the 'micro' debuffs that I'm aware of, which all have a cutoff of 50%.

As such, I've lowered the threshold for this feature to activate from 75% to 50%.

## Changelog

qol: Reduced the threshold where other players get knocked down to 50% size.

### DIFF
--- a/code/modules/vore/resizing/resize_vr.dm
+++ b/code/modules/vore/resizing/resize_vr.dm
@@ -319,7 +319,7 @@
 	now_pushing = 0
 	forceMove(tmob.loc)
 	if(a_intent != I_HELP)
-		if(tmob.size_multiplier > 0.75 && nofetish) //So we can stun micros with step mechanics off, but prevent macros from stunning regular heights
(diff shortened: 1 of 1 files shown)


Raeschen:
I don't really agree. The aim difficulty on a 75% sized standard humanoid silhouette is already difficult enough to warrant some kind of balance payoff.

Raeschen:
for reference:

image

100% - 75% - 50%

Raeschen:
Denied after internal review

PR #7348, opened 2023-12-06
Undo part of #7336

ReoDaProtovali:
## About The Pull Request

Black gloves have a purpose. This PR gives them that purpose back. The temperature resistance of black gloves is because of their ability to let you remove active lights, which otherwise should (but don't because of hugbox) burn you. Just a simple revert of a decision that probably didnt think of the reason black gloves specifically mentioned being different opposed to normal gloves.

## Changelog

:cl:
balance: removed all gloves being temperature resistant. Back to the ones that were specifically defined to be.
/:cl:


ItsSelis:
Has been resolved in #7423

PR #7350, opened 2023-12-07
Exploration Robot Rework

FluffMedic:
Alright. Lets do this.

## About The Pull Request

Okay, so changes.
Exploration Borgs
-> Lost the phase carbine.
What have they gained?
-> Pounce is normal part of their kit.
-> Four items custom made for them.

explotailspear
Force: 25 (25 damage)
AP: 25 (Most mobs that are armored tend to have 50+ melee armor. So both of this hounds weapons have 25 to try and make that less painful but not fully redundant)
Reach: 3 A medium ranged attack. Outside the flare gun, this is the longest form of attacking. Note, if you are a madman to due flare gun only, you have my respect but you are crazy.
projectileparry: 15 As a treat, also getting hit is risky as borg.

EMAG ITEM
explojaws
Force: 40 (Much higher damage due the high risks of melee)
AP: 25
projectileparry: 15

"expirmental shield projector"
Health: 35
Size: 7
2 tiles away from the user, 1 singular dirrection
So, this thing is very fragile. Typically it's going to block 2 or 3 attacks. This is meant to protect the overall group during the start of a fire fight, although there are applications if the borg is alone. The spear can hit things outside the shield.

smallmedi gun
-4 tile range
-Heals worse then an omni-nerd cell. 0.75 burn, brute, and tox healing
-300 charge per shot.

What is the purpose of this?
To try and bring the exploration module into it's own niche.
I am aiming for a more immediate combat support, whilst it contributes in a more melee/short range combat lacking any range options.
Borgmelee Weapons: One rewards very close melee combat, whilst the other tries to not strand the borg if a tad bit of range is required.
Shield Generator: It's large, able to cover multiple folks. Protection would really only be brief, typically the start of a fight, or flickering back in every now and then.
Medi Gun: You can heal folks near you, but the healing is minimal and does nothing for internal wounds, or oxy damage.

## Changelog

:cl:
del: Phase Carbine from explohound
qol: Pounce from emag to default
balance: explojaws, 40 damage, 25 armour pen, 15 projectile parry on emag (Might be changed to claws in a further update)
add: explotailspear, 25 damage, 25 armour pen, 15 projectile parry, 3 fields range
add: expirmental shield projector, 35 shield health, 7 fields long, 2 tiles away and only protects one direction. Follows along, but direction remains static
add: smallmedigun: 4 tile range, heals 3.75 burn and brute per use, 4 uses per full 2400 charge. Consumes 600 power per use
add: self-repair-system: Damaged armour can be repaired while stationary, heals 2.5 burn and brute on plating every 2.5 seconds while stationary, costs 50 power each tick. This does not apply to internal modules or destroyed plating.
/:cl:


Aroliacue:
Keep in mind that the rollerbed was there as there's no way for borgs to safely drag people without using vore mechanics. That's pretty important for exploration!

FluffMedic:
Fair.
I honestly thought it was for more medical purposes.
Proabaly going re add that in a moment

Kashargul:
To be tested.
The modules which are mostly copied can be sub pathed. (jaws/big/explo ; combat_borgblade/speer)
In the sleeper, we don't need the following line
`stabilizer = TRUE`
That's been an oversight during my borg fixes. Sleeper has this set to TRUE by default

Kashargul:
Requires some rework upstream as the original modules were never defined as melee.

FluffMedic:
kash has helped with making things look nicer organization wise, and suggested and added an armor repair thing for explo borgs.
Will need to readd the medigun changes, and there  is...to be changes done to the jaws(soon to be claws?)

Echopsychotic:
Some things needing to be fixed, Borg belly cannot consume (I tried on living and dead mobs)
Explorer comms seems to not function correctly on the borg (Not sure if it was client side problem or other problems.)

Will post here otherwise

Kashargul:
Will take a look later today if there's any issue with either.

Also you talk about the sleeper? RobotNom is independent from the module.

Kashargul:
Radio channel "Exploration" is indeed not connecting. I will look into that.

PR #7363, opened 2023-12-08
[MIRROR] Vorepanel Escape/Struggle customization

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/15559
--------------------
-> Adds customization to struggle, escape, transfer and the interaction chance vorebelly messages
-> Allows to adjust the absorb escape chance
-> Experimental VChat filter for voregan messages 

For the latter, if there's any further messages which should be added to it, just list them.

I've tested if the options can be changed and stored, but I couldn't test if all of them are displayed.
I've tested the filter as far as I could, but there might be some messages that slipped.

**!Test before Merging
!We are touching save files, create a recent backup of all player saves**


Kashargul:
Needs to be added to the chomp_modular panel.

Kashargul:
Files are resolved with patch 2 to match the upstream changes. We will need to pull the edited ones out and add them to the upstream PR containing the TG-chat Filter.

PR #7382, opened 2023-12-09
[MIRROR] Follow up fix for #15559

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/15582
--------------------
fix: the new message options can now also be reset to default by resetting the belly


Kashargul:
close, will be included in #7363

PR #7415, opened 2023-12-13
self recolour for simple mob / robot

Kashargul:
## About The Pull Request
It's finally done... not the best implementation, but due to the differences between the object handling in the machine compared to a self recolour, it's mostly duplications for now and no fully datum migration. So this is a stripped down version of BM's matrix colour code for self use.

-> Simple Mobs can use it once per round
-> Robots can use a module reset board to use it again

## Changelog

:cl:
add: self recolour verb for simple mobs and robots
/:cl:


FluffMedic:
I understand simple mobs.
But department based colors feels like a unique part of robots identity, but that may just be me.

Kashargul:
Well, same reason as of why crew can recolour their gear. Some people like to have the ability to customize some. Right now, those requests often need to be admin handled. This will reduce the admin handling and even though available. I highly doubt a lot of borg players will constantly use it.

PR #7478, opened 2023-12-25
Makes all OM shuttles breakable and docked plus misc fixes

tacoguy7765093:
## About The Pull Request

All of the overmap tiles have gotten changed into the 'hull' material the other server use for shuttles, as the shuttle walls we already have are unbreakable. The sprites are different but Kassc was okay with it. This also places simple docking controllers in the Carrier so the shuttle airlocks open properly.

This also fixes some power issues in some of the space POIs, notably the ones with turrets.

dreamseeker_Tm2FN0X60V
dreamseeker_T7vzrXdRaw
dreamseeker_xxmyu4iViN
dreamseeker_HCiCZ5h2Nd

## Changelog

:cl:
remap: The shuttle docking systems in the carrier docks are now functional.
/:cl:

### DIFF
--- a/maps/southern_cross/southern_cross-7.dmm
+++ b/maps/southern_cross/southern_cross-7.dmm
@@ -128,14 +128,15 @@
 	base_area = /area/expoutpost/hangarfive;
 	base_turf = /turf/simulated/floor/reinforced;
 	landmark_tag = "needle_dock";
-	name = "Needle Dock"
+	name = "Needle Dock";
+	docking_controller = "needle_dock"
 	},
 /obj/structure/cable/blue{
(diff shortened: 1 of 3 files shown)


Kazzc:
Not a fan of these walls but at least the shuttles are now destructible. Maybe in the future we can get the old style walls functional.

tacoguy7765093:
I have no idea with the best implementation will be for it, because it's multiple colors of walls and I would have to make a brand new subtype, Don't know what the best way would be without excessive snowflake.

Darlantanis:
do I need to make a version of these that use the unbreakable shuttle sprites

PR #7482, opened 2023-12-29
Replaces shadekin's dark vendors with non-event versions

tacoguy7765093:
## About The Pull Request

The vendors in the dark were accidentally the event versions, which enabled shadekin to have alien tools, like 50 e-swords, and all the translocators to cheese with. This replaces them with the same vendors you'd find on the station. 
Access locks on some vendors have been lifted.

## Changelog

:cl:
remap: - Nerfs shadekin retreat vending machines.
/:cl:


Kazzc:
Oof, that's a big oof on my part. Nice catch!

PR #7499, opened 2024-01-01
Small fix to the Deck 1

TheCaramelion:
## About The Pull Request

Whoopsie, small mistakes were made. Adding the shutters, plates and missing decal. And the autolathe that I accidentally removed.


TheCaramelion:
Lesson learned, take in-game screenshots for people to point out wrong stuff before actually comitting instead of just SDMM.

PR #7506, opened 2024-01-04
Valley Anniversary Update

FluffMedic:
-Removes the valley

## About The Pull Request

Sooo performance issues because of z levels is a thing.
Some having strange purposes.
There has been some minor discussion, hoping this pushes it further.

And there has been bits talking about the valley and it's strange existence.
It being just another explo zone.
Fusing it with wilderness.
Only half a dozen folks go to it, and it always feels like it gets rushed by those folks.

I dislike the concept of just shoving it into wilderness, so not going to do that as of yet. Will likely adapt some POIs to it eventually, others, especially a fair few valley cavern POIs less so.
And if anyone is going to kill this, it may as well be me. This was, no still is one of my favorite things to make, and expand.
If concepts and such are liked, and folks would want to see it in main wilderness, sure, I can adjust things for it.
But I rather be the one to do it.

And the eclipse.
My strange little faction I was slowly trying to flesh out.
Probably my most controversial design, that design bit rather netuered.
They're probably going to stay removed, might put them in a gateway, but ehh.
Still buffed them slightly because the armor nerf really did go slightly too hard.
The buff is 10 AS to 15. So...maybe a pistol no longer hurts them.
I think it's safe to say average damage is round 20? And the normal tankness is about 50, so even with 5 damage on some weapons, that's 10 shots.
Although unsure this reasoning matters if I am snapping it.

## Changelog

:cl:
balance: buffs the Eclipse
/:cl:


Kazzc:
I argue heavily against this, having more areas to explore is the whole point of exploration. Valley is a very good addition with a lot of content and options, albeit it is heavily focus on the use of mechanics that's the goal of exploration content. An option for players to wander out and get into an adventure whenever vore does not pan out.

If performance is the issue, we should cut content from areas that has no traffic. This being Sky islands and NortherStar mining, I think we have an other empty z-level with no content.

Aroliacue:
I think this is a good opportunity to go over the Valley and most POIs as a team and figure out not only the best way to make things performance-friendly, but also to go over balance and expectactions for what should and shouldn't be in a POI to begin with. I think the Valley is a fun concept but there's been quite a few issues that temporarily removing it for a rework may just be the best idea.

The valley has always just been considered a 'harder explo' area. I know plenty of people like the Eclipse faction conceptually, so maybe keep that in mind while working on another area that may have more lore and, importantly, more relevance with Sif itself.

Kashargul:
Considering the performance issues, we really should look into a lot of the combat runtimes. During testing the exploration hound, I could watch the runtime counter skyrocket during heavier combat.

User-U-U:
Please vaporize it, less Z-level bloat.

Aroliacue:
> Please vaporize it, less Z-level bloat.

No. We've fixed a lot of our performance concerns thanks to the dev team, so bloat really isn't a concern. It just needs to be updated.

PR #7510, opened 2024-01-04
Map Bugfix

Missile597:
## About The Pull Request

Fixes some map related bugs

## Changelog

:cl:
fix: Fixed the Large Alien Ship shelter to be spawnable
fix: Fixed the inventories of the Garden and Xenobotany seed storages to have the correct list of seeds at round start (from seed_storage_vr.dm)
remap: Added a missing air alarm to Chemistry Distillery
/:cl:


Darlantanis:
Tweaking starting seeds via mapedit is bad practice. Missing seeds should be added in the appropriate .dm file with // CHOMPedit comments as necessary.

Missile597:
> Tweaking starting seeds via mapedit is bad practice. Missing seeds should be added in the appropriate .dm file with // CHOMPedit comments as necessary.

Hi Nadyr, the seeds list is actually already in the .dm files. There's two .dm files and the seed storages were pulling from the shorter list without the additional seeds. The storages were pulling from [CHOMPStation2](https://github.com/CHOMPStation2/CHOMPStation2/tree/master)/[code](https://github.com/CHOMPStation2/CHOMPStation2/tree/master/code)/[modules](https://github.com/CHOMPStation2/CHOMPStation2/tree/master/code/modules)/[hydroponics](https://github.com/CHOMPStation2/CHOMPStation2/tree/master/code/modules/hydroponics)
/seed_storage.dm instead of [CHOMPStation2](https://github.com/CHOMPStation2/CHOMPStation2/tree/master)/[code](https://github.com/CHOMPStation2/CHOMPStation2/tree/master/code)/[modules](https://github.com/CHOMPStation2/CHOMPStation2/tree/master/code/modules)/[hydroponics](https://github.com/CHOMPStation2/CHOMPStation2/tree/master/code/modules/hydroponics)
/seed_storage_vr.dm

I actually couldn't find the correct seed storage referred to in the seed_storage_vr.dm on SDMM, so the only solution I could think of was to directly alter the list to match the list on the .dm file. I'm actually not sure why the storages started pulling their inventories from a different .dm file in the first place. Is there another solution you could recommend that might fix this outside of just mapping it? I could take a look but I don't know of any other way (still picking this up as I go)

Missile597:
> Tweaking starting seeds via mapedit is bad practice. Missing seeds should be added in the appropriate .dm file with // CHOMPedit comments as necessary.

Hi again Nadyr,

A very kind bean helped be find the .dm file causing the issue. This should be resolved outside of the map files and I've reverted the seed storages on the map back to the default prefab (which has been pulling from a .dm file in the modular files, turns out, would explain why the inventory changed). Let me know if there are any other issues that need fixing.

PR #7512, opened 2024-01-05
Minor Fixes - The Dark

Missile597:
## About The Pull Request

Some minor map tweaks in The Dark. One minor QoL addition and fixing the broken section of the disposals.

## Changelog

:cl:
add: Added a recharger station for QoL to allow shadekin to recharge cells at the hideout
remap: Added a missing disposal pipe segment to complete the loop.
/:cl:


Kazzc:
Not having a recharger on the hideout was intentional, I was mostly worried about it cause the hideout has infinite power. For convenience sake having one might be for the best.

Nice catch on the disposal pipe, surprised I missed that one.

PR #7513, opened 2024-01-06
Disposals Fix - Carrier

Missile597:
## About The Pull Request

Fixes the various issues with the disposals loop on the exploration carrier. This should allow the disposals bins to all function correctly and send trash to the disposals section of the carrier (instead of ejecting it from other bins or through the floor).

## Changelog

:cl:
del: Removed some excessive pipework causing redundant loops and potential "stuck" trash
fix: Fixed some disposals junction's orientations so that the trash flows correctly towards the disposals sorting room
fix: Connected a couple of disposals bins to the loop that were previously not connected
remap: Added a disposals bin in the gateway prep room, and connected it to the dead end pipework that was heading in that direction
/:cl:

### DIFF
--- a/maps/southern_cross/southern_cross-7.dmm
+++ b/maps/southern_cross/southern_cross-7.dmm
@@ -443,6 +443,10 @@
 /obj/machinery/atmospherics/pipe/manifold/hidden/scrubbers{
 	dir = 1
 	},
+/obj/structure/disposalpipe/segment{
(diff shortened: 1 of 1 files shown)


Aroliacue:
Looks good to me! Thanks for sorting this out (no pun intended)

Kazzc:
Nice catch, thanks for the fix.

PR #7533, opened 2024-01-10
fix runtime, increase sm explosion damage

Kashargul:
Increasing the damage of a delam. The maximum range now reaches up to the tech storage. Minimum range is also less forgiving, messing up the engine room.

We really shouldn't call procs on objects we have QDEL_NULL deleted. Therefore, we now first reset the alarms and prevent all changes on power change on nulled objects.

## Changelog

:cl:
balance: increased SM delam explosion range by 50 % (MIN 8 -> 12. MAX 16 -> 24)
fix: fixed a runtime in the alarm system.
/:cl:

### DIFF
--- a/code/game/machinery/air_alarm.dm
+++ b/code/game/machinery/air_alarm.dm
@@ -845,6 +845,8 @@
 	spawn(rand(0,15))
 		update_icon()
 		// CHOMPEdit Start: Looping Alarms
+		if(!soundloop)
+			return
 		if(stat & (NOPOWER | BROKEN))
 			soundloop.stop()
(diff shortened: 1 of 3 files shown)


Kazzc:
I love this, but this may tank server performance, even more during events. Wonder if I can crash an event by busting an SM at mid round.

Kashargul:
Sounds good to test it first. I think our biggest threat to crash the server during events are still a lot of unhandled combat / destruction runtimes...

Darlantanis:
I miss the old SM explosion.

Not seeing a need to TM. This should still be below the explosion cap.

PR #7537, opened 2024-01-12
Lost and Found Remap

Missile597:
## About The Pull Request

Remaps the lost and found system, hopefully adding some QoL changes and some more lost and found bins in public bathrooms for easier and more discreet return of prey's items. The new lost and found bins can be found in the public bathroom areas near common scene locations (bar, gym, dorms and near the qpad - for those returning from Sif).

## Changelog

:cl:
add: New lost and found bins added to dorms bathrooms, the bathroom near the qpad, the bathroom near the bar and the bathroom near the gym, to allow for easier and more discreet item returns. These connect to the lost and found loop and also clean prey items.
del: Removed the junk on the shelves in Autoresleeving so you don't have to go digging through random junk to get your stuff. Removed one shelf to make room for chute.
remap: Moved the lost and found exit out of the wall, so any players summarily "tossed" into a bin don't get wall-stuck. Aimed the chute at a wall so that there's no risk of any window breaking if hit excessively.
/:cl:

### DIFF
--- a/maps/southern_cross/southern_cross-1.dmm
+++ b/maps/southern_cross/southern_cross-1.dmm
@@ -32,6 +32,7 @@
 	icon_state = "1-2"
 	},
 /obj/effect/landmark/event_spawn/morphspawn,
+/obj/structure/disposalpipe/segment,
 /turf/simulated/floor/plating,
 /area/maintenance/firstdeck/foreport)
 "aae" = (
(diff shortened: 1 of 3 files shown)


Missile597:
I want to screeeeeam! For some reason github didn't give me the up to date map files and previous PR's have been undone. I will... try and re-do this I guess ;w;

PR #7551, opened 2024-01-14
Metamorphosis Ray design for RnD

Fracshun:
## About The Pull Request
A fun scene tool now more available, similar to how RnD can print others like the Body Snatcher. Tech and Material costs based off the Recombobulation Ray design which is effectively the same gun but only reverts. Added requirement for Illegal level 5 and placed under the energy weapon category.

## Changelog

:cl:
add: Added the Metamorphosis Ray to science's protolathe designs
/:cl:


Kashargul:
Could be put directly into chomp_modular

Fracshun:
Not only could it be moved there, but it seems the confetti cannon in there was using the same sort string. So moved and fixed.

PR #7568, opened 2024-01-17
Cursed Traits

FluffMedic:
Hello, here are traits that folks probably didn't think of, nor wanted, but I desired them. Oh and there are a few folks did want but those aren't important.

## About The Pull Request

Alright, new traits. First positives.
For 1 point, you can get Virus immune. The effect is in the name. No, this trait is not worth more then 1 point currently seeing as the only realistic way of getting a virus on the average round is a random event that may or may not proc.

Then for 3 points, EMP resist. Synths can take 30% less EMP damage with this trait.
And there is the 5 cost upgraded version, of 50% less EMP damage.

Now, negatives.
For -4, thick digits. You cannot use guns.

For -3, you can get faulty wiring. You take 30% MORE Emp damage with this trait.
For --5, you get fault construction, taking 50% more EMP damage.

Note, any trait that mentions EMP is exclusive to one another.

For -2, you can obtain the no defib flag.

For -1, you can obtain the no scan flag, which is no resleeving. Note, digestion prey that enjoy coming back, do not take this trait.

For -1, you can obtain the water weakness trait. Congrats, you are now a promethean, and melt in water.

For -3, you can obtain the upgraded water weakness trait, which is nearly triple water damage. Bring an umbrella.

## Changelog

:cl:
add: More traits!
/:cl:


ReoDaProtovali:
can I suggest:
A trait for prommies that's like the opposite of the water vulnerbility trait that makes them instead safe in water
making the EMP traits also give radiation defense/immunity and reference circuit hardening (which is a real life thing used to make electronics durable against electromagnetic interference

Echopsychotic:
Wonder if it should be that shadekin should get the thick fingers trait as a species trait...

FluffMedic:
For the suggestions.
I don't really like the promethean one, since I feel like it's apart of the species identity.
The rad resist on the emp trait is nifty but I rather it not conflict with the existing rad traits traits.

And for the shadekin bit, there has been chat about that. Well sort of. They're not meant to be able to use advanced tools, and their is actually a flag for that. And I think guns are apart of that.

I think it could be a good way to balance kin but it be controversial as hell.

TheCaramelion:
In my opinion I think the no resleeve/defib should be 0 points, otherwise crew that doesn't put themselves in danger or simply hang out on the station will just grab them because free points.

theXDkidoflol:
> \>5 points if you just resleeve instead of defib (Most people do this anyways)
> \>Only costs 1 point to remove all viruses 
4ADC30FB-A963-4090-A855-23EF807081C2

I know that explo is the only department that receives updates, but does that really mean we should remove the only elements of gameplay that medical gets? That’s just my two cents, but this PR WILL kill medical as a whole if passed through in its current state.

FluffMedic:
> > For -3, you can obtain the no scan flag, which is no resleeving. Note, digestion prey that enjoy coming back, do not take this trait.
> 
> I have mixed feelings there.
> 
> I am someone who loves not coming back. However, giving me a bonus of 3 point is excessive as it provides no mechanical advantage since players who weren't coming back were doing that anyway. It's free points, and too many free points for that matter. Nothing prevents me from rejoining the round anyway by just using respawn. This only bars me from being resleeved and presumably the autosleever.
> 
> That being said, the powerful -3 may be justified if an explorer for example screws up and _dies_ cold and alone and their body will never be found and now they cannot rejoin the round. But what is to prevent them just respawning? I do see potential of this to discourage lonewolf gameplay, which is good.

There is actually a thing that stops you. Staff members, and rules.
As far as I am aware, respawning as such, and getting noticed is a descent way to get bapped.
At least currently, Xenochimrea, outsiders, and other folks without access to a resleever it seems that currently, headmins are against them respawning even if they died of vore.

Although I can see that I may have been to excessive on the point values.

FluffMedic:
> I know that explo is the only department that receives updates, but does that really mean we should remove the only elements of gameplay that medical gets? That’s just my two cents, but this PR WILL kill medical as a whole if passed through in its current state.

How will it kill medical?

FluffMedic:
> In my opinion I think the no resleeve/defib should be 0 points, otherwise crew that doesn't put themselves in danger or simply hang out on the station will just grab them because free points.
> 
> For example I will absolutely pick up the no defib one as my xenochimera because I'd love to use the mechanics of my race instead of having Medical defib me and that's it. I would love to reconstitute, and have Medical force the process faster instead.

I can see the argument for zero points, but for folks who go into danger and such will have an actual risk.
But yeah was proabaly overkill for the points last night.

SpadesNeil:
> > > For -3, you can obtain the no scan flag, which is no resleeving. Note, digestion prey that enjoy coming back, do not take this trait.
> > 
> > 
> > I have mixed feelings there.
> > I am someone who loves not coming back. However, giving me a bonus of 3 point is excessive as it provides no mechanical advantage since players who weren't coming back were doing that anyway. It's free points, and too many free points for that matter. Nothing prevents me from rejoining the round anyway by just using respawn. This only bars me from being resleeved and presumably the autosleever.
> > That being said, the powerful -3 may be justified if an explorer for example screws up and _dies_ cold and alone and their body will never be found and now they cannot rejoin the round. But what is to prevent them just respawning? I do see potential of this to discourage lonewolf gameplay, which is good.
> 
> There is actually a thing that stops you. Staff members, and rules. As far as I am aware, respawning as such, and getting noticed is a descent way to get bapped. At least currently, Xenochimrea, outsiders, and other folks without access to a resleever it seems that currently, headmins are against them respawning even if they died of vore.
> 
> Although I can see that I may have been to excessive on the point values.

As I have said a million times before, relying on staff to correct easily exploitable game mechanics is lazy, stupid, and unreliable.

This rule will be unenforceable. Instead just code it to make sure the player can't just abuse respawn to circumvent the free points. _Or_ take away the free points. As a player who will use this, I'm not using it for the free points.

modular_chomp/code/modules/mob/living/carbon/human/species/station/traits/negative.dm
```diff
@@ -53,7 +53,7 @@
 /datum/trait/negative/noresleeve
 	name = "Unsleeveable"
 	desc = "Your genetics have been ruined, to the point where resleeving can no longer bring you back, including the autoresleever."
-	cost = -3
+	cost = -1
```
SpadesNeil, line 64: This is far more reasonable.

modular_chomp/code/modules/mob/living/carbon/human/species/station/traits/negative.dm
```diff
@@ -44,7 +44,7 @@
 /datum/trait/negative/nodefib
 	name = "Unreviveable"
 	desc = "For whatever strange genetic reason, defibs cannot restart your heart."
-	cost = -5
+	cost = -2
```
SpadesNeil: I think this is fair. Now people might be thinking, "Okay but they can just resleeve."

Yes, but you can't be brought back in the field. Your ass has to be dragged all the way back to medical if you forgot your scan, or, you at the very least upon being resleeved have to go _all the way back_ to where you left off. It's still punishing to get killed.

Without this trait, if you get killed in the field, you can still get some hasty medical aid and get defibbed without going home.

SpadesNeil (approved):
My problems with this have been resolved.

PR #7569, opened 2024-01-17
Item TF device: Mind Binder

Fracshun:
## About The Pull Request
Adds a new illegal hack of the SleeveMate which enables binding minds to any item. Does not have all the functionality of a SleeveMate. It simply stores a mind within itself while leaving the body behind, and if it contains a mind it can place it within any item or back into a mob. Also has a mode to bind your own mind, along with plenty of warnings of how bad of an idea that might be unless you have help around you. Can be obtained through trash, contraband, the protolathe, and by emag'ing a SleeveMate.

Misc changes along with this is allowing those who are turned into objects, and being digested inside a belly with recycling, to transfer into becoming the digested material object which builds up with recycling enabled. Also found that the brain organ would teleport carbon's back into it on gib even if they no longer inhabited their body. So fixed that, and now you can turn your friends into items and dispose of the evidence without them teleporting back into their brain.

## Changelog

:cl:
add: Added new illegal item for handheld Item TF: Mind Binder
add: Sentient items digested in a belly with recycling now become the recycled matter
/:cl:

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -269,7 +269,7 @@
 			ghost.assumeform(src)
 			ghost.animate_towards(user)
 	//VORESTATION EDIT START. This handles possessed items.
-	if(src.possessed_voice && src.possessed_voice.len && !(user.ckey in warned_of_possession)) //Is this item possessed?
+	if(src.possessed_voice && src.possessed_voice.len > 1 && !(user.ckey in warned_of_possession)) //Is this item possessed?
 		warned_of_possession |= user.ckey
 		tgui_alert_async(user,{"
 		THIS ITEM IS POSSESSED BY A PLAYER CURRENTLY IN THE ROUND. This could be by anomalous means or otherwise.
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -269,7 +269,7 @@
 			ghost.assumeform(src)
 			ghost.animate_towards(user)
 	//VORESTATION EDIT START. This handles possessed items.
-	if(src.possessed_voice && src.possessed_voice.len > 1 && !(user.ckey in warned_of_possession)) //Is this item possessed?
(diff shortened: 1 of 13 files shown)


code/game/objects/items.dm
```diff
@@ -269,7 +269,7 @@
 			ghost.assumeform(src)
 			ghost.animate_towards(user)
 	//VORESTATION EDIT START. This handles possessed items.
-	if(src.possessed_voice && src.possessed_voice.len && !(user.ckey in warned_of_possession)) //Is this item possessed?
+	if(src.possessed_voice && src.possessed_voice.len > 1 && !(user.ckey in warned_of_possession)) //Is this item possessed?
```
Darlantanis, line 272: needs comment

code/modules/vore/eating/digest_act_vr.dm
```diff
@@ -76,8 +77,6 @@
 			var/obj/item/device/pda/P = src
 			if(P.id)
 				P.id = null
-		for(var/mob/living/voice/V in possessed_voice) // Delete voices.
```
Darlantanis, line 79: Better to comment these lines out with //CHOMPedit comments instead of deleting, otherwise it may cause confusion in a merge conflict

code/modules/vore/resizing/holder_vr.dm
```diff
@@ -48,7 +48,7 @@
 		dropInto(user.drop_location())
 		dropped(user)
 	//VORESTATION EDIT START. This handles possessed items.
-	if(src.possessed_voice && src.possessed_voice.len && !(user.ckey in warned_of_possession)) //Is this item possessed?
+	if(src.possessed_voice && src.possessed_voice.len > 1 && !(user.ckey in warned_of_possession)) //Is this item possessed?
```
Darlantanis, line 51: needs comment

Fracshun:
Added that moved portion back in where it was originally and commented it out, along with adding the other Chomp edit markings.

PR #7588, opened 2024-01-19
[MIRROR] Many bug fixes, QoL, and PDA notes port (now with sensible diffs)

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/15689
--------------------
Fixes many old bugs, nerfs a foam greatsword exploit, lets players OOC escape from microwaves, and ports an upgrade to the PDA notes app. Same as https://github.com/VOREStation/VOREStation/pull/15684 but with diffs fix; thank you to @ Kashargul for helping me fix this.

Changes

- Microwaves are now OOC escapable, just adds to the OOC if() checks.
- Adds "can_sharpen" bool to material datums with TRUE by default, sets it to FALSE for foam, and adds it to the sharpen() check. This prevents players from creating the foam durasteel greatsword, a HALLOS weapon which can knock out most species in one hit.
- Tweaks codexes to generate a single codex_tree per book type instead of a tree per book, so 300 books can use 1 massive datum instead of 300 identical massive datums which also patches a memory leak when codexes are deleted. Also fixes a bug with the "go back" button permanently breaking due to a bad math.
- Makes a "rocks" decl and assigns it to the outdoors/rocks turf, reintroducing correct footstep sounds.
- (Hopefully) fixes a bizarre edge case with tg/map_template where submaps which do not allow duplicates repeatedly spawns on top of itself by checking "allows_duplicates" in the while() loops
- Adds extra note sections to the PDA notes app, allows the notes to copy text from papers, and allows notes to print onto blank papers
- Fixes a chem ID for zombie powder so it appears to stop hearts again.
- Tweaks highlander mode contents deletion loop to no longer delete organs, which seems to have fixed a crash issue. I doubt anybody uses this, but having less server crash buttons is probably a good thing.
- Adds check so conveyor belts can no longer move obj/effect/abstract, so it can't steal your flashlight light anymore.
- Adds a missing flag to particle accelerator particles and spawns the particle on the device itself, so now it bumps and irradiates mobs in front of the particle accelerator again.
- Adds missing surgery names
- Tweaks photographs and dogborg sleepers to account for MaxHealth so species with lower max health no longer appear hurt while healthy.
- Adds loops to organ spawning to properly run meat initializing, making organs butcherable again. Also changes a forceMove which could leave organs stuck to player HUD's.

Everything has been tested locally. Big thanks to [@ Willburd](https://github.com/Willburd) for the bulk of this code.

Downstream changelog
🆑Upstream
qol: Microwaves are now OOC escapable
balance: Nerfs foam durasteel greatsword
fix: Fixes memory leak with codexes
fix: Fixes bug which bricks codex "go back" button
fix: Reimplements footstep sounds for rock turfs
fix: Fixes edge-case which sometimes repeatedly spawned POI's on top of each other
add: Adds more PDA notes, adds copy and print function to PDA note app
fix: Zombie powder makes hearts appear stopped again
fix: Highlander mode no longer crashes the server, and also doesn't delete your organs anymore
fix: Conveyor belts no longer move flashlight light cones
fix: Particle accelerator properly applies radiation to bumped mobs again
add: Adds missing surgery names
fix: Healthy tesh no longer appear hurt on photographs and dogborg sleepers
fix: Organs no longer get stuck to the HUD
fix: Organ butchering works again
/:cl:


ItsSelis (changes requested):
changes in pda_notekeeper.jsx have to be also copied over to the same file in the tgui_ch package.

Kashargul:
tgui bundle conflict with the vorepanel update.

PR #7593, opened 2024-01-19
Solar grubs can now evolve

TheCaramelion:
## About The Pull Request
**It was not the solar moth in the valley after all.**
An evolving solargrub attempting to expel prey from an empty belly would throw an exception and wouldn't allow it to evolve, leaving them as solar grubs forever.

Additionally, killing a solar moth would not bring the solar moth count down, which only allowed a singular moth to ever exist per shift.

## Changelog

:cl:
fix: Solar grubs can now evolve
remap: Solar moth re-added to the valley
/:cl:


Darlantanis:
why was that a list

PR #7597, opened 2024-01-20
Secret Resleever Fix

Missile597:
## About The Pull Request

Replaces the secret resleever with one that works, for all you preds that love to camp the resleever <3
Now you too can create your own medical gameplay

## Changelog

:cl:
remap: Replaced the secret console with a working resleever console
remap: Replaced the secret cloning pod with a sleeve growing pod
/:cl:

### DIFF
--- a/maps/southern_cross/southern_cross-3.dmm
+++ b/maps/southern_cross/southern_cross-3.dmm
@@ -12903,9 +12903,7 @@
 /turf/simulated/floor/tiled,
 /area/bridge)
 "lrr" = (
-/obj/machinery/clonepod,
-/obj/item/weapon/reagent_containers/glass/bottle/biomass,
-/obj/item/weapon/reagent_containers/glass/bottle/biomass,
+/obj/machinery/clonepod/transhuman/full,
 /turf/simulated/floor/plating,
 /area/maintenance/thirddeck/dormsstarboard)
 "lsm" = (
@@ -27275,7 +27273,7 @@
 	name = "Third Deck Aft Maintenance"
 	})
 "xnt" = (
-/obj/machinery/computer/transhuman{
+/obj/machinery/computer/transhuman/resleeving{
 	dir = 1
 	},
 /turf/simulated/floor/plating,


Kazzc:
Unsure about placing an actual resleever, when I originally mapped it I was against it since it was abused in the past, reason I placed a broken one.
But we now have autoresleevers. . . I wonder if we could place an autoresleever instead. So if players want to privately camp someone, its at their own consent.

Missile597:
> Unsure about placing an actual resleever, when I originally mapped it I was against it since it was abused in the past, reason I placed a broken one. But we now have autoresleevers. . . I wonder if we could place an autoresleever instead. So if players want to privately camp someone, its at their own consent.

Hi Kassc, sorry for taking a while to get back to you. This change was actually a fix to a resleever that previously got added to the area and merged before, which was not functioning due to a couple of issues with the wrong objects being placed. I did consider maybe placing the parts needed to build a new resleever machine there instead of plain fixing the resleever, but I added the resleever on request of a couple of other users, who just wanted the existing one to be working for scene use.

I would like to think it'd be interesting to add it and see if it gets abused or not (we can remove it if it does), but ultimately up to the moderation team if this wants to be done differently. A few have raised the idea that it'd be nice to have a resleever to use for a scene somewhere private without having to use the one in medical or the auto resleever that others will want to use. Might be worth discussing with the others and letting me know what they think, as I can always take another look at this if the staff have concerns or other better ideas :)

Kazzc:
I will still prefer to have a hidden autoresleever than a resleever, since this is for roleplay purposes the current resleever requires biomass, power and a console that has access to everyones body records and mind records. A player can easily sleeve bodies from other players and upload minds, I can see this becoming an issue sooner than later.

The best approach here is to have a hidden autoresleever instead, so any predator can camp it and eat the same or multiple prey that choose to spawn on that location.

Kazzc:
I will still prefer to have a hidden autoresleever than a resleever, since this is for roleplay purposes the current resleever requires biomass, power and a console that has access to everyones body records and mind records. A player can easily sleeve bodies from other players and upload minds, I can see this becoming an issue sooner than later.

The best approach here is to have a hidden autoresleever instead, so any predator can camp it and eat the same or multiple prey that choose to spawn on that location. Lets see if this gets abused, if it does then we find a better way to adjust this.

PR #7616, opened 2024-01-23
Exploration Equipment Vendors and misc changes

Aroliacue:
:cl:
add: Added new standalone exploration vendors for each of the roles with standardized equipment. Not mapped in as of this update, but available for gamemasters to use.
add: Added tinyfans to the 25x25 'Diner' shelter.
qol: Significantly reduced use delays on all pouches.
balance: Fultons per pack have been reduced from 5 uses to 3 uses.
balance: Mech shoes can now hold knives.
balance: Flares and glowsticks are now one size smaller. Flare boxes now contain 14 flares instead of 4.
balance: Flare pouches can now hold 14 flares.
balance: Removed guaranteed stun and weaken statuses from the 'magnetic' weapon parent type.
balance: Regular synthetics are now immune to welders.
fix: Fixed missing icon for mech shoes.
/:cl:


Kashargul:
Makkin, runtime!
[2024-01-23T14:52:30] Runtime in unsorted.dm,1536: /obj/item/stack/marker_beacon/thirty does not have a valid icon state,

Aroliacue:
Me when an item that has existed for some time causes runtimes when put in a vendor

PR #7618, opened 2024-01-24
Altering space disposal and middle maints access

FluffMedic:
Title

## Changelog

:cl:
maptweak: more folks can go into middle maints and the maints disposals
/:cl:


Kazzc:
No
Central access is locked since its a breaching point for Tcoms, Teleporter and HoP office.
Disposals is locked for cargo personnel only and whenever personnel tries to run away from security by jumping down disposals.

PR #7619, opened 2024-01-25
Upgrades Sif grass, fixes wingrilles

Darlantanis:
-Tweaks turf edges so normal Sif grass can grow onto dirt paths.
-Makes wingrilles spawners able to function after round start. Why were these disabled when specific code exists to trigger an inactivate spawner?
-Tweaks wingrille spawner layer for mappers.

:cl:
qol: Adjust turf edges priority
qol: Tweaks windgrille spawner layer
fix: Makes wingrille spawners work after roundstart
/:cl:


Aroliacue:
Great changes, much appreciated!

PR #7620, opened 2024-01-25
Tooling updates and removal of corrupt maps [MDB IGNORE]

Drulikar:
## About The Pull Request

This PR ports various changes to tooling for mapmerge2, dmimerge, and update paths from TG. The primary reason was to get the hook for map merging so conflicts are much more likely automatically handled by you simply merging master into your PR.

In the event that a map touches the exact same tile as what someone else also altered and the merge tool couldn't automatically resolve this, this now places a new object at that location that throws a runtime when it is initialized so it should be clear both in strongdmm where it is, and yell at you in game should you have forgotten about it.
conflict
strongdmm
runtime

Example of a DMI merge:
image

The hooks can now be installed in \tools\hooks as either install.bat (windows) or install.sh (linux). For more information see the readme in the hooks folder.

For UpdatePaths, there is a readme in that folder, but for example the paths I updated already are in the scripts subfolder as an example (and could be re-used if maps from upstream are ported). I did not exhaustively fix all the broken paths in maps, but I did have it do everything I saw upstream tweaking in the maps already for broken paths (namely sofas, ammo boxes, some clothing, and pointdefenses)

There are also pre-commit hooks that ensure commited map files can be decoded the same as they are encoded. If you encounter a traceback for a particular file you may want to investigate that file, consider maybe pulling the file from upstream, otherwise consider discarding it. I personally could not figure out what the heck upstream has been doing to their maps to break their map's keys, but it probably doesn't help that they've been leaving merge conflict markers everywhere.

Line endings for files are also now enforced the same as TG.

All merge conflict markers are now removed with this PR. Be careful with mirrors from upstream; they have conflict markers everywhere.

## Changelog

:cl: Drathek
del: Removed corrupted maps in virgo maps folder: groundbase, surface_submaps/pains, depreciated talon2, and cynosure
fix: Removed conflict markers in various maps and parts of code
fix: Fixed various minor warnings and errors; primarily bringing parity in Virgo maps folder closer to upstream
fix: Fixed most line ending inconsistencies
maptweak: Repathed some broken paths in maps (namely sofas, ammo boxes, some clothing, and pointdefenses)
/:cl:


Drulikar:
Unfortunately it looks like mapdiffbot is attempting to parse the maps I removed and of course it too dies when it encounters the merge conflict markers :P Hopefully that is fixed after this is merged though.

Regardless I'm happy to address any concerns with how broadly I had to alter files to get the tooling happy with parsing the codebase. There are likely some files that are not partially corrupted in the groundbase folder if that needs to be restored.

Drulikar:
Will see if aa07 can tweak mdb to work in a situation like this; but this pr in general probably touches too many maps for it.

Drulikar:
BTW you may want to hide whitespace changes when reviewing this PR; most of the changes are line ending corrections. 
image

Darlantanis:
Is it possible to correct line endings without github detecting changes to the entire file? That's the main thing holding up this PR right now as it breaks blame for all modified .dm files.

Drulikar:
> Is it possible to correct line endings without github detecting changes to the entire file? That's the main thing holding up this PR right now as it breaks blame for all modified .dm files.

Ultimately it is a change to the final byte(s) of the line so it is correct that it is a line change. But often diffs include means to ignore whitespace changes because it can make changes harder to distinguish.

I wouldn't worry about blames too much - all history will still exist for the file. Just yes if my PR is the last commit altering it then you would have to view a previous blame.

In VSC with gitlens extension installed, viewing a previous blame looks like this:
image
And in fact VSC might already be ignoring whitespace changes, because I'm only seeing it blame me for non-whitespace changes:
image
But it could just be VSC being weird.

On github for example it will blame me like it does viewing off of my blob: https://github.com/CHOMPStation2/CHOMPStation2/blame/0c06f5f57b167bfc1fc25e92d2da8e6ecf89cccd/code/__defines/misc_ch.dm
But it too easily allows you to view the previous blame: 
image
image

Drulikar:
But if you all would rather me discard the gitattributes changes I can too; but its probably best to standardize the line endings as much as possible sooner than later.

ItsSelis:
> Is it possible to correct line endings without github detecting changes to the entire file? That's the main thing holding up this PR right now as it breaks blame for all modified .dm files.

It's also possible to ignore a (or multiple) commits from being blamed. TG has done too:
https://github.com/tgstation/tgstation/blob/master/.git-blame-ignore-revs

Darlantanis:
Good to merge whenever conflicts are resolved.

PR #7623, opened 2024-01-25
fixes 2 runtimes in the liquid system

Kashargul:
## About The Pull Request

Firstly, the change that splashing is preferred over touching led to runtimes on robots which are not able to hold the variable. So the proc as been restored into its original state and the preferred splashing action moved towards carbons which support it.

Secondly, the reagent application calls itself during the atom check until destruction. With the object moved into null space before deletion, this led to loc.loc ending up as null.loc. We only ever should have this lookup when the object is not already on its deletion path.

## Changelog

:cl:
fix: runtime during belly acid reagent applications on robot robot noms
fix: runtime during acid reagent application on destructible objects
/:cl:

### DIFF
--- a/code/modules/reagents/holder/holder.dm
+++ b/code/modules/reagents/holder/holder.dm
@@ -268,9 +268,13 @@
 //If for some reason touch effects are bypassed (e.g. injecting stuff directly into a reagent container or person),
 //call the appropriate trans_to_*() proc.
 /datum/reagents/proc/trans_to(var/atom/target, var/amount = 1, var/multiplier = 1, var/copy = 0)
-	if(ismob(target))  //CHOMPEdit
-		return splash_mob(target, amount * multiplier, copy) //Touch effects handled by splash_mob
+	//CHOMPEdit Start, we only can do the splashing first on carbon (Runtime reasons)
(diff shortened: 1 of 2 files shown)


code/modules/reagents/holder/holder.dm
```diff
@@ -268,9 +268,13 @@
...
 //call the appropriate trans_to_*() proc.
 /datum/reagents/proc/trans_to(var/atom/target, var/amount = 1, var/multiplier = 1, var/copy = 0)
-	if(ismob(target))  //CHOMPEdit
-		return splash_mob(target, amount * multiplier, copy) //Touch effects handled by splash_mob
+	//CHOMPEdit Start, we only can do the splashing first on carbon (Runtime reasons)
+	if(iscarbon(target))
+		return splash_mob(target, amount * multiplier, copy)
+	//CHOMPEdit End
 	touch(target, amount * multiplier) //First, handle mere touch effects
+	if(ismob(target))
```
Raeschen, line 276: Chompedit comments here and 277 (one below)

Kashargul: no, reverted to upstream code.

PR #7631, opened 2024-01-27
Buffs the speed of all surgeries plus minor change.

tacoguy7765093:
## About The Pull Request

This PR will reduce the speed of most surgeries to 5-6ish seconds with a couple exceptions while also significantly lowering the RNG time between each step, (Some steps can take up to 10 seconds longer or shorter just because). Hopefully this makes surgery just feel less unnecessarily time-sinky

This was originally just a ATK surgery buff but decided to go all the way. Synth surgeries are not touched because power tools are easy to get and very powerful.

Also damaged organs are now able to be re-inserted into bodies if some of you crazys want to completely re-brain someone if they get gibbed.
## Changelog

:cl:
balance: Surgery speed has been massively buffed.
/:cl:

### DIFF
--- a/code/modules/surgery/bones.dm
+++ b/code/modules/surgery/bones.dm
@@ -18,8 +18,8 @@
 	can_infect = 1
 	blood_level = 1
 
-	min_duration = 50
-	max_duration = 60
+	min_duration = 30 //CHOMPedit
+	max_duration = 40 //CHOMPedit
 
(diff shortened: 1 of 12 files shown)


Aroliacue:
This was discussed in our staff dev channels and I think it's fair to say that this should help enormously with the ridiculously long surgery times that can come from very heavily damaged patients. On the flipside, this should also help dissuade chemistry always being the superior option.

PR #7635, opened 2024-01-27
[MIRROR] remove static chat colour tags

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/15695
--------------------
This aims to remove all static chat colour tags and instead rely on the css tags.

Old chat has no changes in colours as the current colours have been added to the default byond style.

For the TG-chat themes, the light themes currently also use the default colours. The dark mode had the colours overhauled to finally be readable.

**!IMPORTANT!
The font color tag is now to be seen deprecated in all chat messages and shall never be used again. Use the new span_col() function to tag messages with colours.**

🆑 Upstream
qol: replaces (hopefully all) static chat colour tags with class tags
/🆑 

[issue link stripped]


Kashargul:
Todo ME

PR #7636, opened 2024-01-27
Remove EVA Slipping, Make gravity 'thunk' less impactful

Aroliacue:
:cl:
balance: Removed the random EVA 'slipping' chance. This contributed nothing besides making spacewalks without magboots incredibly annoying.
balance: The 'thunk' proc (walking onto a gravity tile from zero gravity) now only stuns you for 1 tick (around one second or less) and no longer weakens you. This'll cause you to drop what you're holding but otherwise it's only a brief, mild disruption for not using magboots or walk intenting onto a tile with gravity.
/:cl:

### DIFF
--- a/code/game/area/areas.dm
+++ b/code/game/area/areas.dm
@@ -455,11 +455,11 @@ var/list/mob/living/forced_ambiance_list = new
 			return
 
 		if(H.m_intent == "run")
-			H.AdjustStunned(6)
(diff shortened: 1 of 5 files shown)


tacoguy7765093:
I was going to get around that. Thank you for doing it first
Especially since the overmap is going to be more used in the future, anything to make it less annoying for the sake of realism.

PR #7650, opened 2024-01-29
Map Bugfixing

Missile597:
## About The Pull Request

Fixes a few map-related bugs that I've been logging. Added a seed that I'd missed in a previous PR to the seed dispensers, added some shutters that were missing, fixed an airlock which was popping up an error message at the start of every round and fixed the layout of the shutters on deck three that were cutting off areas without a scrubber/pump. Oh, and a heater got moved too cause it was in someone's way lol

## Changelog

:cl:
add: Cinnamon seeds added to the seed storages in garden and xenobotany
add: Added emergency shutters that were missing on the south part of deck 1
del: Removed two shutter doors in the dorms corridor that were cutting off certain areas from a scrubber or a pump when closed due to an alarm
qol: Moved a space heater that was blocking a trash pile
fix: Fixed the airlock on the Ninja shuttle so the airlock helper doesn't panic when loading up the round. Also moved the button so that it doesn't get left behind when the shuttle jumps (not that it gets used)
/:cl:

### DIFF
--- a/maps/southern_cross/southern_cross-1.dmm
+++ b/maps/southern_cross/southern_cross-1.dmm
@@ -3928,6 +3928,11 @@
 /obj/item/tape/engineering,
 /turf/simulated/floor/tiled/steel_grid,
 /area/hangar/two)
+"arL" = (
+/obj/effect/wingrille_spawn/reinforced,
+/obj/machinery/door/firedoor/border_only,
+/turf/simulated/floor,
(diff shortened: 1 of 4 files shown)


Kazzc:
Amusing, that heater position is such a negligible change. But I might take inspirations from this and intentionally choke maintenance halls on future maintenance map projects.

Thanks for catching all these, good stuff.

PR #7658, opened 2024-01-30
Deus Vault

FluffMedic:
Hullo, I have come with a change no one asked for. Wait, no the vault has been memed on a tad on how you can burrow through maints, or through the janitor's floor and directly into the vault. So I rotated everything by 180 degrees.
There is also some new maints to play with.

## About The Pull Request

image

NOTICE
This pr also alters deck 1, and should proabaly be merged first. I did ask the person what they altered and just mimiced what they told me. Missing shutters on some glass, easy fix.
https://github.com/CHOMPStation2/CHOMPStation2/pull/7650

## Changelog

:cl:
add: More maints near the vault
fix: Rotates the vault 180 degrees
/:cl:

### DIFF
--- a/html/changelogs_ch/AutoChangeLog-pr-7650.yml
+++ b/html/changelogs_ch/AutoChangeLog-pr-7650.yml
@@ -1,0 +1,8 @@
+author: "Missile597"
+delete-after: True
+changes:
+  - rscadd: "Cinnamon seeds added to the seed storages in garden and xenobotany"
+  - rscadd: "Added emergency shutters that were missing on the south part of deck 1"
+  - rscdel: "Removed two shutter doors in the dorms corridor that were cutting off certain areas from a scrubber or a pump when closed due to an alarm"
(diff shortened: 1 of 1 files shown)


TheGreatKitsune:
On a normal server the ability to burrow into vault would actually be intended design. But then again we aren't a normal server so.

TheGreatKitsune:
After some discussion in discord I think a better solution might be to add some sort of reinforced plating that takes more steps to undo, but. I do actually appreciate seeing a maintenance connection there plus a new secret room.

TheGreatKitsune:
I'm just worried that this might make any events involving the vault and being raided by volunteers a bit too difficult

Kazzc:
I like this a lot more, if there is one section of the map that should be a challenge to break in, should be the vault. Will not be surprise if players find creative ways to break in. 
I will note, think you can update this PR once I merge this PR map update? https://github.com/CHOMPStation2/CHOMPStation2/pull/7650

PR #7659, opened 2024-01-31
[MIRROR] fix some more font tags, syncing stylesheets

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/15715
--------------------
This PR focusses on 2 things, firstly, some oversighted font tags displayed in chat.
Secondly, syncing the stylesheet with downstream, so that maintaining the chat is easier for now while still being in the implementation stage. (It merely adds some more colour definitions and fixes some broken ones)

🆑 Upstream
add: timestamps to all radio messages and makes borg radio timestamps depending on the client pref
/🆑


Kashargul:
Warning, not shown conflict in style sheet!

Kashargul:
This one is the correct one, so should be just to accept incoming now, interface existed twice in the sheet.

PR #7667, opened 2024-01-31
Adds neutral trait 'hide'

Raeschen:
Adds teshari's hide for all to use. 0 cost trait.

:cl:
add: 'Hide' neutral trait.
/:cl:


Kazzc:
I'd argue it cost 1 pt at least.
During events or a general player chase this can have an advantage.

Raeschen:
> I'd argue it cost 1 pt at least. During events or a general player chase this can have an advantage.

It's countered by the fact it's a physical hide. Larger creatures will have trouble obscuring themselves behind objects. Small characters deserve that ease anyway.

Cyrelius:
Maybe rather than adding this is a trait, make it a verb that all carbons can use.

Raeschen:
I'd leave it as an option people can choose personally, otherwise all the neutral traits would have that same argument.

PR #7699, opened 2024-02-08
Shadekin rebalance tweaks

Raeschen:
Rebalances shadekin HP/energy recharge/max energy

Nutrition now CANNOT be turned into energy. Being in the dark WILL feed you however, to a barely-satisfied amount.
No shadekin lose energy in the light now (Light is supposed to be not bothersome to shadekin)
Most shadekin have reduced HP except for greens
Most shadekin have increased energy except for purples
Yellows are now GLASS, but have overall best energy performance.

I believe the intention here is to push away from gamer and push towards weird strange otherplanar creature / roleplay instigation. Subject to change.

:cl:
balance: Nutrition cannot generate energy, but darkness generates a little nutrition to a limit.
balance: shadekin BLUE max hp 100 -> 75
balance: shadekin BLUE max energy 100 -> 125
balance: shadekin RED max hp 200 -> 150
balance: shadekin RED energy change in light -1 --> 0
balance: shadekin RED energy change in dark 0.1 --> 0.5
balance: shadekin RED max energy 100 -> 150
balance: shadekin PURPLE max hp 150 -> 100
balance: shadekin PURPLE energy change in light -0.5 -> 0
balance: shadekin PURPLE energy change in dark 1 -> 2
balance: shadekin YELLOW max hp 100 -> 50
balance: shadekin YELLOW energy change in light -2 -> 0
balance: shadekin YELLOW energy max 100 -> 150
balance: shadekin GREEN energy change in dark 2 -> 1
balance: shadekin GREEN max energy 100 -> 125
balance: shadekin ORANGE max hp 175 -> 125
balance: shadekin ORANGE energy change in light -0.5 -> 0
balance: shadekin ORANGE max energy 100 -> 175
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species/shadekin/shadekin.dm
+++ b/code/modules/mob/living/carbon/human/species/shadekin/shadekin.dm
@@ -125,7 +125,7 @@
 	var/doing_phase = FALSE //CHOMPEdit - Prevent bugs when spamming phase button
 	var/manual_respite = FALSE //CHOMPEdit - Dark Respite
 	var/respite_activating = FALSE //CHOMPEdit - Dark Respite
-	var/nutrition_energy_conversion = FALSE //CHOMPEdit - Add toggle to nutrition and energy conversions. FALSE for now for rebalance
+	var/nutrition_energy_conversion = TRUE //CHOMPEdit - Add toggle to nutrition and energy conversions.
 
 /datum/species/shadekin/New()
 	..()
@@ -339,9 +339,9 @@
 		else
 			dark_gains = energy_light
 		//CHOMPEdit begin - Energy <-> nutrition conversion
-		if(nutrition_energy_conversion && get_energy(H) == 100 && dark_gains > 0)
+		if(nutrition_energy_conversion && get_energy(H) == 100 && dark_gains > 0 && nutrition_conversion_scaling > 0)
 			H.nutrition += dark_gains * 5 * nutrition_conversion_scaling
-		else if(nutrition_energy_conversion && get_energy(H) < 50 && H.nutrition > 500)
+		else if(nutrition_energy_conversion && get_energy(H) < 50 && H.nutrition > 500 && nutrition_conversion_scaling > 0)
 			H.nutrition -= nutrition_conversion_scaling * 50
 			dark_gains += nutrition_conversion_scaling
 		//CHOMPEdit end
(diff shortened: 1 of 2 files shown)


Raeschen:
Don't merge until there's an agreement to this in the staff channel

Raeschen:
Yellow health reduced to 50 a bit more from 75

Yellow has the best energy stats of all, so it'll be the glass shade

Allow nutrition from dark but no energy from nutrition

HatArchAzel:
> No shadekin lose energy in the light now (Light is supposed to be not bothersome to shadekin)

**????????????????????????????????????????????????????????????????????????????**

Why though?? The dark creatures from literal The Dark, where one of their primary defining traits is that they want to be **_in the dark_** is now... functionally unaffected by light?? With the combination of this and the energy buffs this feels extremely poorly thought out and seems to achieve exactly the opposite of what it sets out to do. I feel like this will exacerbate SK issues and also is just extraordinarily baffling from a conceptual standpoint.

theXDkidoflol:
as mister ‘anti-shadekin’ this is one step forward and an entire flight of stairs back. 
Hold a vote or something, this is a terrible change.

Raeschen:
> > No shadekin lose energy in the light now (Light is supposed to be not bothersome to shadekin)
> 
> **????????????????????????????????????????????????????????????????????????????**
> 
> Why though?? The dark creatures from literal The Dark, where one of their primary defining traits is that they want to be **_in the dark_** is now... functionally unaffected by light?? With the combination of this and the energy buffs this feels extremely poorly thought out and seems to achieve exactly the opposite of what it sets out to do. I feel like this will exacerbate SK issues and also is just extraordinarily baffling from a conceptual standpoint.

@Kazzc is the forerunner on this part, perhaps he can comment

HatArchAzel:
I really don't think I've seen a PR that I've so immediately and overwhelmingly disliked, and I'm not even someone that has strong opinions on shadekin one way or the other normally to begin with. This feels... incredibly out of touch and I'd really, really urge the staff to reconsider these changes. This hurts the SK identity as a whole while only strengthening the issues others find with them-- it is a loss on all fronts imo.

zeskorion:
if shadekin are no longer losing energy in the light, they should take other penalties- perhaps not being able to phase out?

DerBlitzkrieg:
They're called SHADEkin
They come from the DARK
They fuck with the LIGHTS 
I struggle to actually fathom the thought process behind making these creatures, who have had darkness as part of their core identity, no longer care about it.
From a balance standpoint its laughable and lorewise its even worse. 
Why is the community never included in these major changes pushed by the headmins

theXDkidoflol:
<@91968285012140032> 

as far as im concerned, damage on being out of energy in light would mean shadekin cant scene in light, or at least restrict their options to do so; this damages scening. a lot. It restricts ‘kin to scening somewhere with light.

right now, the main issue I- and most other people who have issues with ‘kin- have, is that there’s LITERALLY nothing you can do to a phased kin. that, and this meaning that phase helps explo kin massively to do their things.

that latter issue is being addressed with POI phase blockers, which is good! but the former isn’t being addressed. there is no possible way to counter a phased kin, **period**, let alone an accessible one.

this current PR is addressing… neither, of those issues. it does address some other minor things, like explo kin rendering medics meaningless and the quite frankly crazy health SOME of them have, but the main darkness change… generates a new problem, instead of fixing any. the darkness weakness isn’t just a good way to catch an unphased kin, its part of their identity. removing this is killing more than their weakness, it’s killing their species identity in exchange for nothing.

that’s my stance on the PR.

Kazzc:
Shadekin have been unchanged for a very long time, looking to rebalance them and hopefully push them to a more roleplay friendly environment. Reducing their health, energy loss from light and increase energy pools to give kin players more options without needing to remain in the dark or removing light sources.

If you are curious about future nerfs here is a list of what I have planned.
Considerations for shadekin phasin.


-Shadekins can phase in dim lit rooms at a higher cost, the brighter the higher cost.
-Security cameras count as an observer and can prevent shadekin from phasing in and out.
-A fully lit room mixed with cameras can outright block shadekin from phasing in and out.
-Shadekin can phase for free in a dark room.
-Shadekin can't phase while being directly observed in a lit room.
-/obj/item/clothing/glasses/meson Meson's can be used to see shadekin that are phased out. 
Note. Meson's can't see players past walls.
-While phased out if a shadekin wanders near any light source say 3x3 tiles, there is a 25% chance will cause lights & flashlights to flicker for a brief moment. 
Note. Triggers every 5 seconds, so standing near a light source will eventually cause it to flicker.
-If a shadekin wanders near a /obj/item/weapon/bluespace_crystal & /obj/item/weapon/bluespace_crystal/artificial say a 3x3 tiles, will force the shadekin to phase out and shatter the bluespace crystal.
Note. Does not trigger bluespace crystal inside inventories.
Note. Bluespace crystal can be thrown and trigger a phase out.

Do note that any changes we make is going to shake shadekin playstyle, but changes need to happen if we want to test and experiment on new concepts. We can always tune and rebalance later down the road but we need to test changes to see what can be improved.

PR #7713, opened 2024-02-09
Ooc panel overhaul (redone)

Kashargul:
## About The Pull Request
Panel:
grafik
grafik

Chatprint:
grafik
grafik

Character Directory:
grafik
grafik

## Changelog

:cl:
qol: allows the ooc panel to be used in field / list mode and adds two more categories for more detailed information
/:cl:

PR #7714, opened 2024-02-09
More whitespace tweaks

Drulikar:
## About The Pull Request

This PR is a follow-up to #7620 reverting #7711 and #7710 because I didn't know chomp had a different editor config that was also managing line endings for dme files.

Of course if you want it a different way that's fine, just it needs to line up with the .gitattributes and then become the standard. I just figured following the standard TG had is likely desirable.

## Changelog

No player facing changes.


Raeschen:
Wait for #7713 so kash doesn't scream

Kashargul:
Merge it, I need to look into the tgui screaming anyway.

Raeschen:
my bad

PR #7749, opened 2024-02-15
Reenables Doorbells 2

Aroliacue:
Please don't mind the filechange size, this is just a removal of some chompedits. CRLF/LF changes haven't been spread to all files and so we'll likely be seeing a huge influx of bloated filechanges for a while. Also FUCK I hate fucking code jesus christ why can't I just remove committed files from a PR AEEUGH.

:cl:
add: Reenabled doorbells on airlocks and knocking on regular doors. This can be done by leftalt-clicking on a door.
/:cl:

PR #7750, opened 2024-02-15
Moar whitespace normalization [MDB IGNORE]

Drulikar:
## About The Pull Request

This PR is a follow up to #7620 and #7714 to normalize more files that some reason didn't get updated in the previous PR. All the files modified here are what results from running `git add --renormalize .` In the previous PRs I was doing `rm --cached -r .` followed by `reset --hard` which *should* have had the same result but I guess not... An example of this is in #7749 where the file being modified is still in crlf.

I also updated the vscode workspace settings for dm EOL; but I don't think it ultimately was affecting this.

## Changelog

No player facing changes


cadyn:
PRBot please be nice to me 🙏

PR #7761, opened 2024-02-16
Attempting to please two crowds RE Shadekins and tcomms

Raeschen:
Adds a anchored emergency closet to the TCOMMS turret room with the crystals. Particularly swift-reacting shadekin might be able to hide from the turrets in here.

This is an attempt to please both "Lethal consequences" and "less than lethal consequences" crowd.

Lethal consequences crowd still have the lethal turret room as the trap
Less than lethal consequences crowd have a locker to try hide in now.

:cl:
add: Emergency closet shelter in the tcomms supplies room
/:cl:

### DIFF
--- a/maps/southern_cross/southern_cross-1.dmm
+++ b/maps/southern_cross/southern_cross-1.dmm
@@ -26916,19 +26916,11 @@
 /turf/simulated/floor/plating,
 /area/construction/firstdeck/construction3)
 "rxy" = (
-/obj/structure/table/standard,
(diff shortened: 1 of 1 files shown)


Skysteak:
Whatever the outcome of this is, thank you for at least trying to find a compromise, with all the potential flak and moaning it may inspire in some people .

For the PR itself? I feel it still keeps the spirit of why there is a lethal danger in Tcomms, whilst not just making the outcome outright 'You die, gg', which in turn can lead to its own interesting stories.

PR #7793, opened 2024-02-20
Dominate Predator : mob/ssd changes

Delams-The-SM:
Gives dominate predator trait ability to dominate mobs / out of body players
Also makes the whitelist from mind binder a global one

## About The Pull Request

Adds ability to dominate mobs via the dominate predator trait, this means that any player with said trait is able to dominate non player mobs with a belly that are on the var/global/list/mob_takeover_whitelist whitelist.

var/global/list/mob_takeover_whitelist became a global whitelist that contains mind binder's whitelist.
The device and dominate predator trait now use it.

Additionally to make it more in line with the mind binder it is now possible to take over bodies without active players in them just like one can do with a mind binder(with proper logging of such occurrences)
That means if for example we would use mind binder to take the mind off someone's body we can then dominate them
SSD players are still safe

## Changelog

:cl:
add: Ability to dominate mob via dominate predator
add: Ability to dominate body without player's mind  (The same way one can do via mind binder)
code: Mob whitelist from mind binder became a global list var/global/list/mob_takeover_whitelist
/:cl:

### DIFF
--- a/code/_helpers/global_lists_vr.dm
+++ b/code/_helpers/global_lists_vr.dm
@@ -61,6 +61,50 @@ var/global/list/item_vore_blacklist = list(
 		/obj/item/weapon/disk/nuclear,
 		/obj/item/clothing/suit/storage/hooded/wintercoat/roiz)
 
+//CHOMPEdit start - Global Whitelist mobs for takeover (mind binder/domination)
+var/global/list/mob_takeover_whitelist = list(
+		/mob/living/carbon,
+		/mob/living/silicon,
+		/mob/living/simple_mob/animal/sif,
+		/mob/living/simple_mob/animal/passive,
+		/mob/living/simple_mob/slime,
+		/mob/living/bot,
+		/mob/living/simple_mob/vore/horse,
+		/mob/living/simple_mob/vore/wolf,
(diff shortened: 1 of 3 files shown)


Kashargul:
Seeing issues with the big dragons and the SCP.

cadyn:
If this is added it should be restricted to a whitelist of certain mobtypes, or I could very easily see issues arising. Kash only mentioned two but a LOT of mobs have bellies that you could potentially force yourself into.

Kashargul:
> If this is added it should be restricted to a whitelist of certain mobtypes, or I could very easily see issues arising. Kash only mentioned two but a LOT of mobs have bellies that you could potentially force yourself into.

Already clarified it we have a whitelist in the mind binder to use with the initial proc.

Raeschen:
Use the mindbinder whitelist, yeah.

Delams-The-SM:
above ideas implemented + what was discussed in #ss13-development about mind binder's abilities
in short :
mind binder whitelist is used (it became a global list)
the trait's behaviour is now overall similar to the mind binder when it comes to new mob interactions and out of body player bodies, while keeping the original pred domination behaviour

Delams-The-SM:
Above review addressed, proper comments are added

PR #7804, opened 2024-02-21
[MIRROR] chat Timestamps for TG chat + some fixes

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/15770
--------------------
The ingame chat timestamps are quite wonky with not being on every message and using server time. It's good for admins to look into things, but players might prefer to use local time. So they easily can use this setting now to enable client side timstamps on all chat messages.

Not sure which update broke the NPC Say filter. Anyway. Reformatted it, now it shouldn't matter anymore.

Also... no more tgui warnings with missing imports...

🆑 Upstream
add: Allows to enable chat timestamps for TG chat using local time
fix: moves the first healthbar proc from 99% to 95% to avoid some edge cases.
fix: fixes a displaying bug in the chat healthbar leading to digesting instead of digested being shown on the final proc
fix: NPC Say Chat filter
fix: respawn time on Process
fix: fixing up the sorting section of the character directory
/🆑 

[issue link stripped]


ItsSelis (changes requested):
tgui.bundle.js and tgui-panel.bundle.js are not allowed to be merged.

ItsSelis (approved):

PR #7822, opened 2024-02-24
Crawling

TheCaramelion:
## About The Pull Request
Adds crawling. Proper crawling.

When laying down, you'll be able to move around at a reduced speed.
When crawling and weakened, you'll move much slower, giving you a chance to escape from mob vore.
You won't be able to crawl if you're stunned, knocked out, or paralyzed, and you'll drop your items on your hands as well.

Turrets will also shoot you, if you can still crawl.

Also all species should be crawling at same speed, no matter the equipment, chemicals or turf they're currently on.

## Changelog

:cl:
add: Crawling.
qol: Laying down won't make you drop items in your hands.
qol: You can now equip/unequip items while laying down.
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -836,6 +836,15 @@
 /mob/living/proc/slip(var/slipped_on,stun_duration=8)
 	return 0
 
+// CHOMPAdd - Drop both things on hands
+/mob/living/proc/drop_both_hands()
+	if(l_hand)
+		unEquip(l_hand)
+	if(r_hand)
+		unEquip(r_hand)
+	return
+// CHOMPEnd
+
 /mob/living/carbon/drop_from_inventory(var/obj/item/W, var/atom/target = null)
 	return !(W in internal_organs) && ..()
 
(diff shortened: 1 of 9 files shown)


FluffMedic:
I better still get a free nom from a flashback.
Blep

code/modules/mob/living/living.dm
```diff
@@ -968,17 +977,28 @@
...
 			unEquip(r_hand)
 		for(var/obj/item/weapon/holder/holder in get_mob_riding_slots())
 			unEquip(holder)
+	*/
 		update_water() // Submerges the mob.
+		stop_pulling() // CHOMPAdd - Stops pulling things
+		pass_flags = PASSTABLE
 	else
 		density = initial(density)
+		pass_flags = 0
```
Raeschen, line 1001: You are changing pass_flags here but many other sources can edit pass_flags.

You're gunna need to store whatever your old passflags were before modifying them

e.g
`pass_flags_precrawl = passflags
passflags |= PASSTABLE
else
pass_flags = pass_flags_precrawl`

Raeschen: Bitflags are also appended, not set. There are several flags in one var

`BITFLAG = A FLAG`
will clear all flags and set it to ONLY your flag.

`BITFLAG |= A FLAG`
will add your flag to all the existing flags set in the var. If A FLAG already exists, then there is no change.

It is also possible to ONLY remove ONE flag from a bitflag var, that is achieved by
`BITFLAG &= ~A_FLAG`

PR #7844, opened 2024-02-28
[READY TO MERGE] Public Sheet Storage, Timeclock Changes

Aroliacue:
:cl:
qol: Timeclock cooldown down to 1 minute, everyone can set anyone else to off-duty.
maptweak: Moved science's sheet storage to lobby to make it publicly accessible. Changed it to persistent lossy, which means stored sheets will transfer between rounds at a loss.
/:cl:

### DIFF
--- a/code/game/machinery/computer/timeclock_vr.dm
+++ b/code/game/machinery/computer/timeclock_vr.dm
@@ -221,7 +221,7 @@
 /obj/machinery/computer/timeclock/proc/checkCardCooldown()
 	if(!card)
 		return FALSE
-	var/time_left = 10 MINUTES - (world.time - card.last_job_switch)
+	var/time_left = 1 MINUTE - (world.time - card.last_job_switch) // CHOMPedit: 10 minute wait down to 1 minute.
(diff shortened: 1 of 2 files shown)


Aroliacue:
image

CMDRYouth:
Now miners can use all those Sheet Snatchers themselves, and they won't need Science to be present to drop off materials.
Similarly, now people can withdraw materials (IE Medical with Phoron) for their own wants and needs.

CMDRYouth:
Downside:  Scientists will lose that specific spot as a popular Battery Charger location.  They might have to start using the other window.

Aroliacue:
image
The sheet storage has been moved to cargo instead. It'll still retain its public access, though cargo has more oversight over it now.

Kazzc:
Think you can make an admin-log anytime the timeclock is used? location/ckey/item logs will be helpful.

Aroliacue:
This PR is finished and ready to be merged (once feature freeze is over)

Raeschen:
this one gets a pass due to being a TM before the freeze

PR #7857, opened 2024-03-01
Configuration Controller

ItsSelis:
# Attention downstreams
You **will** have to edit your configuration files to make stuff still work.
Changes that are required were made in the examples, so its best to check the diffs.

The configuration examples have been updated to how the configuration keys (for example PROBABILITY changed to PROBABILITIES) should look like.

The game will write debug logs in case it does not recognize a specific configuration string when booting up.

## About The Pull Request

This PR ports over tg's configuration controller.

There are also a bunch of config entries (marked with FIXME in either of the config files inside of config or code/controllers/configuration_ch/entries) that seem to be unused or not polled nor set by any code.

Todo:
- [ ] Updating config/examples
- [x] Solving issues with gamemode selection
- [x] Lobby tab seems to be still empty
- [x] Making Holidays work
- [x] Making ENABLE_STAT_TRACKING work
- [x] Fixing /datum/config_entry/keyed_list/event_first_run
- [x] Events too frequent
- [x] Revision not printing correctly
- [x] Jukebox tracks are added twice to the config variable.
- [x] Some of the config strings not properly working (see runtimes). This is most likely due to some stuff trying to fetch 
something from the config before the configuration controller is fully loaded but already created.
- [x] Jukebox broken again
- [x] the brain (/obj/item/organ/internal/brain/vox): New(the plating (197,246,7) (/turf/unsimulated/floor)) causing error while fetching default_brain_health
- [x]  the brain (/obj/item/organ/internal/brain/vox): New(the plating (197,246,7) (/turf/unsimulated/floor)) causing error while fetching defib_timer

## Changelog
:cl:
refactor: Replaced the old configuration with the configuration controller.
admin: Added "Reload Configuration" button to Debug tab (requires DEBUG permission), to reload config from files.
/:cl:


Kashargul (approved):
Except for the example still using url tags and not plain urls as we need for TG-Chat the rest is fine. Should maybe have that MOTD example updated, too.

PR #7859, opened 2024-03-01
Tesh Steppies

Darlantanis:
Adds new default step sounds for tesharii. This will only affect turfs using the default steps; anything using a different set of footsteps will still use those for tesharii. 
-Turfs now support different footstep sounds for different species. Just use SPECIES_WHATEVER defines when adding footseps

Sounds are unfortunately a bit echoe-y at higher volumes, but this is the best quality I've been able to manage.
https://gyazo.com/da2cbd97ae99146cf8cd6871c705a378

https://www.youtube.com/watch?v=Wt7LJ-7rh0U

## Changelog
:cl:
add: Turfs now support different species footstep sounds
add: Adds new default step sounds for tesharii
/:cl:


TheCaramelion:
This is pretty neat, it's gonna be fun listening to wet steppies from Prometheans. :3c

Raeschen:
Unless the intention is just to have it working only on untiled turfs, recommend you also add your sounds here:

https://github.com/CHOMPStation2/CHOMPStation2/blob/master/code/game/turfs/flooring/flooring.dm#L399

What you could do is make defines for each material property as well. But this is just a future suggestion, likely out of scope for this. 

e.g 
#define FOOTSTEP_SOUNDS_METALLIC ...
#define FOOTSTEP_SOUNDS_SOFT ...
#define FOOTSTEP_SOUNDS_GRASS ...

Darlantanis:
This is intentional as these are meant as tesh hull sounds. base_footstep_sounds should be used by any simulated floor that lacks a flooring var (unless Init is overwritten). That said, I would not mind if these are used for more turfs if it proves popular.

Turf sounds would probably be best converted to the component system, but that is currently beyond my skill level.

PR #7883, opened 2024-03-04
Ion Storm Trigger

TheCaramelion:
## About The Pull Request
One small problem with this event: It requires an AI to be active. AI players are very rare, but we have quite a bunch of borg players; Changing this to require at least two borgs to be online and one Scientist that could hopefully reset the borg's laws if they happened to receive an Ion law.

Currently it requires 1 AI player
This would change it to 3 borgs needed, plus 1 Scientist

## Changelog
:cl:
balance: Modified Ion Storm requirements to be more common
/:cl:

### DIFF
--- a/code/modules/events/ion_storm.dm
+++ b/code/modules/events/ion_storm.dm
@@ -4,6 +4,7 @@
 	has_skybox_image = TRUE
 	announceWhen = -1 // Never (setup may override)
 	var/botEmagChance = 0 //VOREStation Edit
+	var/ionBorgs = TRUE // CHOMPStation Edit
 	var/cloud_hueshift
 	var/list/players = list()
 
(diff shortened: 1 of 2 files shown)


Skysteak:
I would personally prefer this to take the classic single AI and/or Three Borgs. I do slightly worry that it requiring two borgs may _just_ push it into being more annoying for everyone, but that's just my opinion.

SpaceMan-Jim:
As a AI 'player'. I never seen the actual event proc more than once. Is it just AI needed, or does it need borgs aswell?

TheCaramelion:
> I would personally prefer this to take the classic single AI and/or Three Borgs. I do slightly worry that it requiring two borgs/one Sci may _just_ push it into being more annoying for everyone, but that's just my opinion.

Changing it to 3 borgs, but still keeping the 1 Sci in. This should give borg players to have a chance to get their laws reset back to normal.

> As a AI 'player'. I never seen the actual event proc more than once. Is it just AI needed, or does it need borgs aswell?

Currently it needs 1 singular AI, but borgs, engineers and scientists make this chance higher.

Issue #7911, opened 2024-03-08
[GENERAL] You can ascend from space in deck one through the floor of deck 2.


Raeschen:
Go to deck 1
go in space somewhere where there is station above you
ascend. You phase through the floor.

PR #7912, opened 2024-03-08
fix ascending through floors

Raeschen:
Bug was introduced here:
#6889 
Specifically:
https://github.com/CHOMPStation2/CHOMPStation2/pull/6889/files#diff-284f84b4d2e4bef4fe808771c2a07749f5f1513a292ea10f225435fa0d983e16L49

The check was accidently only made to work on going down instead of both directions when the special case was added.

:cl:
fix: fixed ascending through floors
/:cl:


Raeschen:
Fixes #7911

PR #7936, opened 2024-03-11
Optimisations: Cache-ing ghosts/NewPlayers

Raeschen:
## About The Pull Request

Caches player-owned /mob/observer/deads and /mob/new_players as these are reused frequently

## Changelog
🆑
refactor: Refactor how /mob/new_player and /mob/observer/dead are handled (one of each is cached per ckey instead of destroy/new)
/:cl:


Raeschen:
I don't think this is necessary with how far we've come. I'll save it somewhere locally just incase.

PR #7948, opened 2024-03-12
Properly defines a few global vars

Raeschen:
reset because oops


Raeschen:
The upstream merge of their painter happened during this, so you may want to check ther non-modular gear_tweaks.dm

PR #7951, opened 2024-03-12
Switch nanomap rendering to spacemanDMM

cadyn:
## About The Pull Request

Fix nanomap generation by switching it to use SpacemanDMM. Also makes it render everything on map instead of just tiles.

## Changelog

:cl:
add: Nanomap now renders non-tile objects
/:cl:


ItsSelis (changes requested):

Raeschen:
changing the dir means you'll have to change the part of the code that requires the minimap too

cadyn:
It doesn't change the dir, just the intermediate dir. The place where it
ends up in icons is the same.

PR #7958, opened 2024-03-14
the smoothening

cadyn:
## About The Pull Request
Set world fps to 40 and add movement smoothing

note: ticklag should be set to 0.25 in config or it will continue running at 20fps

try test merging this (and setting ticklag to 0.25 in config) and see if anything breaks. Everything seemed to work fine in testing.
## Changelog
:cl:
add: movement smoothing
config: 40fps time
/:cl:


cadyn:
todo for lazy self or somebody else:

- [ ] fix ice (probably set to 0.5)
- [ ] try to fix ghost jumping around issues (ghost too fast, jump ahead of following mob, looks quite jarring)

PR #7965, opened 2024-03-14
Adds boneless trait

TheGreatKitsune:
## About The Pull Request

Boneless trait, for custom species organics only. Cannot break or dislocate limbs, but damage is spread and max limb damage is set to 45% of the normal limb's damage amount.

## Changelog

:cl:
add: New trait boneless
/:cl:

### DIFF
--- a/modular_chomp/code/modules/mob/living/carbon/human/species/station/traits/neutral.dm
+++ b/modular_chomp/code/modules/mob/living/carbon/human/species/station/traits/neutral.dm
@@ -16,3 +16,18 @@
 /datum/trait/neutral/hide/apply(var/datum/species/S,var/mob/living/carbon/human/H)
 	..()
 	H.verbs |= /mob/living/proc/hide
+
+/datum/trait/neutral/boneless
+	name = "Boneless"
+	desc = "You have no bones! Though your limbs are also easier to gib in exchange."
+	cost = 0
+	custom_only = TRUE
+	can_take = ORGANICS
+
+/datum/trait/neutral/boneless/apply(var/datum/species/S,var/mob/living/carbon/human/H)
+	..()
+	for(var/obj/item/organ/external/ex_organ in H.organs)
+		ex_organ.cannot_break = 1
+		ex_organ.dislocated = -1
+		ex_organ.spread_dam = 1
+		ex_organ.max_damage = floor(ex_organ.max_damage * 0.45)


cadyn:
I like the idea of this for roleplay purposes but it should probably be a positive trait (2 cost seems reasonable just as a starting point take that with a grain of salt), and should definitely conflict with any damage resistance traits to get ahead of people finding ways to gamer this

HatArchAzel:
Very curious to see how this plays out playtesting-wise-- when I suggested it I definitely imagined the baked in downsides being pretty significant, and if I'm reading this right they definitely seem to be.

May need expansion at some point on account of surgery- with no bones, you have no ribcage to clamp during torso surgery nor skull to get in the way of brain implants, but this already avoids most of the weirdness I was worried about.

TheGreatKitsune:
Ah, I can address the issues for surgery. As for balance wise, this is... actually worse then Promethean stats for limbs.

HatArchAzel:
Echoing my thoughts from the discord here for more formal input- right now this in theory is a pretty big net negative, since losing a limb as a non-prommie/non-xenochim is generally pretty catastrophic, much moreso than having one break. With a low limb health threshold, getting into the same situation as a prommie would (as the only other real comparison point for this being a regular threat to a character) is MASSIVELY more punishing for a boneless character- they have no way to regenerate those limbs and would require more involved help to fix. Imagine getting hit in the chest and that damage spread suddenly takes off both of your arms. Imagine losing your leg while not on station, especially in gateway conditions. It'd suck a lot, to say the least- though I think testing will help a lot to form a more informed stance.

HatArchAzel:
Also have concerns about this potentially applying to decapitation as well and making it stupidly easy to just get instakilled as a result but I dont know whether it actually does that, admittedly

cadyn:
I think my main concern is that it's uncharted territory. Part of the balance for prommies is how they interact with meds and water, which this doesn't have. I would say it should conflict with damage resist traits (also probably the bone traits?) at the very least.

HatArchAzel:
Oh yeah, since it's also been discussed previously, should also make this incompatible with the other bone break traits too, since I don't imagine the two will exactly interact properly even outside of potential balance issues.

HatArchAzel:
> I think my main concern is that it's uncharted territory. Part of the balance for prommies is how they interact with meds and water, which this doesn't have. I would say it should conflict with damage resist traits (also probably the bone traits?) at the very least.

I think it's probably fine to at least test as-is-- losing a limb as a non-prommie character is hugely more impactful than as a prommie, so their downsides are inherently compensated for here by the sheer impact of every time the perk actually comes into play. Always room to give it a balancing pass later if things do end up being a problem, though.

cadyn:
> > I think my main concern is that it's uncharted territory. Part of the balance for prommies is how they interact with meds and water, which this doesn't have. I would say it should conflict with damage resist traits (also probably the bone traits?) at the very least.
> 
> I think it's probably fine to at least test as-is-- losing a limb as a non-prommie character is hugely more impactful than as a prommie, so their downsides are inherently compensated for here by the sheer impact of every time the perk actually comes into play. Always room to give it a balancing pass later if things do end up being a problem, though.

Honestly I hadn't considered the fact that this doesn't include limb regeneration, perhaps in that case it would be better to include a fixed brute damage resistance and burn damage multiplier similar to prommies, and then to lock out the damage mod traits. In general I just think it's better to heir on the side of caution with trait combos, since while individual traits can be broken, it's trait combos that have historically been the most problematic.

HatArchAzel:
As things stand right now, this definitely needs further work before it can be merged- my own attempts at testing failed to get limbs to gib _at all_ with this, even with thousands of damage applied. Will happily test more once it's working, though.

tacoguy7765093:
I wonder if #8219 was why it broke, we should test it again

HatArchAzel:
Definitely worry this is a lot more severe than I was imagining, but still *very* eager to try this out now that it's working.

modular_chomp/code/modules/mob/living/carbon/human/species/station/traits/negative.dm
```diff
@@ -101,3 +101,20 @@
 	name = "Carbon Dioxide Breather"
 	desc = "You breathe carbon dioxide instead of oxygen, much like a plant. Oxygen is not poisonous to you."
 	var_changes = list("breath_type" = GAS_CO2, "exhale_type" = GAS_O2, "ideal_air_type" = /datum/gas_mixture/belly_air/carbon_dioxide_breather)
+
+/datum/trait/negative/boneless
+	name = "Boneless"
+	desc = "You have no bones! Though your limbs are also much, much easier to gib in exchange. (Seriously this can result in one shot deaths and similar)"
+	cost = -4
```
Cameron-The-Raven, line 108: honestly if we're going by 'extremely severe traits' it should be -8 just like bioincompatible major negative trait

SpadesNeil:
# :b:oneless

modular_chomp/code/modules/mob/living/carbon/human/species/station/traits/negative.dm
SpadesNeil: I personally think, start it here, don't over-tune it.

If after players use it and decide "Wow this really sucks and is totally not worth it" _then_ make the value a stronger reduction.

SpadesNeil:
Oh yeah, if boneless is enabled, should 'leave remains' be automatically disabled? Since, bones... but boneless...

PR #7971, opened 2024-03-15
[MIRROR] hands over the round ID to TGchat

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/15823
--------------------
Requires round IDs and DB subsystem

Downstream: Uncomment the round ID handover, we already can use it

🆑 
fix: handing over the round ID to the chat properly
code: optimizes chat export verb
/🆑


Raeschen:
needs bundle removing

PR #7992, opened 2024-03-17
Remove recursivemove + signals from being on EVERY machine by default

Raeschen:
## About The Pull Request

there are about 35k machines that all have this on and never really need it. 

This saves about: 35,000 x (datum + list() element + list() element)

## Changelog
:cl:
refactor: Remove obj/machinery recursivemove placement + signal registration as a default, instead an optional
/:cl:

### DIFF
--- a/code/game/machinery/machinery_power.dm
+++ b/code/game/machinery/machinery_power.dm
@@ -104,7 +104,7 @@
 	. = ..()
 	update_power_on_move(src, old_loc, loc)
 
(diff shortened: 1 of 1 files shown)


cadyn:
This is a bit reckless. Recommend trying to preserve functionality.

Raeschen:
> This is a bit reckless. Recommend trying to preserve functionality.

that's why i'm testing it. Most if not all obj/machines are always on a turf, and they have their own move signal monitoring. Know any cases that go against this?

PR #8003, opened 2024-03-18
[MIRROR] fixes execution order for stair pulling

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/15838
--------------------
Somehow, we were checking the object pulling only after we have moved... force move breaks pulling though, in the case of normal stair use, pulling objects along never could've worked. We first need to check if we are pulling and assign it to a variable then move ourselves.

🆑 Upstream
fix: fixed and issue preventing objects to be pulled through stairs
/🆑


Kashargul:
Already changed downstream

PR #8006, opened 2024-03-18
Delete tgui/packages/tgui_ch

Kashargul:
## About The Pull Request
With TGUI 5, we moved to a much better modular system. Most UI's since then had overhauls to work properly under TGUI 5. The code in this folder is seen as deprecated for a while. No more need to carry it along.
## Changelog
:cl:
del: tgui_ch
/:cl:


ItsSelis:
There still might be references to tgui_ch inside
- .eslintignore
- .prettierignore
- tsconfig.json

Kashargul:
> There still might be references to tgui_ch inside
> 
> * .eslintignore
> * .prettierignore
> * tsconfig.json

Gonna check those later

PR #8007, opened 2024-03-19
Tgpanel

cadyn:
## About The Pull Request
it is TGPanel time my boys. Apparently making this switch actually fixes problems with client FPS, to the point you can actually run the game smooth as hell now.
Suggestion: test merge
Suggestion: make an announcement explaining how to set client FPS so the boys can get the most out of this, recommend 100
## Changelog
:cl:
add: TGPanel (cred Selis for port)
/:cl:


code/game/atoms.dm
```diff
@@ -247,7 +247,8 @@
 
 // Don't make these call bicon or anything, these are what bicon uses. They need to return an icon.
 /atom/proc/examine_icon()
```
Kashargul, line 249: Doing this could interfere with procs like on the gargoyle. it called . = ..()

In my tests, we could drop almost all examine_icon. Only relevant might be the gargoyle.

Raeschen:
Something in those two reverted commits is causing crashlooping i think. Reverting to test.

Raeschen:
Something in the last four reverts was potentially causing crashes. Reverting to test for stability...

Kashargul:
This fixed borgs, greatwol and other large sprites we use... Sadly still no lead why dragon icons are not working.

PR #8017, opened 2024-03-20
Shadekin related fixes

TheCaramelion:
Some shadekin stuff I've been meaning to fix, oop. Also creatures without lungs won't cough anymore when exposed to smoke.

The stuff in resizing_vr is now handled by Bump(), so that should cover all cases of kin getting knocked over or stepped on while phased.
## Changelog
:cl:
fix: Phased shadekins being able to be attacked
fix: Bump should ignore phased shadekins now
fix: Smoke won't make creatures without lungs cough
/:cl:


cadyn:
hsadekni...

PR #8019, opened 2024-03-20
Outsider SuperPose pods now require power plus misc improvments.

tacoguy7765093:
## About The Pull Request
This PR aims to make most of the outsider pods require power now, all of the pods have been supplied with their own power generation methods depending on the pod to make it self sustaining, every pod at the very least has a pacman and battery rack mapped in, or if it doesn't, it has the methods of building one.

Some pods do have RTGs for power generation but those create low amounts of power passively, most pods also have a solars crate thrown in for building solar panels.

DIY pods now have engineers blueprints included in them for custom buildings.

Most of the "Survival Pods" still have infinite power for the consistency with other non-superposed pods. This might need adjusting with stuff like the science pods.

My aim for this PR was to.
1. Fix a bug where you were unable to place ANY wall mounted machinery due to requires_power being set to true.
2. Make pods less set and forget.
3. Encourage multiple outsiders or outsiders to interact with the station for things like power.

This PR also removes some broken pods because I was unable to get them working, those pods were not practical for outsider use anyway.
## Changelog
:cl:
del: Removed the Crashed Alien Ship, Large Alien Ship, Merc ship outsider pods due to them not properly spawning.
balance: Outsider pods now require power, but all outsider pods now have the means to generate their own power.
maptweak: All outsider pods rewired and now with power generation options.
maptweak: Removal of restricted head consoles from science pods.
maptweak: Indestructible walls removed off certain outsider pods like the trade ship.
maptweak: DIY outsider pods now have non-CE blueprints for off-station construction projects.
/:cl:

### DIFF
--- a/modular_chomp/code/modules/mining/shelter_atoms_ch.dm
+++ b/modular_chomp/code/modules/mining/shelter_atoms_ch.dm
@@ -1,5 +1,6 @@
 // Unique areas were requested for each superpose map for future use. The plain superpose area is for future varedits to all superpose submaps.
 /area/survivalpod/superpose
+	requires_power = TRUE
 
 /area/survivalpod/superpose/CrashedInfestedShip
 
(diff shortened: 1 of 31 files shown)


Raeschen:
Very good

PR #8098, opened 2024-03-29
Statpanel + holder runtimes

Raeschen:
:cl:
fix: statpanel runtimes
fix: mob holder runtime
/:cl:


code/modules/mob/holder.dm
```diff
@@ -1,4 +1,4 @@
-var/list/holder_mob_icon_cache = list()
+ var/list/holder_mob_icon_cache = list()
```
Kashargul, line 1: horror! indent is off by 1

PR #8122, opened 2024-03-31
Makes the trash eat message ambigious.

tacoguy7765093:
## About The Pull Request
This was a feature we had that broke from a recent Virgo mirror, this fixes it.
## Changelog
:cl:
add: The trash eat verb now gives a more ambiguous message instead of 'swallowing' regardless of  vorgan.
/:cl:

### DIFF
--- a/code/modules/vore/eating/living_vr.dm
+++ b/code/modules/vore/eating/living_vr.dm
@@ -969,7 +969,7 @@
 
 		else
 			to_chat(src, "<span class='notice'>You can taste the flavor of garbage. Delicious.</span>")
-		visible_message("<span class='warning'>[src] demonstrates their voracious capabilities by swallowing [I] whole!</span>")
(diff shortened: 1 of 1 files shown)


code/modules/vore/eating/living_vr.dm
```diff
@@ -969,7 +969,7 @@
 
 		else
 			to_chat(src, "<span class='notice'>You can taste the flavor of garbage. Delicious.</span>")
-		visible_message("<span class='warning'>[src] demonstrates their voracious capabilities by swallowing [I] whole!</span>")
+		visible_message("<span class='warning'>[src] demonstrates the voracious capabilities of their [lowertext(vore_selected.name)] by making [I] disappear!</span>")
```
Kashargul, line 972: Missing CHOMPEdit

as voregan is references, you want to use vwarning as class.

PR #8143, opened 2024-04-03
Confirmed outbreak of a less stupid biohazard system

SpadesNeil:
## About The Pull Request
I call it stupid but really this is just a result of really old Baycode that no one ever bothered to clean up or make consistent in any way so people have just been sort of blindly doing their own thing because nobody told them otherwise.

From now on.
Level 7 biohazard = stuff that's a danger but isn't immediately deadly. Level 5 biohazard = more severe stuff that you ought to react to like right now. The lower the number goes, the more urgent it is to handle. Level 7 is just the start of things that all crew have to react to.  Basically, the lower the number, the more urgent the issue is. A hypothetical level 1 biohazard would be just "the station is no man's land, arm the nuke, it must never be allowed to spread, nothing may survive, and quarantine the area afterwards just to be safe". Like that's a universe level threat kind of biohazard and will never be encountered in gameplay.

Furthermore, I made "Biohazard Alert" actually tell you what the fucking alert is now. So no more stupid "Wtf is a level 5 again?" "It's virus" "No it's vines" "No it's wallrot." Turns out, the alerts were made up and the numbers didn't matter. I fixed that.

Vines no longer trigger a level 5 alert also.

In conclusion, this is not a good solution but it's better than what we had.

Really we should just redo literally all the AI voice lines, but that's out of the scope of this PR.

## Changelog

:cl:Ace
qol: Made Biohazard levels actually tell you what they are.
/:cl:

### DIFF
--- a/modular_chomp/code/modules/event/infectedroom.dm
+++ b/modular_chomp/code/modules/event/infectedroom.dm
@@ -90,7 +90,7 @@
 		message_admins("Infected room event started; Virus: <a href='?src=\ref[virus];[HrefToken()];info=1'>[virus.name()]</a>")
 
 /datum/event/infectedroom/announce()
-	command_announcement.Announce("Confirmed outbreak of level 7 biohazard aboard \the [location_name()]. All personnel must contain the outbreak.", "Infectious Contaminant", new_sound = 'sound/AI/outbreak7.ogg')
+	command_announcement.Announce("Confirmed outbreak of level 7 biohazard aboard \the [location_name()]. All personnel must contain the outbreak.", "Infectious Contaminant in [target_area.name]", new_sound = 'sound/AI/outbreak7.ogg')
 	// Chomp edit: Changed "Biohazard Alert" to "Infectious Contaminant" and also changed level 5 to level 7.
 
 /datum/event/infectedroom/start()
--- a/modular_chomp/code/modules/event/infectedroom.dm
+++ b/modular_chomp/code/modules/event/infectedroom.dm
@@ -91,7 +91,6 @@
 
 /datum/event/infectedroom/announce()
 	command_announcement.Announce("Confirmed outbreak of level 7 biohazard aboard \the [location_name()]. All personnel must contain the outbreak.", "Infectious Contaminant in [target_area.name]", new_sound = 'sound/AI/outbreak7.ogg')
-	// Chomp edit: Changed "Biohazard Alert" to "Infectious Contaminant" and also changed level 5 to level 7.
 
 /datum/event/infectedroom/start()
 	var/obj/effect/decal/cleanable/mucus/mapped/M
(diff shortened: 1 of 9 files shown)


TheCaramelion:
These are neat announcements

SpadesNeil:
I can't tell if the space vines files are duplicates or one isn't used but I changed both to be safe.

modular_chomp/code/modules/event/infectedroom.dm
```diff
@@ -90,8 +90,8 @@
 		message_admins("Infected room event started; Virus: <a href='?src=\ref[virus];[HrefToken()];info=1'>[virus.name()]</a>")
 
 /datum/event/infectedroom/announce()
-	command_announcement.Announce("Confirmed outbreak of level 5 biohazard aboard \the [location_name()]. \
-	All personnel must contain the outbreak.", "Biohazard Alert", new_sound = 'sound/AI/outbreak5.ogg')
+	command_announcement.Announce("Confirmed outbreak of level 7 biohazard aboard \the [location_name()]. All personnel must contain the outbreak.", "Infectious Contaminant", new_sound = 'sound/AI/outbreak7.ogg')
```
SpadesNeil, line 93: I honestly am considering going back and changing this to instead be like, a janitorial issue, and make it not a level 7 alert, but a "hey janitor the air filters backfired pls go clean up" or whatever.

Right now it's level 7 because people _could_ be sick and it _could_ spread but this is not an ideal solution because right now it doesn't tell you where the infected room actually is.

SpadesNeil: `Infectious Contaminant` to `Infectious Contaminant in [target_area.name]` would that make the name of the infected room show up?

TheCaramelion: target_area might work, but yes.

modular_chomp/code/modules/event/infectedroom.dm
```diff
@@ -90,7 +90,7 @@
 		message_admins("Infected room event started; Virus: <a href='?src=\ref[virus];[HrefToken()];info=1'>[virus.name()]</a>")
 
 /datum/event/infectedroom/announce()
-	command_announcement.Announce("Confirmed outbreak of level 7 biohazard aboard \the [location_name()]. All personnel must contain the outbreak.", "Infectious Contaminant", new_sound = 'sound/AI/outbreak7.ogg')
+	command_announcement.Announce("Confirmed outbreak of level 7 biohazard aboard \the [location_name()]. All personnel must contain the outbreak.", "Infectious Contaminant in [target_area.name]", new_sound = 'sound/AI/outbreak7.ogg')
 	// Chomp edit: Changed "Biohazard Alert" to "Infectious Contaminant" and also changed level 5 to level 7.
```
TheCaramelion, line 94: No need for a chomp edit here, it's in the modular folder

PR #8149, opened 2024-04-04
Adjust Growth/Shrink Trait

Sharkmare:
Both traits are now in line with the per tick strength of macro/micro shrooms. Also we now perform the same checks as on shrooms to allow for oversizing in dorms
## About The Pull Request
Adjusts strength of the Size gain/loss traits to be in line with other similiar code
## Changelog
Growth and Shrinking power of the Size traits is now in line with size shrooms and respects dorms uncapped growth
:cl:
change: minorly stronger growth/shrink effect on size traits
fix: Traits now respect dorms uncapped size
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -1220,12 +1220,12 @@
 			if(!isnull(mod.metabolism_percent))
 				nutrition_reduction *= mod.metabolism_percent
 		//CHOMPEdit Begin
-		if(nutrition > 1000 && species.grows && size_multiplier < RESIZE_HUGE)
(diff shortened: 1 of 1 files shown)


Raeschen:
Does this prevent you from resizing yourself huge and just exiting dorms?

Sharkmare:
> Does this prevent you from resizing yourself huge and just exiting dorms?

The mechanism for resizing people when leaving dorms has nothing to do with this

Raeschen:
> > Does this prevent you from resizing yourself huge and just exiting dorms?
> 
> The mechanism for resizing people when leaving dorms has nothing to do with this, this just makes the trait behave like other already existing resize code including shrooms and sizeguns

Nonetheless, does it?

PR #8170, opened 2024-04-05
Fixes Viro, hopefully

FluffMedic:
Adds a camera and airlock control to viro
## Changelog
:cl:
add: Added camera and airlock controller to viro
/:cl:


Raeschen:
Beware, this will conflict with #8146 which is now TM'd.

PR #8207, opened 2024-04-10
Kinetic Things and Weird Goggle

FluffMedic:
Okay
Does a funny fix/balance change? for rare kinetic weaponry
And adds an admin spawn item, mostly for discussion and expirmentation around such an effect existing.
## Changelog
:cl:
add: Added admin spawn set of goggles that's sole purpose is seeing kin (Rae: commented out the meson/kin hybrid though)
balance: Makes it so the large kinetic weapons don't need wielding since that code is appears broken and I doubt the two very rare items will become broken if they get the kinetic blast effect.
/:cl:


Kazzc:
I like this a lot, may come in handy with some of my events.

Kashargul:
Were those tested? Invisibility used here was never defined as plane, so this might runtime horribly.

modular_chomp/code/modules/clothing/glasses/glasses.dm
```diff
@@ -6,3 +6,28 @@
...
+
+	icon = 'modular_chomp/icons/inventory/eyes/glasses.dmi'
+	icon_override = 'modular_chomp/icons/inventory/eyes/mob.dmi'
+	icon_state = "darksight"
+
+	action_button_name = "Toggle Goggles"
+
+	toggleable = 1
+	enables_planes = list(INVISIBILITY_SHADEKIN)
+
```
Kashargul, line 23: This is not a plane.

~~enables_planes = list(INVISIBILITY_SHADEKIN)~~


see_invisible = INVISIBILITY_SHADEKIN

modular_chomp/code/modules/clothing/glasses/glasses.dm
```diff
@@ -6,3 +6,28 @@
...
+/obj/item/clothing/glasses/meson/darksight
+	name = "optical meson scanner"
+	desc = "Used for seeing walls, floors, and stuff through anything."
+	icon_state = "meson"
+	item_state_slots = list(slot_r_hand_str = "meson", slot_l_hand_str = "meson")
+	action_button_name = "Toggle Goggles"
+	origin_tech = list(TECH_MAGNET = 2, TECH_ENGINEERING = 2)
+	toggleable = 1
+	vision_flags = SEE_TURFS
+	enables_planes = list(VIS_FULLBRIGHT, VIS_MESONS, INVISIBILITY_SHADEKIN)
```
Kashargul, line 33: enables_planes = list(VIS_FULLBRIGHT, VIS_MESONS, ~~INVISIBILITY_SHADEKIN~~)
see_invisible = INVISIBILITY_SHADEKIN

Kashargul (changes requested):

modular_chomp/code/modules/clothing/glasses/glasses.dm
Kashargul: if making it consistent, SEE_INVISIBILITY_SHADEKIN

create a new define in misc.dm with that name, value 55

Kashargul (approved):

Raeschen:
not the keenest on shadekin sight, even if adminspawn

Raeschen:
IMO, the meson/darksight hybrid shouldn't exist. The sole-shadekin sight glasses ARE ok i guess if only adminspawn, but i really want to keep an ability like this purely to a specialised thing.

PR #8219, opened 2024-04-11
Fix config multiplier values

### LINKED ISSUE #8218: [GENERAL] Bone break sounds don't play at all
#### Brief description of the bug
Bone break sounds do not actually emit sounds in any means necessary.

When a limb has high enough brute damage, it should make a crunch sound. bonebreak.ogg. The only situation that makes bone sounds is using admin tools to apply thousands of damage.

#### What actually happened
No sounds are played, this applies to gib, harm intent steppies at 600%, high brute damage vorgans, attacking with weapons. However, using admin tools to apply thousands of damage makes bone sounds still, but that might be a different proc being triggered all together.

#### Steps to reproduce
- (Step 1) Spawn a carbon mob
- (Step 2) Attack the same limb with a weapon, or set your size to 600% and harm-intent step on them, or use a high damage brute gut.
- (Step 3) No crunch is heard,.

#### Code Revision
 

Raeschen:
Fixes #8218 

:cl:
fix: fixed a couple bad config multipliers
/:cl:

### DIFF
--- a/code/controllers/configuration_ch/entries/game_options.dm
+++ b/code/controllers/configuration_ch/entries/game_options.dm
@@ -12,10 +12,12 @@
 /datum/config_entry/flag/limbs_can_break
 
 /datum/config_entry/number/organ_health_multiplier
-	default = 1
+	integer = FALSE
+	default = 1.0
 
 /datum/config_entry/number/organ_regeneration_multiplier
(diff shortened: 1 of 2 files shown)


ItsSelis (approved):
🦴💥

PR #8228, opened 2024-04-14
port TG title shenanigans

cadyn:
## About The Pull Request
Ports some stuff from TGStation to help the splash screen actually be consistent. Basically the issue is that atoms in nullspace is somewhat of undefined behavior in terms of displaying things, even in the client screen variable. mob/new_player now has an actual location, which is setup to show the splashscreen, and is on the centcomm Z-level

## Dear downstreams/mappers
You will need to map in the splashscreen object as has been done on the centcomm Z-level here. Copy the format, but it should be a 15x15 grid of `/turf/closed/indestructible/start_area` with `/turf/closed/indestructible/splashscreen` at the bottom left corner. In the center tile of this you need to place a `/obj/effect/landmark/start/new_player`
## Changelog
:cl:
fix: Splash screen inconsistency
remap: add splash screen to centcomm Z-level
/:cl:


cadyn:
> This require any extra effort to implement? or just good to go?

Nah, this is good to go.

PR #8277, opened 2024-04-21
[MIRROR] some borg upgrade overhaul, power use function, bluespace pounce and some fixes

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/15924
--------------------
The sight modules are mostly meant for events to give borgs different view options in a quick modify robot. They still need to equip the module though. So it's a selective option. (The anomalous one is similar to thermal for seeing invisible level 2 mobs, but without seeing through walls. Also anomalous sight gives borgs the bluespace pounce, including a glowy pounce icon..., higher range, leaping through objects at a higher power cost, instant pounce without miss chance. For downstream, it will dephase shadekin)

🆑 Upstream
add: adds some more upgrade modules to give different vision flag modules to borgs. currently admin only modules
add: vorespan defines to the chat spans
fix: icon for borg x-ray vision
fix: a rare case in which the store HUD icon could be overwritten by throw mode
fix: borgs accidentally draining their power cell fully with generate liquids
refactor: new function (use_direct_power(amount, lower_limit)) to directly use power for borgs with a lower power limit to prevent full self-discharge. Power is now used before the action.
code: ports some defines upstream, just to prevent compatibility issues in the future, so that no one uses them for something else
code: reworks borg upgrades again to clean them up some more
balance: robot shields now overload when the power cell has 200 charge remaining instead of fully depleting the robot
/🆑 
:cl:Raeschen
remove: Removes shadekinsight and 'bluespace pounce' from this PR (#8277)
/:cl:

Downstream: Don't merge right away. It needs some things added for us before merging.


Raeschen:
we're going to have to talk about the shadekin sight + the 'bluespace pounce'.

code/modules/mob/living/silicon/robot/inventory.dm
```diff
@@ -252,6 +254,32 @@
...
+			pounce.bluespace = TRUE
+	else
+		var/obj/item/weapon/dogborg/pounce/pounce = has_upgrade_module(/obj/item/weapon/dogborg/pounce)
+		if(pounce)
+			pounce.name = initial(pounce.name)
+			pounce.icon_state = initial(pounce.icon_state)
+			pounce.desc = initial(pounce.desc)
+			pounce.bluespace = initial(pounce.bluespace)
+	*/
+	var/obj/item/weapon/dogborg/pounce/pounce = has_upgrade_module(/obj/item/weapon/dogborg/pounce)
```
Kashargul, line 276: The section below is not needed if the one above is removef

PR #8325, opened 2024-04-28
One Life trait

FluffMedic:
I lWrite later
## About The Pull Request
Solely made this on my phone during a 10 minute break.
Wonder how horrid it went.
Cost is 4 currently. Update, changed to 2
Feels both too little and too much to me but ehh folks will proabaly holler.

## Changelog
:cl:
add: Added new mechanics or gameplay changes
add: Added a one life trait. Combination of no resleeve and no defib.
/:cl:

### DIFF
--- a/modular_chomp/code/modules/mob/living/carbon/human/species/station/traits/negative.dm
+++ b/modular_chomp/code/modules/mob/living/carbon/human/species/station/traits/negative.dm
@@ -86,7 +86,7 @@
 /datum/trait/negative/onelife
 	name = "One Life"
 	desc = "For whatever reason, once you dead, that is final."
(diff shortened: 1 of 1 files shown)


cadyn:
-4 is excessive. I would not consider the inability to be revived a significant mechanical disadvantage when players can either
A. not die (positive traits like adren rush, damage resist, etc. as well as abusing metas)
B. rejoin the round (whether as the same, or even just a different busted character)

-2 feels more reasonable to me. I don't think players should get 4 points for having their character get permanently removed from the round, it _should_ feel like the trait isn't worth it because god forbid permadeath becomes the meta

PR #8327, opened 2024-04-28
fuck everything (I tried to fix GC stuff again)

cadyn:
immense pain. I still don't know what causes ghosts to hard delete. All I know is that it's probably a sleeping proc somewhere, but we have a fucking lot of those.

oh also have the byond tracy dll I guess because I was using that one point
## Changelog
:cl:
fix: gc agony
/:cl:


ItsSelis (changes requested):
I'm not sure how good it would be to actually ship the prof.dll inside of the codebase, as that dll can also very easily (basically every time byond gets a new minor update) get outdated, meaning that someone would have to keep maintaining it so that issues are not being opened about receiving an error of the byond version being unsupported.

Issue #8332, opened 2024-04-30
[MAPPING] Sci-Med Maints not rad-proof (Area issue)


Missile597:
#### Brief description of the bug
Sci-Med maints do not have rad shielding or other maints properties (like vine prevention) due to being mapped in as a hallway. Should be an easy fix though... it's a small thing.

#### ~~What you expected to happen~~ How it currently is
Currently the whole maints area is classifies as part of the same sc-med hallway, therefore it's eligible for some events and not rad-proof
![image](https://github.com/CHOMPStation2/CHOMPStation2/assets/150307788/7677e775-ae6f-46e2-8646-b5e360c5769c)

#### ~~What actually happened~~ How to fix it
Ideally either need a newly defined maints area for this part of maints, or using an existing one that's not used elsewhere. Then redefine the maints area under this new area. Also, the hallway portion would need an APC as the area would be split in two (hallway and maints).

Missile597:
#8413 should have fixed this. Closing for now

PR #8336, opened 2024-05-01
2nd floor new maints is now Rad shield

FluffMedic:
## About The Pull Request
Added the rad shielding flag to the area encompassing the new 2nd floor maints
Not much else to note.

## Changelog
:cl:
fix: fixed rad shielding
/:cl:

### DIFF
--- a/maps/southern_cross/southern_cross_areas.dm
+++ b/maps/southern_cross/southern_cross_areas.dm
@@ -1176,9 +1176,9 @@ z
 	icon_state = "docking_hallway"
 
 /area/hallway/secondary/seconddeck/research_medical
-	name = "Research Medical Hallway"
+	name = "Research Medical Maintenance"
 	icon_state = "hallS"
-        flags = RAD_SHIELDED //chompEDIT: This place is maints now.
+	flags = RAD_SHIELDED //CHOMPedit: Adding rad proofing to new maints
 
 /area/medical/foyer
 	name = "\improper Medbay Foyer"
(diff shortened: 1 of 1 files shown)


Missile597:
Random suggestion:
Would it be better to assign the area to the maintenance category instead of just throwing rad shielding on it? It'll still spawn events this way after all as it's still a "hallway". Assigning it to maintenance would provide rad shielding and prevent event spawns by default.
Also, this is related to issue #8332

Missile597:
An alternate solution could be to just extend the maints area from the west through the corridor and into the newer maints without even needing to create a new area

maps/southern_cross/southern_cross_areas.dm
```diff
@@ -1176,8 +1176,9 @@ z
 	icon_state = "docking_hallway"
 
 /area/hallway/secondary/seconddeck/research_medical
-	name = "Research Medical Hallway"
+	name = "Research Medical Maintenance"
 	icon_state = "hallS"
+	flags = RAD_SHIELDED //CHOMPedit: Adding rad proofing to new maints
```
TheCaramelion, line 1181: Instead of adding rad proof, change the area in the map to "/area/maintenance/research_medical". This will prevent vines, viruses and any other kind of stuff to spawn in these areas, and make them rad proof

Missile597: This would work fine I reckon, but also don't forget to remove one of the APCs as well, as this would combine two areas into one leaving the combined area with duplicate APCs

FluffMedic: Oh, yeah mentioned this on discord but my strongDMM keeps crashing whenever I attempt to open anything with it.
Putting it here for easier relevant viewing

FluffMedic:
Fixed this in a diffrent PR that has been merged

PR #8364, opened 2024-05-08
Makes trashpiles climbable

FluffMedic:
## About The Pull Request
Changes density on trashpiles to false
Removed Herobrine

## Changelog
:cl:
qol: trash piles can be walked over
/:cl:

### DIFF
--- a/code/game/objects/structures/loot_piles_ch.dm
+++ b/code/game/objects/structures/loot_piles_ch.dm
@@ -110,14 +110,6 @@
 
 /obj/structure/loot_pile/maint/technical
 	density = FALSE
-	anchored = FALSE
 
 /obj/structure/loot_pile/maint/boxfort
 	density = FALSE
-	anchored = FALSE
-
-/obj/structure/loot_pile/maint/trash
-	anchored = TRUE
-
-/obj/structure/loot_pile/maint/junk
-	anchored = TRUE
--- a/code/game/objects/structures/loot_piles_ch.dm
+++ b/code/game/objects/structures/loot_piles_ch.dm
@@ -106,3 +106,18 @@
 /obj/structure/loot_pile/mecha/ripley/pirate
 	icon = 'modular_chomp/icons/mob/animal_ch.dmi'
 	icon_state = "pirate-broken"
+
+
+/obj/structure/loot_pile/maint/technical
+	density = FALSE
+	anchored = FALSE
+
+/obj/structure/loot_pile/maint/boxfort
+	density = FALSE
+	anchored = FALSE
+
+/obj/structure/loot_pile/maint/trash
+	anchored = TRUE
+
+/obj/structure/loot_pile/maint/junk
+	anchored = TRUE


Kazzc:
Eeehh. . . Climbing over them, sure!
Being able to move them, players can move and hoard them elsewhere. Good and bod in so many ways, I can see this being abused very easily.

cadyn:
I think trash piles being draggable is probably bad. We talked a bit in the dev channel and I would like to propose:
Instead of them being draggable, make them removable using a shovel after a progress bar of 10-15 seconds
For broken machines maybe a wrench/screwdriver instead

cadyn:
> I am not seeing the issue with making them draggable. Removing them would be nice but there are issues with folks wanting their maintenance loot.

Couple things:

- Makes it too easy to abuse. As somebody pointed out, it would be really easy for somebody to, on the premise of "cleaning" maints, just go and drag all of the lootpiles into an airlock and send them to the depths of space.
- Even if it was used as intended, there's definitely an argument to be made that it would just add even more clutter to the already cluttered maint tunnels if somebody needs to drag out a bunch of a loot piles likely just into the nearest tunnel.

FluffMedic:
Actually on 2nd thought, someone else can deal with the moveable/removable thing.
I don't think I will understand the other opinions around this.

FluffMedic:
> > I am not seeing the issue with making them draggable. Removing them would be nice but there are issues with folks wanting their maintenance loot.
> 
> Couple things:
> 
> * Makes it too easy to abuse. As somebody pointed out, it would be really easy for somebody to, on the premise of "cleaning" maints, just go and drag all of the lootpiles into an airlock and send them to the depths of space.
> * Even if it was used as intended, there's definitely an argument to be made that it would just add even more clutter to the already cluttered maint tunnels if somebody needs to drag out a bunch of a loot piles likely just into the nearest tunnel.

I feel someone would notice if someone was dumping random piles into space, and the sheer time that would take. The suggestion of.tool to remove it with 10-15 seconds sounds faster then that.

And for the second thing....cool, maints clutter is cool to me at least.

cadyn:
> > > I am not seeing the issue with making them draggable. Removing them would be nice but there are issues with folks wanting their maintenance loot.
> > 
> > 
> > Couple things:
> > 
> > * Makes it too easy to abuse. As somebody pointed out, it would be really easy for somebody to, on the premise of "cleaning" maints, just go and drag all of the lootpiles into an airlock and send them to the depths of space.
> > * Even if it was used as intended, there's definitely an argument to be made that it would just add even more clutter to the already cluttered maint tunnels if somebody needs to drag out a bunch of a loot piles likely just into the nearest tunnel.
> 
> I feel someone would notice if someone was dumping random piles into space, and the sheer time that would take. The suggestion of.tool to remove it with 10-15 seconds sounds faster then that.
> 
> And for the second thing....cool, maints clutter is cool to me at least.

As someone mentioned, the problem is specifically with them being draggable AND not having density, because at that point you can stack them on one tile and put them in like a closet or such, which would make them very easy to remove. I think though in general them being draggable just makes it a bit too low of barrier for somebody to be a nuisance with them imo.

cadyn (approved):
most of the discussion was around the draggable part, since it's just density being changed now, any more contentious changes can be made in a seperate PR and this one should absolutely merged.

PR #8369, opened 2024-05-10
[MIRROR] Fix for Digest Pain not loading preference

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/15971
--------------------
Fixed Digestion Pain preference not loading from saved preferences, causing it to reset every shift.


Kashargul:
Requires #8346

PR #8394, opened 2024-05-14
Tiny fan map removal

TheCaramelion:
## Changelog
:cl:
del: Removed a pair of tiny fans from under sheet storages
/:cl:

### DIFF
--- a/maps/southern_cross/southern_cross-2.dmm
+++ b/maps/southern_cross/southern_cross-2.dmm
@@ -57709,7 +57709,6 @@
 /area/hallway/secondary/docking_hallway2)
 "kar" = (
 /obj/machinery/door/firedoor/border_only,
-/obj/structure/fans/tiny,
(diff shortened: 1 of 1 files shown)


Skysteak:
What is the purpose of their removal? I believe their existence  originally it was to stop atmosphere from R&D/Cargo leaking out through the machine? Is that not the case then?

TheCaramelion:
> What is the purpose of their removal? I believe their existence  originally it was to stop atmosphere from R&D/Cargo leaking out through the machine? Is that not the case then? 

Thought they were a mistake; If that's their function I'll close this PR

Skysteak:
> > What is the purpose of their removal? I believe their existence  originally it was to stop atmosphere from R&D/Cargo leaking out through the machine? Is that not the case then?
> 
> Thought they were a mistake; If that's their function I'll close this PR

I am going off what I vaguely remember, so I am not 100% reliable, but I swear there were conversations about this prior!

PR #8405, opened 2024-05-15
Desert club sp pod

SlaysInCode:
## About The Pull Request

Sand enjoyers rejoice!

This addition to our list grants a purely social SP-Pod for Outsiders!
The pod is fitted with the following:

1 Bar
1 Kitchen
1 Botany Room
1 Storage Room
1 Bathroom
1 Dining Floor
1 Gaming Floor
1 Courtyard
2 Lockable Dorms

The pod is designed with a desert design in mind, made to look as though it was indeed placed there rather then built there.

It is size 33x33 so when marking out a spot to place it go to where you want any corner, then go in a 17x17. So if standing where you want the South-West corner to be, walk to the East 16 tiles (Not including the one you were standing on), then North 16 tiles (same here). I'm going to try making my pods uneven sizes like this so it'll be easy to spawn them.

## Changelog
:cl:
add: One SP-Pod.
/:cl:
image


Raeschen:
Wrong export format. You need to export as 'TGM'.

Also this strongly doesn't really fit the theme of pods on Sif.

SlaysInCode:
Will fix the format. Was told the theme would be okay, though will double check with them.

Raeschen:
> Will fix the format. Was told the theme would be okay, though will double check with them.

Well, we're hitting subjective opinion realm here. I find plonking a desert/sandstone building down in an ice planet's forest or plains a bit of an eyesore but i'll discuss with others to balance my viewpoint out

Raeschen:
Eh. on second thought you can disregard the 'theme' comment. Some others made a fair case for it.

SlaysInCode:
All good, I understand your concern so it's no hard feelings.

PR #8428, opened 2024-05-22
More Draketaur!

SilencedMP5A5:
adds a whole bunch of spotted variants for draketaurs (makkindorn took too long) and also gives draketaurs 3 tummy states!


SilencedMP5A5:
bro why is git clowning on me :cry:

PR #8438, opened 2024-05-23
Fix bad exec order in dme

Raeschen:
another PR made me aware of this. something fiddled with these files.


Raeschen:
this is the exec order that dream maker tries to enforce

PR #8499, opened 2024-06-08
Transfer Message Changes

Fracshun:
## About The Pull Request
Splits transfer messages, keeping the original for struggle transfers as they were intended and adding in new ones for auto-transfers. Disabling Entrance Logs will stop the pred from being spammed with these messages when auto-transfer happens, especially useful for belly loops. Also fixed bellies tasting prey entering them that aren't directly being eaten, like prey inside containers or being held by others.
## Changelog
:cl:
add: Separated messages for auto-transfers versus struggle transfers
qol: Disabling Entrance Logs also stops the pred from seeing auto-transfer messages
fix: Fixed tasting things not directly entering a belly, such as in boxes or held by others 
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/chompstation/VorePanel/VoreSelectedBellyDescriptions.tsx
+++ b/tgui/packages/tgui/interfaces/chompstation/VorePanel/VoreSelectedBellyDescriptions.tsx
@@ -408,6 +408,58 @@ const VoreSelectedBellyDescriptionsTransfer = (props) => {
           </Button>
         </>
       )}
+      {(message_mode ||
+        !!interacts.transferlocation ||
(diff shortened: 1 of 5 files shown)


Kashargul (changes requested):
Missing changes in the export . tsx

tgui/packages/tgui/interfaces/chompstation/VorePanel/VoreSelectedBellyDescriptions.tsx
Kashargul, line 1: Logic should be checked that only the suitable messages are shown

Kashargul:
Files are not typscript conform.

Kashargul (approved):

PR #8508, opened 2024-06-09
implant drop

Kashargul:
## Changelog
:cl:
balance: Shadekin now drop all shrapnels, NIFs and most implants on phase shifting. Size and Compliance implants are whitelisted from this.
/:cl:

### DIFF
--- a/modular_chomp/code/modules/mob/living/carbon/human/species/shadekin/shadekin_abilities.dm
+++ b/modular_chomp/code/modules/mob/living/carbon/human/species/shadekin/shadekin_abilities.dm
@@ -211,9 +211,13 @@
 		canmove = FALSE
 
 		//CHOMPAdd Start
+		var/list/allowed_implants = list(
+			/obj/item/weapon/implant/sizecontrol,
+			/obj/item/weapon/implant/compliance,
+		)
 		for(var/obj/item/organ/external/organ in organs)
 			for(var/obj/item/O in organ.implants)
-				if(istype(O, /obj/item/weapon/implant))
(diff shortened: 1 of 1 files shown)


Missile597:
Okay, there's obviously been a disagreement on this but I acknowledge that staff have decided NIF's shouldn't be compatible with shadekin. Fair enough. Though I would argue this seems like a lazy way to automoderate shadekin with existing NIFs (why not just make the NIFs incompatible like normal?), this is one way of solving it I suppose.

What I do want to ask though is this: I am okay with this being merged but *please*, can we have an alternative added for rp tools such as soulcatchers added *before* we merge this. It's no secret this was made purely because people rioted over one shadekin with a crew monitor. While I'm okay with shadekin having NIFs removed as they're usually such a small thing, it's mainly the loss of the soul catcher that bothers most of us. The soul catcher is a fun vore scene tool and currently it's still locked behind requiring a NIF to use, and taking NIFs away from shadekin locks shadekin out of this kind of roleplay for good, all because one shadekin broke the rules and nothing was done about it.

TLDR - Please can we hold off merging it until soulcatchers and other scene tools are available? Maybe make is a shadekin only neutral trait or something, or make a shadekin specific NIF alternative.

Kashargul:
> TLDR - Please can we hold off merging it until soulcatchers and other scene tools are available? Maybe make is a shadekin only neutral trait or something, or make a shadekin specific NIF alternative.

We'll not keep off rule enforcements, bug fixes or exploit patching due to other features not yet being implemented.

PR #8519, opened 2024-06-13
Fixes bad runtime spam whenever anyone without a nif cryos out

Verkister:
hmngh what nif
## About The Pull Request
Fixes bad runtime spam whenever anyone without a nif cryos out
## Changelog
:cl:
fix: fixed runtime spam when cryoing without a nif.
/:cl:


Kashargul:
Shouldn't be set like this, it can cause issues. The correct one would be nif &&

PR #8526, opened 2024-06-13
vore soulcatcher

Kashargul:
## About The Pull Request
This is an early test release. Any bugs, right below during test phase. Details as follows:

Adds a new item, soulgem to all mobs. The preferences are stored similar to the vore organs.

If enabled, this item allows to capture the souls of prey on death. Self catching has not yet been tested. So would be good if someone would test it!

Inside the vore panel, there's now a new tab, the Soulcatcher. This tab allows you to interact with the caught souls.

It should be possible to capture all kind of mobs... even borgs.

A single belly overlay can be selected to be displayed within the soulcatcher!

The owner can enter and exit the soulcatcher at any time themselves!

Upon releasing a soul, they will be ghosted, so be sure that you want to release them!

Upon transferring, you can transfer them into a sleevemate in your hands, for borgs in their active module, or into a held mmi. To make use easier, you also can just stand on the item in the same time. The item must not have a mind loaded though before transfer!

Upon deletion, you will round remove a prey, make sure that you got OOC consent! I might add a confirmation prompt for the prey later... but for now I want feedback on functionality. This has it's own preference and it is clearly marked as dangerous if enabled!

Due to how the capturing works, borgs will have their current in round name transferred to the brainmob. This was the only way not to have their character slot shown.

In the vore panel, there's a new preference section now which will show the soulcatcher prefs. Allow capturing, as well as allow transferring and deleting. Those three are all disabled by default and need to be enabled explicitly.

grafik
grafik

_____________________________________________
grafik
grafik
grafik

## Changelog
:cl:
add: vore soulcatcher
add: experimental resize, costing 125 Nutrition
code: changes all our UI files to typescript
/:cl:


Kashargul:
Open points to check:
- [ ] can't see / interact with the AR projection without a nif / not at all?
- [ ] add tooltips to the buttons
- [ ] Critical: can't catch the mind back out after self capture, leading to softlocks


Sometime if feeling like it:
- [ ] allow custom vorefx

PR #8531, opened 2024-06-16
fixes phantom blood stains

tacoguy7765093:
## About The Pull Request
If you at any point stain yourself or an object with blood, the first color of blood will always be what color the bloodstain is no matter what. This fixes that, thank god everything is in one proc. This should fix soap, hand washing, showers, etc. Did surface level testing and found no runtimes.
## Changelog
:cl:
fix: fixes blood cleaning on objects.
/:cl:

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -493,6 +493,7 @@
 	if(istype(blood_DNA, /list))
 		blood_DNA = null
 		return TRUE
+	blood_color = null //chompfixy, cleaning objects saved its future blood color no matter what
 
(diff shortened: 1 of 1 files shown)


code/game/atoms.dm
```diff
@@ -493,6 +493,7 @@
 	if(istype(blood_DNA, /list))
 		blood_DNA = null
 		return TRUE
+	blood_color = null //chompfixy, cleaning objects saved its future blood color no matter what
```
Kashargul, line 496: Should be tested with the robolimb datums. They all have custom blood_color values defined.

Kashargul: Should be fine, they are independent.

PR #8569, opened 2024-06-23
New gateway: Skyscraper.

tacoguy7765093:
## About The Pull Request
In a gateway far away, a rundown skyscraper with many of their inhabitants disappearing without a trace, forcing infestation, piracy, and whatever evil shit you can think of.

Adding a new gateway complete with 11 "multi-z" floors with their own theming on each floor. There is a bit of mobs but I tried to keep it a bit on the down low as the focus was more around the maze like structure and getting lost/falling to your death. However this is still not a solo gateway. The gateway was completed with 2 people with difficulty with minimal prep. Group of 3 should have no problem if prepared.

Also contains shuttle landmarks if placed in the overmap. 

Also falling is very deadly. Dont slip!

dreamseeker_nwq56c9wuK
dreamseeker_RVLXbgwWrP
dreamseeker_jtJzYJIDCO
dreamseeker_Aqo3JTg2lL
## Changelog
:cl:
add: Added a new gateway map called Diescraper. Dont fall!
add: Added single use for mapping purposes.
maptweak: Fixes slab IDs in Downtown city
/:cl:


tacoguy7765093:
Strange, shuttle shouldnt exist when the gateway doesnt get loaded

tacoguy7765093:
Its trying to place the shuttle in a non existent location if the map doesn't load... but the file is only included when the map in included so im not sure where this is coming from.

tacoguy7765093:
OK, I need help figuring out how to only include diescraper.dm if it is selected for load. It is loaded in via southern_cross_submaps

JohannaMEriksson:
> OK, I need help figuring out how to only include diescraper.dm if it is selected for load. It is loaded in via southern_cross_submaps

comment out the other maps in that file

tacoguy7765093:
The list of what maps are loaded are shown in southern_cross_defines line 135, and without the files included, the areas for all the gateways wont load period.

JohannaMEriksson:
sorry, i believe i misunderstood.
if i am not mistaken, all of the gateway maps' defines are loaded regardless of whether or not the map is loaded. you'd have to either a) separate the elevator's definition from the map definition file and find a way to only load that file if the map is selected, or b) create a holding area for the elevator, possibly on the centcom z-level, and find a way to transport it after loading the map. just two ideas from me. i'm a mapper, not so much a coder, so there's not much more help i can provide. sorry.

tacoguy7765093:
I will likely get around to making a hacky (temp) fix ((actually just the final method)) and just fuck around with teleporters because this is way beyond my coding capabilities.

tacoguy7765093:
I give up, fuck shuttle code, you get qpad

Kazzc:
Sucks that the elevator code is fucked, funny as it was to see it working. Impressive map, may snag it one day and see if I can a event on this thang.

PR #8576, opened 2024-06-26
Footstep element

TheCaramelion:
## About The Pull Request

Fresh and warm, straight from TG. Ports the Footstep element and makes lists in turfs a thing of the past. We welcome our optimization overlords.

Drafting because I've yet to make the teshi sounds and other species with claws, and possibly a bunch of simple mobs.

## Changelog
:cl:
add: Footstep and slosh elements
code: No more lists in every single turf in existance with footstep sounds
/:cl:


code/game/turfs/simulated/floor.dm
```diff
@@ -51,8 +51,8 @@
 		set_flooring(get_flooring_data(floortype), TRUE)
 		. = INITIALIZE_HINT_LATELOAD // We'll update our icons after everyone is ready
 	else
```
Kashargul, line 53: If you remove an else case, make sure to remove the entire section and don't leave it standing alone

TheCaramelion: I may be dumb

PR #8676, opened 2024-07-21
[MIRROR] small mistake in the UIs, noticed due to conflict

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/16122
--------------------
The reading value is either TRUE or FALSE, in case there is a sensor, then we have a TRUE value... it is never -1 and currently we would always show the error...

🆑 Upstream
fix: an error in the shuttle UI preventing sensors from being shown when they should.
/🆑


Kashargul:
part of #8669

PR #8695, opened 2024-07-26
Custom Footstep sounds

TheCaramelion:
## About The Pull Request

Allows players to choose what footstep their characters make. 

## Changelog
:cl:
add: Added an option for players to choose their footstep sound
qol: Reduced footstep volume
code: Less procs, more signals
/:cl:


TheCaramelion:
Works on my machine

Kashargul:
> Works on my machine

IMG_0753

code/modules/vore/eating/living_vr.dm
```diff
@@ -88,6 +88,8 @@
 		if(!soulgem)
 			soulgem = new(src)
 		return TRUE
+
+	AddElement(/datum/element/slosh)
```
Kashargul, line 92: // Make sure you don't call ..() on this one, otherwise you duplicate work.
/mob/living/simple_mob/init_vore()

simple mob has it's own init vore

Kashargul: Also, you are adding the element AFTER the returns, it neds to go to the TOP.

Kashargul:
Seems we still have calls where it gets readded, is there a way to prevent that, or shall we use override TRUE?

PR #8704, opened 2024-07-28
New weather: Extreme Monsoon

tacoguy7765093:
## About The Pull Request
Literally the fatal rain in Rainworld has been added, currently its adminspawn only and nothing short of 100% armor can actually save you from this rain.
Starts with a 1-2 minute warning complete with rumbling sounds telling you to get to shelter, before 1 minute of the downpour starts, only knocking you down. You better crawl away in time before the real fatal rain comes in and damages _every living thing_ including mobs that are caught outside.

The rain does not damage you when dead, so no 6000 brute corpses.

Oh yeah, and also fixes defines for the other new weathers that were added giving them real names

https://github.com/user-attachments/assets/5e2a796a-9a78-48e0-a611-9e5389d862d7

## Changelog

:cl:
add: Added a new extreme weather for admins called "extreme monsoon"
fix: Weather defines updated
/:cl:


Kashargul (commented):
when you can, use span_danger(), span_notice() etc instead of e.g. <span class="danger"></span>

Kashargul (approved):

Kashargul:
Question, does it damage non player mobs?

PR #8708, opened 2024-07-30
Step Element runtime fix

TheCaramelion:
## About The Pull Request

Players won't notice a thing, probably. Hello.

## Changelog
:cl:
fix: Fixes a small oversight that made foostep element throw constant runtimes
/:cl:

### DIFF
--- a/modular_chomp/code/datums/elements/footstep.dm
+++ b/modular_chomp/code/datums/elements/footstep.dm
@@ -43,8 +43,6 @@
 			footstep_ret = GLOB.lightclawfootstep
 		if(FOOTSTEP_MOB_CLAW)
 			footstep_ret = GLOB.clawfootstep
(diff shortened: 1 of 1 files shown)


modular_chomp/code/datums/elements/footstep.dm
```diff
@@ -53,6 +51,8 @@
 			footstep_ret = 'modular_chomp/sound/effects/footstep/slime1.ogg'
 		if(FOOTSTEP_MOB_SLITHER)
 			footstep_ret = 'modular_chomp/sound/effects/footstep/crawl1.ogg'
+		else
```
Kashargul, line 54: This seems dangerous. There should be a return after each successful set

Kashargul: Switch-case default state

PR #8713, opened 2024-07-31
reduce nutrition gain from trash on sleepers

### LINKED ISSUE #8709: verb to disable or reduce nutrition gain on self clean function for janitor borgs.✨
#### Brief description of the feature
Now that janitor cyborgs can use the self clean to get glass to refill their light replacer. I have discovered that using the self clean feature on trash leaves you constantly slowed by nutrition. But I also know the slow is to deal with borg powergaming. but as someone who is literally just trying to clean, this blows.

#### What you want to happen
 Janitor cyborgs can opt out of nutrition so they're not crippled when getting material from garbage. EXCLUSIVELY for /obj/item/trash, to prevent the destruction of all maint items. and EXCLUSIVELY in the trash compactor belly, not vore bellies. 
 
 In summary, /obj/item/trash in the Garbage processor borg belly, not supplying nutrition at all.

- (Suggestion just for janitor cyborgs because self clean is essential to their updated kit.)

Kashargul:
## Changelog
:cl:
qol: reduces nutrition gain of trash for sleepers from 5 to 1, there's enough ways to get nutrition and some units, like janitors can collect a lot of trash into their sleeper
/:cl:

closes #8709

### DIFF
--- a/code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
+++ b/code/modules/mob/living/silicon/robot/dogborg/dog_sleeper_vr.dm
@@ -664,6 +664,7 @@
 				//CHOMPAdd Start
 				if(T.reagents)
 					volume = T.reagents.total_volume
+				var/is_trash = istype(T, /obj/item/trash)
 				//CHOMPAdd End
 				var/digested = T.digest_act(item_storage = src)
 				if(!digested)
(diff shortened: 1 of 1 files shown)


EtheoBoxxman:
I don't think this actually solves the problem that was there before, because the problem in the linked issue is that gaining nutrition at all, especially constantly gaining it, causes you to be in a perpetual state of unnecessary slowdown.

Kashargul:
> I don't think this actually solves the problem that was there before, because the problem in the linked issue is that gaining nutrition at all, especially constantly gaining it, causes you to be in a perpetual state of unnecessary slowdown.

1000 items is very very unlikely. Also there's a purge nutrition verb to drop some nutrition.

PR #8731, opened 2024-08-03
Vorespawn Whitelist and Absorb

Fracshun:
## About The Pull Request
Adds two new belly options for vorespawns. A whitelist to only show the belly on vorespawn for particular ckeys, and a preference on whether that belly absorbs preyspawns or not.

The whitelist allows hiding of certain bellies so only those the pred would want spawning in them can attempt to. This should assist in more graphic belly names not being so easily visible but still vore spawnable for continued scenes.

The vorespawn absorb is a belly choice with three options. Defaulted to No, where all prey spawns will just be regular vore. The Yes option will make prey spawning there absorbed, but the prey get a warning prompt saying this will happen before the request is sent to the pred. The Prey Choice option lets the prey choose between being absorbed to start or not, and if the prey chooses to be absorbed the pred will see this in the notification they receive before accepting.

In addition, the vorespawn pred now has an additional option to instantly absorb those they spawn on. This is articulated to the prey receiving the request. It doesn't require any belly specific options set for either party.
## Changelog
:cl:
add: Two new belly options for vorespawn. Whitelisting ckeys and whether the belly absorbs prey on spawn.
add: Added a prompt for vorespawn preds to instantly absorb their chosen prey.
/:cl:

### DIFF
--- a/code/__defines/belly_modes_ch.dm
+++ b/code/__defines/belly_modes_ch.dm
@@ -43,4 +43,4 @@
 
 //Vorespawn flags
 #define VS_FLAG_ABSORB_YES		0x1
-#define VS_FLAG_ABSORB_CHOICE	0x2
+#define VS_FLAG_ABSORB_PREY		0x2
--- a/code/__defines/belly_modes_ch.dm
+++ b/code/__defines/belly_modes_ch.dm
@@ -40,3 +40,7 @@
 #define AT_FLAG_ORES			0x40
 #define AT_FLAG_CLOTHES			0x80
 #define AT_FLAG_FOOD			0x100
+
+//Vorespawn flags
+#define VS_FLAG_ABSORB_YES		0x1
+#define VS_FLAG_ABSORB_CHOICE	0x2
--- a/code/game/jobs/job_controller.dm
+++ b/code/game/jobs/job_controller.dm
@@ -673,6 +673,7 @@ var/global/datum/controller/occupations/job_master
 	var/datum/spawnpoint/spawnpos
 	var/fail_deadly = FALSE
 	var/obj/belly/vore_spawn_gut
+	var/absorb_choice = "No" //CHOMPAdd - Ability to start absorbed with vorespawn
 	var/mob/living/prey_to_nomph
 	var/obj/item/item_to_be //CHOMPEdit - Item TF spawning
 	var/mob/living/item_carrier //CHOMPEdit - Capture crystal spawning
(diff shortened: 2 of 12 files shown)


code/game/jobs/job_controller.dm
```diff
@@ -673,6 +673,7 @@ var/global/datum/controller/occupations/job_master
 	var/datum/spawnpoint/spawnpos
 	var/fail_deadly = FALSE
 	var/obj/belly/vore_spawn_gut
+	var/absorb_choice = "No" //CHOMPAdd - Ability to start absorbed with vorespawn
```
Kashargul, line 676: Please don't use strings for numeric choices! Display the texts UI side!

use defines to set names in code.
#define SPAWN_ABSORBED 1


In this case, 3 Flags would even be preferred, flag Math exists on the soulcatcher to show the implementation for the UI with multiple states.

Kashargul (changes requested):

code/__defines/belly_modes_ch.dm
```diff
@@ -40,3 +40,7 @@
 #define AT_FLAG_ORES			0x40
 #define AT_FLAG_CLOTHES			0x80
 #define AT_FLAG_FOOD			0x100
+
+//Vorespawn flags
+#define VS_FLAG_ABSORB_YES		0x1
```
Kashargul, line 45: Use the defined flags, don't hardcode numbers later on.

0 is not a valid flag as it prevents further usages

code/modules/vore/eating/belly_obj_ch.dm
```diff
@@ -55,6 +55,9 @@
 	var/liquid_fullness4_messages = FALSE
 	var/liquid_fullness5_messages = FALSE
 	var/vorespawn_blacklist = FALSE
+	var/vorespawn_whitelist = list()
+	var/vorespawn_absorbed = 0
+	var/tmp/static/list/vorespawn_absorbed_flags_list = list("No" = 0, "Yes" = VS_FLAG_ABSORB_YES, "Prey Choice" = VS_FLAG_ABSORB_CHOICE)
```
Kashargul, line 60: Why do we have this list???

Kashargul:
In a valid flag arithmetic, we could reduce this to merely 2 states:

VS_FLAG_ABSORB_NO 0x1
VS_FLAG_ABSORB_YES 0x2

the third state could be derived by having yes and no enabled at once. Although, that would require very strict comparisons and it's not worth the effort. We'd have to first check if both flags are set, see at the very bottom, then do x otherwise, check each flag singularly to do y, z.

So you can cut it down to using 3 bits, having a lot easier thinking for the arithmetic later.

VS_FLAG_ABSORB_NO 0x1
VS_FLAG_ABSORB_YES 0x2
VS_FLAG_ABSORB_PREY 0x4

now you just need to set / unset the bits when you call your function

to set a bit, flag |= VS_FLAG_ABSORB_NO 
to unset a bit flag &= ~(VS_FLAG_ABSORB_YES)

you'd then only ever use 3 bits of the 32 available bits, leaving the rest for other things for the future.

with a number, you'd use all 32 bits.

to later on check flags, the following is the easiest way (flag & VS_FLAG_ABSORB_NO) 

if you have multiple flags you want to check at once, the following is valid

(flag & (PARAM_1 | PARAM_2)) == (PARAM_1 | PARAM_2)

Kashargul (approved):

PR #8771, opened 2024-08-11
[MIRROR] TGUI Say

CHOMPStation2:
Original PR: https://github.com/VOREStation/VOREStation/pull/16160
--------------------
This adds the say modal from /tg/, affectionately labeled "tgui say". It is on by default, turning it off will result in going back to the way it was. 

The main features:
- Allows switching between different modes with tab/shift.
- It's pretty.
- Typing indicators have a new mode: thinking. This is automatically switched to when you are not actively typing.
- Command bar spying - Typing indicators will now come up (if turned on as normal) for typing in the command bar at the bottom too, only with relevant verbs.
- Shows you what radio channel you're going to speak into ahead of time.

Additional features that I added just for here:
- Shift-enter lets you do multiline input (later stripped for anything other than say and me)
- Shift-tab lets you cycle backwards through the modes.

:cl:
add: Advanced TGUI Say modal
tweak: Verb panel say/me/whisper/subtle have been renamed to "Say verb"/etc - and pop open the tgui say panel/tgui input/normal input instead of always using the native dialog.
imageadd: Every speech bubble now has an extra spinning/"thinking" state.
tweak: Added /tg/'s trait system, currently just used for indicating that you're in the typing state.
fix: TGUI dev server works again (keep_local_name = TRUE is really important, both for that and reliability)
tweak: Subtle/Whisper now has typing indicators by default - you can turn indicators off for these channels specifically with a new pref.
tweak: Typing indicators switched fully to /tg/'s system for them with traits and client integration and... stuff.

/:cl:


Kashargul:
Careful, verbs need adjustments for TGPanel

Verkister:
Can we make this not enabled by default? It's pretty awful and doesn't support multiple inputs or moving them out of the way. People who are gonna be getting driven away by the clunkiness don't tend to be the type to figure out the char setup global buttons on their own.

Kashargul:
> Can we make this not enabled by default? It's pretty awful and doesn't even support multiple inputs or moving them out of the way. As in you cannot even reply to LOOC while in the middle of typing a scene wallpost. People who are gonna be getting driven away by clunky unintuitive systems don't tend to be the types to figure out the char setup global buttons on their own.

we culd simply have the hidden verbs shown alongside aka the old inputs

Kashargul:
> Can we make this not enabled by default? It's pretty awful and doesn't even support multiple inputs or moving them out of the way. As in you cannot even reply to LOOC while in the middle of typing a scene wallpost. People who are gonna be getting driven away by clunky unintuitive systems don't tend to be the types to figure out the char setup global buttons on their own.

-> I did some tests, LOOC works in parallel without issues.
-> You can move tgui say.

Verkister:
There's nothing for the mouse to grab on tho?

Should at least move the toggle pref buttons from the bottom of the tall grass to the top where the other tgui buttons are.

Kashargul:
> There's nothing for the mouse to grab on tho?
> 
> Should at least move the global tab toggle pref buttons from the bottom of the tall grass to the top where the other tgui buttons are.

Just click at the edge? it works fine for me.

CEO-Of-Dogborg-Corp:
This broke chatting normally using the Say " Message " command in the command bar at the bottom.

PR #8806, opened 2024-08-20
fixes goia grabs

tacoguy7765093:
## About The Pull Request
~~This tame code doesn't even work, commenting out for now. Tamed subtype is still chill until grabbed.~~
Fixed the code so goia can be properly grabbed if controlled by a player.
## Changelog
:cl:
fix: Fixed a oversight that prevents zorgoia from being grabbed.
/:cl:


Kashargul:
Why doesn't it work? What it does instead?

You mean you can't grab player controlled mobs? If so, well it returns in every case where HP > 0 and it doesn't have an AI.

PR #8808, opened 2024-08-20
AnomalousWeatherUpdate 2.1

FluffMedic:
## About The Pull Request

Currently adjusted stats due to rescent combat alterations
And reworks the patterns and functions of one boss to give more telegraphs.
Future bugfixes, balance changes, etc may be slapped onto this PR, or in future ones depending on course of events.
## Changelog
:cl:
fix: Weather beasts have their stats adjusted 
qol: The heavy fog weather beast has more of a tell to their attacks.
/:cl:


modular_chomp/code/modules/mob/living/simple_mob/subtypes/vore/weather.dm
```diff
@@ -72,90 +201,48 @@
...
-		adjustToxLoss(-40)
-		adjustOxyLoss(-40)
-		adjustCloneLoss(-40)
-		sleep(1)
-		adjustBruteLoss(-40)
-		adjustFireLoss(-40)
-		adjustToxLoss(-40)
-		adjustOxyLoss(-40)
-		adjustCloneLoss(-40)
+		sleep(30)
```
Kashargul, line 204: why isn't this in a loop?

Kashargul (changes requested):

modular_chomp/code/modules/mob/living/simple_mob/subtypes/vore/weather.dm
FluffMedic: Unsure how to create one, and what purpose that would serve of it being in a loop

FluffMedic: Guessing it's length of code. Altered it to just have a longer cooldown instead of repeated healing overtime alongside larger healing on the outset.
I do apologize if the previous comment came off as harsh, meant to be a question mark

Kashargul: I'll push an example, it's usually better to loop code than to copy paste it:

for(var/i =1 to 6)
    do_something()

the loop will now recall the same function. So instead of copypasting the code very often, you just increase the to value to have it loop more often

Kashargul: For your code it would be something like

```
for(var/i = 1 to 8)
	adjustBruteLoss(-80)
	adjustFireLoss(-80)
	sleep(10)
adjustBruteLoss(-80)
adjustFireLoss(-80)
```

We don't want to sleep after the final proc again

FluffMedic: Oh that is oddly simple. Thank you. Altered the tank code to utilize this, and for another example of it later on.

PR #8809, opened 2024-08-21
Departamental Lathes

TheCaramelion:
Adds Departamental Lathes for every single department.
These departamental lathes aren't as strong as the one in R&D, only allowing to print selected job-related items. They're also not as efficient, making R&D still essential.

Like the R&D lathe, these ones can be upgraded as well, making them print items faster and consume less materials, so you'll have to either ask Science or Engineering for help to upgrade these.
## Changelog
:cl:
add: Departamental Lathes for each department
/:cl:


code/__defines/research.dm
```diff
@@ -18,3 +18,12 @@
 #define MECHFAB		0x0004	//Mechfab
 #define CHASSIS		0x0008	//For protolathe, but differently
 #define PROSFAB		0x0010  //For prosthetics fab
+
+// CHOMPAdd - Departmental Lathes
+#define LATHE_ALL 				0x0001
+#define LATHE_SCIENCE			0x0002
+#define LATHE_ENGINEERING		0x0004
+#define LATHE_MEDICAL			0x0008
+#define LATHE_CARGO				0x0016
```
Kashargul, line 27: This will blow up

PR #8823, opened 2024-08-23
Lesser Lightweight

FluffMedic:
title
## About The Pull Request
Addition: lesser lightweight
Guaranteed knock down but only if you're on Help Intent
Only -1.

## Changelog
:cl:
add: Lesser lightweight
/:cl:

### DIFF
--- a/code/modules/mob/living/living_movement.dm
+++ b/code/modules/mob/living/living_movement.dm
@@ -148,6 +148,12 @@ default behaviour is:
 				H.Weaken(5)
 				now_pushing = 0
 				return
+			//CHOMPSTATION edit Adding alternative to lightweight
+			if(H.species.lightweight == 1 && H.a_intent == I_HELP)
+				H.visible_message("<span class='warning'>[src] bumps into [H], knocking them off balance!</span>")
+				H.Weaken(5)
+				now_pushing = 0
+				return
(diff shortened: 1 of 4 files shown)


modular_chomp/code/modules/mob/living/carbon/human/species/station/traits/negative.dm
```diff
@@ -90,3 +88,10 @@
 	custom_only = TRUE
 	var_changes = list("flags" = NO_SCAN, NO_DEFIB)
 	excludes = list(/datum/trait/negative/nodefib, /datum/trait/negative/noresleeve)
+
+/datum/trait/negative/lightweight_light
```
Kashargul, line 92: shouldn't this exclude normal lightweight?

code/modules/mob/living/living_movement.dm
```diff
@@ -148,6 +148,12 @@ default behaviour is:
 				H.Weaken(5)
 				now_pushing = 0
 				return
+			//CHOMPSTATION edit Adding alternative to lightweight
+			if(H.species.lightweight == 1 && H.a_intent == I_HELP)
```
Kashargul, line 152: Umm, is this intended to check the same var as the normal leightweight?

Kashargul:
Make sure that the exclusion works in both ways if it's not added to each other's exclusion list

PR #8842, opened 2024-08-27
rebalance melee explo borg

Kashargul:
## About The Pull Request
With most bullets dealing a much higher damage now, the explo borg should not be left with less than pistol damage as a melee only class. Also the shield should be raised to be at least partially useful against the higher bullet damage.
## Changelog
:cl:
balance: energy shield life 35 -> 90
balance: energy shield regen 10 -> 25
balance: combat spear damage 25 -> 45 damage
balance: explo jaws: 40 -> 60 damage
/:cl:


Kazzc:
Most mobs have between 50-150 hp.
45-60 damage will be killing most things between 2-3 hits, small mobs in single hit.
Knowing how fast borgs/spears can hit. Killing mobs in 1-3 seconds.

Unsure about such a strong damage buff, a station spider with 200 hp will die in 1-3 seconds.

PR #8873, opened 2024-08-31
Food Replicator

TheCaramelion:
## About The Pull Request

Adds a food replicator machine, along with the circuit to build it. It takes a big chunk of nutriment if not upgraded, and it is also rather slow and power hungry.

To unlock the food, first it **must** be cooked, giving the chef a chance to cook anything they want and then for the crew to replicate it if they want more and the chef is not around.

Science and Engineering can both print this, as long as the minimum research is met.

## Changelog
:cl:
add: Food replicator machine
/:cl:


Kashargul (changes requested):
There should be a blacklist for some specific food items...

-> food with supermatter reagent

modular_chomp/code/game/machinery/food_replicator.dm
```diff
@@ -82,6 +87,9 @@
 			printing = TRUE
 			update_icon()
 
+			if(product_path == /obj/item/weapon/reagent_containers/food/snacks/donkpocket/ascended)
```
Kashargul, line 90: could even check if any food contains supermatter as reagent, so in case anyone adds some new funny food it's handled the same

Kashargul (approved):
Hmmm, so the donks even need to be heated to be replicable?

TheCaramelion:
> Hmmm, so the donks even need to be heated to be replicable?

Mostly because of the ascended one, else it would be easy to just replicate it all over

PR #8947, opened 2024-09-11
explo borg card access

Kashargul:
## Changelog
:cl:
fix: explo borg Id not forwarding access
/:cl:


Kashargul:
Will add a proper rank assignment to the card when doing the deconflict of the mirror

PR #8948, opened 2024-09-13
Reverts the spacebar toggling R key

FluffMedic:
## About The Pull Request
R key already toggles throw, and pressing it accidentally is something rather easy to do.
Space bar is bigger, and utilized further for typing.
Also several folks used it for Macros before.

## Changelog
:cl:
qol: spacebar no longers triggers throw
/:cl:


Skysteak:
Do note that the  original concept of this PR is to have any keyboard macro set to spacebar override this function. This works as designed on Virgo, but not on Chomp.

FluffMedic:
Okay, so. /* chomp edits don't work in the .dmf files.
Probably should bring up discussion of if it's worth properly deleting it here, or just accepting the existence of it.

PR #9001, opened 2024-09-21
[DO NOT MERGE] Temporal Z-Level disable for events

TheCaramelion:
## About The Pull Request

This PR is just temporal, for Test Merging purposes so we can free some memory while we do events and the server won't explode.

No changelog at, this is not supposed to be merged.


Kashargul:
Won't work. We are using a fully new map folder

PR #9041, opened 2024-09-25
Map Update Fuel depo and VR.

Kazzc:
Maps and its areas are outdated on Southern Sun files. This should fix it.


tacoguy7765093:
Duplicate file path in southern_cross_submaps line 97, VR world is overridden by fuel Depot

PR #9076, opened 2024-09-28
updating chest transparency toggle to be head instead. better for slime synths.

Asher-42:
## Changelog
:cl:
qol: swapped chest transparency toggle for proteans to make everything but their head stransparent to allow for a metal shell.
/:cl:


Kashargul:
What exactly is the difference? Got screenshots from before and after?

Asher-42:
> What exactly is the difference? Got screenshots from before and after?

before only the chest would go transparent for sub protean species idea, changing it now to keep the head normal and all else is made transparent, i can make a photo later. 

i just dont want to add a million buttons

PR #9100, opened 2024-09-29
Makes VR and fuel depot modular, changes SC/SS defines

tacoguy7765093:
## About The Pull Request
Modularizes the VR map and the fuel depot so there doesn't have to be 2 separate map files that have to be touched between the 2 maps.
## Changelog
:cl:
code: Modularizes the VR map.
maptweak: VR map is latest.
/:cl:


Kazzc:
Thanks for the fix!

PR #9198, opened 2024-10-11
Reduces the amount of guns in PoIs, adds salvage items

TheCaramelion:
## About The Pull Request

Okay this was kinda big-

Reduces the amount of guns and items in exploration PoIs, most if not all of these items were moved to mob drops instead. Once a mob die, they'll drop almost the exact same loot that was laying around all the time. Guns, items, devices.
Additionally, adds salvage items that can be sold to cargo for a good chunk of points. These will drop from some enemies and guaranteed for some of the strongest ones.

## Changelog
:cl:
add: Added "salvage items" that can be found with mobs or in specific PoIs
balance: Reduced the amount of guns and weapons found in exploration Pois
/:cl:


Kazzc:
Sure! exploration needs more work but this a good step.
Hopefully be a lot less gunwhore.

PR #9202, opened 2024-10-12
Fix the size implant having it's own clamping.

TheGreatKitsune:
## About The Pull Request
Removes the clamping size implants did on their own, allowing larger/smaller sizes to work when in spaces like the dorms that support it.
## Changelog
:cl: AzulKitsune
fix: fixed the size implant not supporting larger/smaller sizes in dorms.
:cl:

### DIFF
--- a/code/game/objects/items/weapons/implants/implant_vr.dm
+++ b/code/game/objects/items/weapons/implants/implant_vr.dm
@@ -104,7 +104,7 @@
 				var/static/regex/size_mult = new/regex("\\d+")
 				if(size_mult.Find(msg))
 					var/resizing_value = text2num(size_mult.match)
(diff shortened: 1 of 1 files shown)


Kashargul:
Are you sure that works without the second variable?

code/game/objects/items/weapons/implants/implant_vr.dm
```diff
@@ -104,7 +104,7 @@
 				var/static/regex/size_mult = new/regex("\\d+")
 				if(size_mult.Find(msg))
 					var/resizing_value = text2num(size_mult.match)
-					H.resize(CLAMP(resizing_value/100 , 0.25, 2))
+					H.resize(CLAMP(resizing_value/100 , 0.01, 6), uncapped = H.has_large_resize_bounds()) //CHOMPEdit - Let resize handle size limits. It's meant to do that.
```
Kashargul, line 107: I have one complaint here, both, the lower and upper size have a define. To prevent having to touch each item if we ever change the limits, those defines should be used.

PR #9284, opened 2024-10-23
Soluna Nexus map fixes 1.1!

Kazzc:
-Fixed various window tint ranges and variables.
-Fixed various door bolts buttons and its variables. 
-Fixed all meteor turrets, shields no longer block their line of fire. 
-Minor decal fixes.
-Adjusted shield gen location. First deck near telecoms. (120 range is enough to cover the whole station.)


Kashargul:
You pushed this to an outdated branch...

PR #9335, opened 2024-10-28
sanity checks some vorepanel things

Kashargul:
let's just catch it if we have no bellies instead of crashing. Not like any functionality will be there anyway


Kashargul:
hmmm no... this can't be null xS

PR #9441, opened 2024-11-07
test them

Kashargul:
Unit test relic base redgates


Kashargul:
needs a mirror from upstream for the rest

PR #9467, opened 2024-11-10
Slight Fix for Metroids on Soluna

EtheoBoxxman:
## About The Pull Request
Adds a third area parent type to the exclusion of metroid spawning. 
On SC, it was area/crew_quarters/sleep, but SN uses area/crew_quarters/sleeping for its area, which isn't excluded as a target for metroid spawns. There was a recent occurrence of someone being locked in with metroids mid scene in dorms because of this oversight.

Doesn't change any map files, so there shouldn't be any conflicts for mappers to worry about.
## Changelog
:cl: Etheo
fix: Fixes metroids spawning in Soluna Nexus' Dormitories
/:cl:


EtheoBoxxman:
Dunno what those failed checks are, Ive never seen them before. Didnt touch any map files, and debug server didnt yield errors.
 o .o

Kashargul:
You are using a single map define instead of protecting the areas in the map file as IT should be done. So map test will ofc fail.

PR #9636, opened 2024-12-16
[MIRROR] Adds a new ghost vis blocking area flag

CHOMPStation2StaffMirrorBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/16696
--------------------
Also using this to add the missed the defines while deconflicting downstream

🆑 
add: a new flag AREA_BLOCK_GHOST_SIGHT to prevent ghosts from seeing certain areas
add: a switch to toggle the vision blocking (not mapped in yet)
admin: adds a verb that allows admins to add / remove blocked zones for events
/🆑 

grafik

todo:
- [x] add verb to add areas for admins
- [x] add verb to remove areas for admins
~~implement a shadekin option to allow event managers to exclude them from events~~ maybe later...
- [x] null the nets in Destroy...


Kashargul:
Need to look into it why it worked on my virgo tests and doesn't work on chomp centcom

SpringSkipper:
Are these later going to be removed from ICly secure areas (armory, evidence, AI, telecomms, HoP office, etc.) and limited only to OOCly sensitive areas (just dorms, basically)?

Kashargul:
> Are these later going to be removed from ICly secure areas (armory, evidence, AI, telecomms, HoP office, etc.) and limited only to OOCly sensitive areas (just dorms, basically)?

More than likely outside of event rounds, but likely will be a day or two as I first need to fix a few of the known issues. Most effecting mostly admins, but also the one where a chunk should ignore an obfuscation update if the current area is the same as a pred ones. All that ghost staying in belly code is a mess in itself though.

SpringSkipper:
Right. I know porting is a whole process so I'll let you handle that. I just have... **strong concern** with the particular choice of areas screened.

PR #9650, opened 2024-12-16
Add a timeclock app to the PDA

NickBelmont:
## About The Pull Request
Adds a timeclock application to the pda, with area checks to try and prevent clocking out while in a restricted access area.
## Changelog
:cl:
add: Timeclock app to the set of default PDA applications
add: AREA_ALLOW_CLOCKOUT flag for setting which areas allow players to clock out using the pda timeclock
/:cl:


NickBelmont:
Right, Soluna Nexus doesn't have area defines. That puts a dent in the idea of preventing clock outs in departments, but I'll see what can be done. Would like input on whether that's something that's even needed as well.

NickBelmont:
Those changes should resolve the issues, while still mostly preventing people from clocking out inside a department or other restricted area.

modular_chomp/code/modules/pda/core_apps.dm
```diff
@@ -0,0 +1,180 @@
+/datum/data/pda/app/timeclock
+	name = "Timeclock"
+	icon = "list-alt"
+	template = "pda_timeclock"
+
+	var/channel = "Common"
+	var/mob/pdaUser = null
```
Kashargul, line 7: This reference also needs to be cleared on closing the UI.

tgui/packages/tgui/interfaces/Pda/pda_screens/pda_timeclock.tsx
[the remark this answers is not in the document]
NickBelmont: This may not be possible? At least without changing a lot of how the pda app code works. It seems that it enforces that all modules must be in pda_screens, if I'm understanding this line right:
`require.context('./pda_screens', false, /\.tsx$/);`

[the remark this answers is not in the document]
Kashargul: custom sub app routing as we do in prefs or upstream pr

modular_chomp/code/modules/pda/core_apps.dm
```diff
@@ -11,6 +11,9 @@
 	//Initialize an intercom to announce going on/off duty
 	if(!announce)
 		announce = new /obj/item/radio/intercom(src)
+/datum/data/pda/app/timeclock/stop()
+	del(pdaUser)
```
Kashargul, line 15: This will wipe the user from the game

pdaUser = null

tgui/packages/tgui/interfaces/Pda/pda_screens/pda_timeclock.tsx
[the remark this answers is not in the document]
NickBelmont: I'll take a look at how it's done with that then, thank you!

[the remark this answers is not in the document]
NickBelmont: That was surprisingly painless, all things considered. Should all be in the right place now.

tgui/packages/tgui/interfaces/chompstation/Pda/index.tsx
Kashargul, line 1: I wouldn't move the entire interface for a require, we have those ignored and comment marked, as said check the preferences. Without having added requires, we can't modularize after all. Without just the index moved, the UI will break on non thoroughly checked upstream mirrors

modular_chomp/code/modules/pda/core_apps.dm
```diff
@@ -0,0 +1,163 @@
...
+		H.mind.assigned_role = pda.id.rank
+		H.mind.role_alt_title = pda.id.assignment
+		announce.autosay("[pda.id.registered_name] has moved On-Duty as [pda.id.assignment].", "Employee Oversight", channel, zlevels = using_map.get_map_levels(get_z(src)))
+	return
+
+/datum/data/pda/app/timeclock/proc/makeOffDuty()
+	var/datum/job/foundjob = job_master.GetJob(pda.id.rank)
+	if(!foundjob)
+		return
+	if((get_area(pdaUser).flags & TIMECLOCK_RESTRICTED))
```
Kashargul, line 121: flag_check

also, those need to be added to the areas as well

Furthermore, I would allow it always if isbelly(loc)

tgui/packages/tgui/interfaces/chompstation/Pda/index.tsx
NickBelmont: Right, so just the function that checks the requires, and the minimum of whatever it needs to integrate with everything else properly, then?

modular_chomp/code/modules/pda/core_apps.dm
NickBelmont: The latter's simple enough, for adding flags to areas, should I just add on to the existing Space Station 13 areas_ch.dm file, or make a new one in modular? Or is there somewhere else I should do that?

tgui/packages/tgui/interfaces/chompstation/Pda/index.tsx
NickBelmont: Only Pda is marked with export though, so I'm not sure if less stuff can be moved? It's entirely possible I'm just not seeing something obvious though.

Kashargul: Do a chompadd / chompedit in the requite section. Those edits are fine to ensure that we can modularize. Usually I've requires as excluded when I sync those.

modular_chomp/code/modules/pda/core_apps.dm
Kashargul: our areas are by map

modular_chomp then
souther_cross_areas.dm
soluna_nexus_areas.dm
relicbase_areas.dm

Kashargul: also from how area flags work, I would turn this into AREA_ALLOW_CLOCKOUT

because most areas already have special flags, aka the bar, maints and dorms which can be appended easier than trying the inverted behaviour.

tgui/packages/tgui/interfaces/chompstation/Pda/index.tsx
NickBelmont: So, after looking a bit more, how preferences accomplishes a similar goal is by modifying the upstream module. Namely this bit here in interfaces/PreferenceMenu/preferences/features/index.ts
```
// CHOMPAdd start for modularity
const requireAdditions = require.context(
  '../../../chompstation/PreferencesMenu/preferences/features',
  true,
  /.tsx$/,
);

for (const key of requireAdditions.keys()) {
  for (const [featureKey, feature] of Object.entries(requireAdditions(key))) {
    features[featureKey] = feature as Feature<unknown>;
  }
}
// CHOMPAdd end
```
Which makes sense and would work in this case as well, but also seems to violate the whole not touching upstream files thing.

modular_chomp/code/modules/pda/core_apps.dm
NickBelmont: Right, I'll do that, then, I've flagged the dorms, most clearly public areas, hallways, and some more distinct parts of maintenance to allow clocking out on all the maps, as well as the planet surface. I'll push the commit once I test to make sure it works as intended, let me know if any of the areas should be unflagged, or if there are any that should be flagged I missed, wouldn't be surprised if I missed a few, though I tried to double check each file. Already know I missed the SC cafeteria, holodeck and game room, because those aren't defined in southern_cross_areas.dm

Kashargul: Be careful with flags, if you apply new flags to any child, the parent flags will not be applied, you ned to explicitly flag areas.

tgui/packages/tgui/interfaces/chompstation/Pda/index.tsx
Kashargul: Routes are fine to be applied to them, we had to do the same for the main tgui package and the preferences. Otherwise, we couldn't have split things to chompstation, this way we can just ensure that if upstream creates a similar interface, we have ours routed first without conflicts.

For preferences it's the other way around though, as if upstream has the same preference, we just want to delete ours.

modular_chomp/code/modules/pda/core_apps.dm
NickBelmont: I think all the flags I set are good. Looked over all of them and didn't see anything immediately being overridden.

tgui/packages/tgui/interfaces/chompstation/Pda/index.tsx
NickBelmont: Okay, I think I have it, then. I changed the function in the base index file that checks for the interface file so it checks our files first, then the upstream if it can't find the right file.

NickBelmont: And since that was all the chompstation/Pda/index file was really doing, I deleted that.

tgui/packages/tgui/interfaces/Pda/index.tsx
```diff
@@ -25,16 +25,26 @@ type Data = {
 };
 
 const requirePdaInterface = require.context('./pda_screens', false, /\.tsx$/);
-
```
Kashargul, line 28: // CHOMPEdit Start

tgui/packages/tgui/interfaces/Pda/index.tsx
```diff
@@ -25,16 +25,26 @@ type Data = {
...
+    try {
+      appModule = requirePdaInterface(`./${name}.tsx`);
+    } catch (err) {
+      if (err.code === 'MODULE_NOT_FOUND') {
+        return routingError('notFound', name);
+      }
+      throw err;
     }
-    throw err;
   }
```
Kashargul, line 48: below the }
//CHOMPEdit End

please tag the full block where editing started and ended to make sure that edit is shown as a single conflict block. Remove the other comments in between

Kashargul (changes requested):

NickBelmont:
That commit should resolve those two changes. Also corrected an error I made in a comment describing the AREA_ALLOW_CLOCKOUT flag, it was still phrased like the flag was for blacklisting areas from the clockout function, rather than the other way around.

NickBelmont:
So I think that's everything, then? I can do a few more passes for areas that should be flagged, but I think I got the important ones, and I think clocking out in the dorms or a belly are the most important ones anyways, and I know I got those, along with most of the hallways and public spaces.

NickBelmont:
Actually, going to push one more thing through in a bit to try and make the UI work more nicely.

NickBelmont:
There we go, that fixed something that was bothering me about how the UI worked, it would hide the last few items behind the footer of the PDA screen sometimes.

NickBelmont:
Also went through and double-checked every area I changed flags on to make sure I wasn't screwing anything up, caught two instances where RAD_SHIELDING needed to be added back, but I exhaustively checked every reference to maintenance and other rad shielded areas and those were the only two that I found after that, so it should be good on that front.

NickBelmont:
Same with the other references, though those tended to be applied selectively enough that I was adding to an existing set of flags, rad shielding was the big one because any maint area has it and needed it readded if clocking out was allowed.

Kashargul:
> Same with the other references, though those tended to be applied selectively enough that I was adding to an existing set of flags, rad shielding was the big one because any maint area has it and needed it readded if clocking out was allowed.

Check my message on discord about the UI datum

NickBelmont:
Right, that all looked good, and is a much better approach than effectively storing the user again. Only change I made was  adding a missing tab on the definition of H on line 138.

NickBelmont:
Also not sure about `foundjob.current_positions--` being in the if statement, seeing as that could possibly cause issues with IDs being assigned without actually changing the number of job slots, but I left it there in case it was intentional.

Kashargul:
> Also not sure about `foundjob.current_positions--` being in the if statement, seeing as that could possibly cause issues with IDs being assigned without actually changing the number of available job slots, but I left it there in case it was intentional. Though I would think either that should be moved out of the if(ishuman(user)) check or `foundjob.current_positions++` in the clock out function should be moved in, just to maintain consistency in what will occur.

Will take a look at that, the issue with PDAs here is more that non humans have the app as well.

Kashargul:
> Also not sure about `foundjob.current_positions--` being in the if statement, seeing as that could possibly cause issues with IDs being assigned without actually changing the number of available job slots, but I left it there in case it was intentional. Though I would think either that should be moved out of the if(ishuman(user)) check or `foundjob.current_positions++` in the clock out function should be moved in, just to maintain consistency in what will occur.

Looking into it, I don't even know why that typecast was done in the first place. user.mind should work just fine.

What needs to be tested after though is what happens if a robot uses it.

NickBelmont:
> Will take a look at that, the issue with PDAs here is more that non humans have the app as well.
That's what I assumed was the reason for the check, in that case, perhaps either check both if the selected job exists and if the user is human at the same time? So at line 128 change it to `if(ptojob && ishuman(user))`? The other option would be to change the define of H to just be a mob, since mind is part of the mob datum, so we don't seem to check if the user is human aside from if we want to keep non-humans from clocking in/out.

Kashargul:
> > Will take a look at that, the issue with PDAs here is more that non humans have the app as well.
> 
> That's what I assumed was the reason for the check, in that case, perhaps either check both if the selected job exists and if the user is human at the same time? So at line 128 change it to `if(ptojob && ishuman(user))`? The other option would be to change the define of H to just be a mob, since mind is part of the mob datum, so we don't seem to need to check if the user is human aside from if we want to keep non-humans from clocking in/out.

adjust it to use user.mind, remove the if and the typecast and then give it a test, we might want to just disable the entire app if it's not a human user. 

if(!ishuman(user))
    close_ui()

NickBelmont:
> Looking into it, I don't even know why that typecast was done in the first place. user.mind should work just fine.
> 
> What needs to be tested after though is what happens if a robot uses it.

I don't see much of a reason aside from maybe preventing borgs, monkeys and simplemobs from using it? If that is the purpose and we want to maintain that, I'd just add the check of being human to the if(ptojob) and if(newjob) checks. If borgs don't accumulate PTO, user.mind should work fine, because they can't clock out in that case. Otherwise, it will need some sort of check I think.

NickBelmont:
> adjust it to use user.mind, remove the if and the typecast and then give it a test, we might want to just disable the entire app if it's not a human user.
> 
> if(!ishuman(user)) close_ui()
> 
> also one more thing to test, cross references, the user of the pda might not be the owner of the ID.

That also works, and was considering the latter as well, given what would seem to happen in that event.

NickBelmont:
So, for cyborgs, their PDAs are apparently defined elsewhere, it seems they don't have the timeclock app by default, and since they can't use other people's, that means the only reason to check if the user is human is for monkeys and simplemobs.

NickBelmont:
Seems to all work as intended. I enforced the human requirement on the UI end, so the app still opens, you just can't do anything but see the current assignment on the id and your pto hours. I'll push that through in a moment. For cross-referencing, how should that be done? All the identifiers an ID has can either be modified via mechanics in some way, right? Unless DNA and fingerprint hash have nothing to do with genetics/UE.

Kashargul:
> Seems to all work as intended. I enforced the human requirement on the UI end, so the app still opens, you just can't do anything but see the current assignment on the id and your pto hours. I'll push that through in a moment. For cross-referencing, how should that be done? All the identifiers an ID has can either be modified via mechanics in some way, right? Unless DNA and fingerprint hash have nothing to do with genetics/UE.

enter your ID into the PDA, drop it, let someone else use the PDA, aka another account. Testing this is best done with 2 accounts as you might be altering the users mind datum over the id owner's.

NickBelmont:
> enter your ID into the PDA, drop it, let someone else use the PDA, aka another account. Testing this is best done with 2 accounts as you might be altering the users mind datum over the id owner's.
> 
> The easiest would be to pass the "name" when the ID is entered and give a message like "this is not your PDA" if the user name does not match the ID name.

Right, that should work most of the time, I suppose, there's still edge cases that can happen. If I recall, the ID modification console can change the name on an ID, and there's a few ways the player name can be changed too I think. Can certainly implement that though.

Kashargul (approved):

Kashargul:
Sadly area flags will have to be adjusted after the other TM has been merged, there will be conflicts.

NickBelmont:
Okay, that's alright, gives me another opportunity to make sure they're all implemented properly anyways when it comes around.

NickBelmont:
I assume the TM you're referring to is the ghost vis blocking one? If so, thankfully it's only a few areas that overlap from what I saw, I'll make the needed changes once it's merged.

Kashargul:
> I assume the TM you're referring to is the ghost vis blocking one? If so, thankfully it's only a few areas that overlap from what I saw, I'll make the needed changes once it's merged.

It is, I just need another iteration on it sadly after rolling it out further, we found a few more issues to address

NickBelmont:
Should I also change the pda app UI to make it consistent with the timeclock cooldown change that was just merged?

Kashargul:
> Should I also change the pda app UI to make it consistent with the timeclock cooldown change that was just merged?

would not be bad to prevent spamming

NickBelmont:
Okay, I'll see about integrating the area flag check into it as well, for the same reasons.

NickBelmont:
I think the name check can stay where it is, because if that fails there's no log or admin message for it just the notice to the user, but that could also be enforced in the UI if desired.

tgui/packages/tgui/interfaces/chompstation/Pda/pda_screens/pda_timeclock.tsx
```diff
@@ -10,14 +10,17 @@ import {
   Section,
 } from 'tgui/components';
 
+import { formatTime } from '../../../../format';
```
Kashargul, line 13: heavily recommending to use tgui/format at that depths

NickBelmont: Right, that would be easier, I'll push that through in a minute.

NickBelmont:
I'll look through to make sure the changes to flags from the merge aren't overridden by any of my changes, already fixed one instance.

Kashargul:
> I'll look through to make sure the changes to flags from the merge aren't overridden by any of my changes, already fixed one instance.

I would heavily prefer if we move flagged areas to ss13 areas_ch and no longer define flags in the map files itself

NickBelmont:
I can try to do that as well, then. Is there a level of depth or specificity where it should stay in the map files, or should any area that's flagged be moved?

Kashargul:
> I can try to do that as well, then. Is there a level of depth or specificity where it should stay in the map files, or should any area that's flagged be moved?

My fear is there that people adjust child flags and forget that parents have map based overrides, so while the names and such can stay, I would at least move the flags to the area file, to be sure that people don't override them on the map level.

NickBelmont:
That should move most of them, I didn't touch anything that was already flagged, along with some of the really specific area defines the Soluna Nexus uses.

NickBelmont:
And that should fix almost all of them. I also moves the turbolift, surface/outside, and surface/cave flag defines, since they were clockout areas, and given how they're defined the same in each map, they seemed like good ones to move the flag defines to the areas file.

NickBelmont:
I can still commit while it's test merged, right? Just want to move a few more AREA_ALLOW_CLOCKOUT flags to the area define file, should be almost all of them on the SC and actually all of them for the relicbase.

Kashargul:
> I can still commit while it's test merged, right? Just want to move a few more AREA_ALLOW_CLOCKOUT flags to the area define file, should be almost all of them on the SC and actually all of them for the relicbase.

Yes, just to let me know when I should update the TM, best on Discord though. As I won't look in here a lot the next days.

PR #9669, opened 2024-12-20
Fire Rework

TheCaramelion:
## About The Pull Request

Reworked fire code. Because there's nothing better than burning things to the ground.

Upports [Fire Code Rework#389](https://github.com/Willburd/CHOMPost21/pull/389)

## Changelog
:cl:
qol: Fire is a proper danger now
add: Electrical Fire event
/:cl:


TheCaramelion:
Oh yeah, Test Merging this would be neat in case something goes terribly wrong (:

Willburd:
Be sure to add the temp limit changes from this commit, or the fire WILL breach the engine. You should also probably limit it to being lower than the melting point of your engine's intended wall material. Or the lingering fires will continue to push the temp upward forever. We found this out the hard way with our hellburn style incinerator breaching near the end of the test round.

https://github.com/Willburd/CHOMPost21/commit/9ff8cb75465506e64021607c6d2c6e5c0564ee73

PR #9671, opened 2024-12-22
Escaping vr is no longer possible

Colfer1:
## Changelog
:cl:
fix: fixed exploit allowing vr users to escape into the real world
/:cl:


Colfer1:
Tested locally and shelter capsules still seem to work outside of VR with no issues, but I did not test it extensively, just with the basic 500 point one. Confirmed shelter capsules do not work in VR with this change, thus, escape is impossible

PR #9691, opened 2024-12-25
fixing empty hardsuit bug in cargo order.

Asher-42:
## About The Pull Request
this was reported in suggestions as changing name to empty to reflect, but the code suggests its intended to have the modules, so i fixed it. it was using the wrong object.

## Changelog
:cl:
fix: adds modules to med solgov hardsuit, due to it using the wrong obj path. 
/:cl:

### DIFF
--- a/code/datums/supplypacks/hardsuits.dm
+++ b/code/datums/supplypacks/hardsuits.dm
@@ -32,7 +32,7 @@
 	containertype = /obj/structure/closet/crate/secure/gear
 	containername = "industrial hardsuit crate"
 	access = list(access_mining,
-				  access_eva)d
+				  access_eva)
 	one_access = TRUE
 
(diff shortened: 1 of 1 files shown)


code/datums/supplypacks/hardsuits.dm
```diff
@@ -84,7 +84,7 @@
 	name = "solgov medical hardsuit (loaded)"
 	desc = "A fully-equipped Commonwealth of Sol-Procyon Medical hardsuit. Requires Medical access."
 	contains = list(
-			/obj/item/rig/baymed = 1
+			/obj/item/rig/baymed/equipped = 1
```
Kashargul, line 87: // CHOMPEdit or upstream, we only marked the descriptions in that section

PR #9663, opened 2024-12-19
Better Biogenerator

Missile597:
## About The Pull Request

Adds some more items to the biogenerator to bridge the gap in botany with some things that are needed/highly desirable in that department. Aim was to make the biogenerator (something that barely gets used at the moment) more useful without just eclipsing chemistry or cargo completely.

## Changelog
:cl:
add: Added new materials to the biogenerator list: Diethylamine, Unstable Mutagen, Plant-B-Gone, Table Salt, Soy Sauce, Flora Disk, Floral Somatoray, Kudzu Seeds, Random Exotic Seeds
balance: Lowered the costs of the fertilisers and botany materials to make them easier to obtain. After all, the department uses these a lot.
/:cl:

PR #9487, opened 2024-11-16
Medical QoL Updates and Chem Rebalance

Missile597:
## About The Pull Request

Just a few adjustments to medical and virology for Southern Cross.

## Changelog
:cl:
maptweak: Added some lights in darker areas of medical
maptweak: Fixed the genetics sorter so the lab is no longer a dumping ground and deliveries will now work
maptweak: Added a bottle of Adranol pills and a bottle of Mutagen to give some starting chems and hints for viro. Also added some phoron and a grinder for same reason.
maptweak: Added the sign by viro disposals reminding people that it goes into space
maptweak: Added the viro requests console back into viro
maptweak: Fixed the chemistry tables
maptweak: Finally moved Erik's Lungs out of the way
maptweak: Added a flu culture to the maints loot with the cold culture
maptweak: Adds droppers to viro
maptweak: Replaced the chem dispenser in xenobotany with 2 bottles of Diethylamine and 2 bottles of Mutagen
maptweak: Removed xenobiology's chem dispenser and halved the starting phoron stacks
maptweak: Removed the chem master and dispenser from the science ship and replaced it with an RnD console and Destructive analyser
maptweak: Removed the chemistry lab in xenoarch - the lab already has a massive coolant supply. Added a sulphuric acid dispenser in its place
maptweak: Removes the crew monitor from xenoarch
/:cl:

PR #9733, opened 2024-12-30
Adds a small xenochemistry area between flora and bio science

Scarlet-Oduko:
## About The Pull Request

Adds a small xenochemistry area between flora and bio science divisions
## Changelog
:cl:
add: Adds a small xenochemistry area between flora and bio science divisions
:cl:

### DIFF
--- a/modular_chomp/maps/southern_cross/southern_cross-2.dmm
+++ b/modular_chomp/maps/southern_cross/southern_cross-2.dmm
@@ -14439,7 +14439,7 @@
 	icon_state = "2-8"
 	},
 /obj/structure/table/standard,
-/obj/machinery/chemical_dispenser/xenoflora/full,
+/obj/item/storage/box/beakers,
 /turf/simulated/floor/tiled/hydro,
 /area/rnd/xenobiology/xenoflora)
 "dZe" = (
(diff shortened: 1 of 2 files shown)


Missile597:
Please can we at least give #9663 a chance before just immediately reverting the change staff requested in #9487 ?

Missile597:
Firstly, this basically now just gives science a chemistry lab of their own which is basically 90% of what medical is for which I'm not keen on at all. Especially as this makes the biogenerator useless again
Secondly, this area isn't functional as it's missing an APC so it would be unpowered
Thirdly, no fire doors on the entrances to this new area and no fire alarm present.

New room is also missing a comms panel and a light switch. In fact, I think it's missing lights completely? Also no disposals bin would likely get annoying too.
image

Kazzc:
Wrong decals under doors
Missing decals & misplaced decals
Missing various wall mount machines
Missing trash chute
Missing wall coms
Missing floor vents and scrubbers
Missing lights
Removed lockers and stock supply items, should be placed in storage or somewhere near.

There are so many missing and misplaced things here. . . If we want to introduce Xenochemistry, best placed somewhere else.
Anywhere on the third floor is more than enough space.

Cyrelius:
It's meant to be chemistry for Xenobotany and Xenobiology, adding it to the third floor would be stupid.
It may indeed need some fixing, but definitely needs to be located around where it is.

Kashargul:
Closing for stale.

PR #9751, opened 2025-01-01
Reverts Marish

whooshboom:
## About The Pull Request
Based on popular consensus, reverts Marish and the lore surrounding it to previous audible condition.

Does not revert the font.

## Changelog
:cl:
qol: Reverts marish to audible condition. Happy marring!
/:cl:


whooshboom:
fucked up and committed two changes to the same request

PR #9786, opened 2025-01-04
Fix multiple KIN exploits

Kashargul:
## Changelog
:cl:
fix: over 10 known shadekin exploits...
/:cl:

### DIFF
--- a/code/datums/diseases/advance/symptoms/sneeze.dm
+++ b/code/datums/diseases/advance/symptoms/sneeze.dm
@@ -109,7 +109,7 @@ Bonus
 
 	var/mob/living/carbon/human/unlucky = locate() in place
 
-	if(unlucky)
+	if(unlucky && !unlucky.is_incorporeal()) // CHOMPEdit
 		if(unlucky.can_be_drop_pred && H.can_be_drop_prey && H.devourable)
 			place = unlucky.vore_selected
 		else if(unlucky.devourable && unlucky.can_be_drop_prey && H.can_be_drop_pred)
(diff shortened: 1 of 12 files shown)


EtheoBoxxman:
What are the exploits though?
I dont think spontaneous predding/preying as a phased shadekin is an exploit. o:

Kashargul:
> What are the exploits though? I dont think spontaneous predding/preying as a phased shadekin is an exploit. o:

Certainly not going to list them in public.

SpringSkipper:
> What are the exploits though? I dont think spontaneous predding/preying as a phased shadekin is an exploit. o:
Seems to be most forms of spontaneous vore, a lot of collision-based oversights with things hitting phased 'kin that shouldn't (throwing, shooting, bumping), and some inventory manipulation stuff while phased.

Honestly only that last one and partially the second one really feel like exploits. I don't see why spont vore stuff and the oversight stuff can't be openly discussed here. Though I'd argue that whether shadekin _should_ be able to spont vore while phased is best brought up in suggestions chat on the discord, if you feel strongly on it.

Not getting bumped and shuffled around by movement as a phased kin is already wonderful. Thanks, Kash!

Kashargul:
> > What are the exploits though? I dont think spontaneous predding/preying as a phased shadekin is an exploit. o:
> 
> Seems to be most forms of spontaneous vore, a lot of collision-based oversights with things hitting phased 'kin that shouldn't (throwing, shooting, bumping), and some inventory manipulation stuff while phased.
> 
> Honestly only that last one and partially the second one really feel like exploits. I don't see why spont vore stuff and the oversight stuff can't be openly discussed here. Though I'd argue that whether shadekin _should_ be able to spont vore while phased is best brought up in suggestions chat on the discord, if you feel strongly on it.
> 
> Not getting bumped and shuffled around by movement as a phased kin is already wonderful. Thanks, Kash!

Shadekin where actively pref breaking around the phase vore pref.

PR #9923, opened 2025-01-23
Prepares pda_timeclock for migration to tgui-core

NickBelmont:
## About The Pull Request
Prepares the PDA Timeclock interface for migration to tgui-core
## Changelog
:cl:
code: Changes tgui/components and tgui/format to their tgui-core equivalents, and replaces a Flex element with Stack
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/chompstation/Pda/pda_screens/pda_timeclock.tsx
+++ b/tgui/packages/tgui/interfaces/chompstation/Pda/pda_screens/pda_timeclock.tsx
@@ -4,12 +4,12 @@ import { useBackend } from 'tgui/backend';
 import {
   Box,
   Button,
-  Flex,
   LabeledList,
   NoticeBox,
   Section,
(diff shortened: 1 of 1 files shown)


Kashargul:
common is known, there's no more need for the ../../../common which makes moving uis a mess. It can simply be accessed by common/Stuff With tgui core we also rid of all static paths outside of the uis, this should make a lot of things easier

NickBelmont:
Does it matter that RankIcon is in packages/tgui/interfaces/common, instead of packages/common where all the other things that were able to just be accessed by common/ are? Are both now known/being checked?

Kashargul:
> Does it matter that RankIcon is in packages/tgui/interfaces/common, instead of packages/common where all the other things that were able to just be accessed by common/ are? Are both now known/being checked?

My mistake then, sadly the common interfaces are still a bit of an annoyance for a while longer.

PR #10024, opened 2025-02-01
partial taur sprite fix

Cameron-The-Raven:
## About The Pull Request

Fixes some taur sprites causing all worn objects to become invisible.
A recent upstream PR made 'taur_clip_mask_def' required in the .dmi.
- The _ch, _ch loaf file had it. _ch files are on upstream now, so they those were fixed there. (I could merge the _ch files into the normal taur sprites upstream, but if I do that we're going to probably get banned from icondiffbot) See: https://github.com/VOREStation/VOREStation/pull/17017

- The _yw file also had it, but it's on chomp only, so it was fixed here.

before fix: 

after fix:

## Changelog

:cl:
fix: _yw taur sprites properly show now
/:cl:


Kashargul:
you can merge them, just not all in one PR, do 1 file at a time with such massive diffs

PR #10142, opened 2025-02-11
[MIRROR] Adds Trait Genetics

CHOMPStation2StaffMirrorBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/16921
--------------------
## About The Pull Request

Adds trait genetics from Outpost 21 (Made by Willburd)
NOTE: MAP STILL NEEDS A GENETIC LAB MAPPED IN TO THEIR APPROPRIATE MAPS. THAT WILL BE FOR ANOTHER PR.
- Tried to update each file it touched to be equivalent to their Chomp specific variants in order to maintain functionality and make sure no oddities would happen due to code differences. (Some things like the eardeaf loop were left there but commented out). 
- This means that a good amount of bugfixes inadvertently got included. (One of note is objects dropped in disposals won't be forever trapped/show up outside of the pipes)
- Some bugs were stumbled upon during testing (AI eye hard deleting, proteans hard deleting) and was fixed while being worked on.
- A bunch of new traits were added as well, expanding both the loadout selection and the ones that genetics can find.
- Has hidden traits that can't be selected at round start and can only be found via genetics. Some of these are dangerous (Gibbingtons) and some are beneficial (no-fingerprints, thermal vision, laser eyes, hulk.)

How it works:
Much like OG genetics, you put a person in the DNA scanner, attach the console to them, and use the console
You can adjust the duration & intensity, trying to raise a gene above DAC to activate it.
Once it's above DAC, you can either watch the monkey to do a 'tell' of what gene it is, or inject yourself/a willing participant and see what happened!  (This causes radiation and damage, so watch out!)
Repeat!

## Changelog

:cl: Will, Kash, Diana
fix: Fixes a server crashing issue with flip emote
fix: Fixes Genetics and all the powers that came with it
add: Adds Trait Genetics to the game. (Genetics can activate/deactivate traits)
add: Various new traits up-ported (drippy, rad-resistance/immunity, Photosynthesis, Rotting Genetics, Incomprehensible, Less blood volume, slowdown EXTREME, low blood sugar, Agoraphobia, Loneliness, Glass Endurance, reduced biocompat, photosensitivity, haemophillia, pain intolerance, sensitive biochem, table passer,)
add: Health analyzers tell you if someone is  suffering from a genetic side effect and what to give to cure them.
add: Allows cocoon weaver to change eye and skin color
fix: Various small bugfixes up-ported from Chomp
sound: Added some founds from Chomp for space inhale/exhale sounds
[...]

PR #10144, opened 2025-02-11
Fix characters with blood volume altering traits spawning with normal blood volume

BigBadE:
## About The Pull Request

Currently, the trait var change sets the species blood_volume after the entity's blood_volume is set, causing anyone with the high blood volume trait to instantly die.

## Changelog

:cl:
fix: characters with blood volume-altering traits spawning with normal blood amount
/:cl:

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -337,6 +337,15 @@ var/list/preferences_datums = list()
 	// Ask the preferences datums to apply their own settings to the new mob
 	player_setup.copy_to_mob(character)
 
+	// CHOMPEdit start Fix blood levels after trait apply (since the extra/less blood traits set the var after)
+	if(character.vessel.total_volume < character.species.blood_volume)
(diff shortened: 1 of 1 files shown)


code/modules/client/preferences.dm
```diff
@@ -337,6 +337,15 @@ var/list/preferences_datums = list()
 	// Ask the preferences datums to apply their own settings to the new mob
 	player_setup.copy_to_mob(character)
 
+	// CHOMPEdit start Fix blood levels after trait apply (since the extra/less blood traits set the var after)
```
Kashargul, line 340: fundamental changes best go upstream

Kashargul (changes requested):

TheCaramelion:
This was included in #10142, I believe

BigBadE:
> This was included in #10142, I believe

Ah, I didn't realize that was also changing the same thing. I guess I'll just close this and wait for that to merge?

PR #10217, opened 2025-02-23
Map Fixes and Access list update

Missile597:
## About The Pull Request

Fixes some minor map issues and updates the config file with an up to date list of the access level numbers to help with proper future mapping. Also had an OCD sergal update the scrubber loops on Sif to help the xenoarch nerds.

## Changelog
:cl:
maptweak: Removed access requirement to main part of space disposals to muffin monster can be accessed without ID to enable retrieval of flushed items
maptweak: Chemical dispensers returned to xenobotany and to xenoarchaeology
maptweak: Sif outpost and xenoarchaology outpost scrubber loops updated to filter properly and also filter volatile fuel
maptweak: Removed access requirement to the central maints tunnel. Any secure exits to this tunnel are still access protected
maptweak: Fixed medical bathrooms not being attached to station atmos. Removed some excess piping and wiring going to no destination in maints
maptweak: Added a bit more power and some lights to the starhunter base. The vending machines should hopefully have enough power now
maptweak: Fixes the heavy duty wire by the outpost R-UST that wasn't properly oriented. Power should flow now
maptweak: Tweaked security doors to have correct access levels. Main areas will now be Security instead of Security equipment.
/:cl:

### DIFF
--- a/config/example/access levels.txt
+++ b/config/example/access levels.txt
@@ -16,64 +16,104 @@ For example, a brig door would have it be "2" while a door that requires you hav
 
 Here is a list of the permissions and their numbers (this may be out of date, see code/game/access.dm for an updated version):
 
-access_security = 1
-access_brig = 2
-access_armory = 3
-access_forensics_lockers= 4
-access_medical = 5
-access_morgue = 6
-access_tox = 7
-access_tox_storage = 8
-access_genetics = 9
-access_engine = 10
-access_engine_equip= 11
-access_maint_tunnels = 12
-access_external_airlocks = 13
-access_emergency_storage = 14
-access_change_ids = 15
-access_ai_upload = 16
-access_teleporter = 17
-access_eva = 18
-access_heads = 19
-access_captain = 20
-access_all_personal_lockers = 21
-access_chapel_office = 22
-access_tech_storage = 23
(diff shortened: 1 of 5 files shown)


config/example/access levels.txt
```diff
@@ -16,64 +16,104 @@ For example, a brig door would have it be "2" while a door that requires you hav
 
 Here is a list of the permissions and their numbers (this may be out of date, see code/game/access.dm for an updated version):
 
-access_security = 1
```
Kashargul, line 19: Isn't this an upstream file?

Missile597: I don't know?
The actual accesses are found in this file https://github.com/CHOMPStation2/CHOMPStation2/blob/9db6849a68309e382deb2923e6c6d5135a91f0d3/code/game/jobs/access_datum.dm#L474
The file I've changed is just the text file in the config/example folder which is supposed to give mappers a guide to what the access numbers are (I just updated it with the correct information. I'm unsure if this file is upstream or if upstream even has the same access codes (likely some will be different.
If you need //CHOMPEdit s or if you would rather do stuff upstream instead just let me know and I can snip out he text file change. I already asked Asmuth to update the wiki anyways

PR #10230, opened 2025-02-26
Update relicbase_things.dm

FluffMedic:
## About The Pull Request
Not sure if this is the best fix, but altering the temperature to match outside fixed the active edge thing on the skymaps.
This doesn't touch broken gateways/redgates

## Changelog
:cl:
fix: Altering the walls to have the same temp as the rest of the Thor
/:cl:


Kashargul:
This needs a larger rework of the entire map

PR #10232, opened 2025-02-27
[MIRROR] Motion tracker subsystem

CHOMPStation2StaffMirrorBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/17215
--------------------
## About The Pull Request
dreamseeker_2025-02-23_20-42-20
(This is with 100 spiders behind those walls)

Adds a subsystem for items and mobs to subscribe to. Any sound, and various motions(and specific events) in the code will trigger pings to the subsystem. Any object subscribed to the subsystem will receive the information in these pings. The item or mob can then decide what to do with this such as flick()ing an animation.

Mobs subscribed to the subsystem may also request an image echo be generated at the location of the ping. Mobs subscribed to the motion tracker will receive a client-side image for each echo relevant to them (based on if the mob should see that echo or not in its signal handler logic), precautions have been made to reduce the overhead of multiple images per client.

Such caution include:

    Logging the turf of the echo's location, and using it's ref in a list to prevent any duplication of echos per turf.
    All echos are processed in a list handled by the subsystem's current run and tick check.
    Echos added during the current run still matter, unless they are already present in the current run, and will thus be ignored(as they have already been handled effectively)
    As the mobs handle the signal themselves to check range and darkness/vis checks, only echos actually relevant to gameplay for the mobs subscribed are actually generated.
    Items subscribed to the subsystem will not request echos by default, and only react to the signal.
    movement echos are self deleting images sent to client image lists.
    all references in the subsystem are either weak references, or reference id strings.

A positive genetrait has been added called "Vibration Sense" it will turn on the species var has_vibration_sense and subscribe the mob to the subsystem. mobs with has_vibration_sense cannot be unsubscribed from the motion tracker subsystem unless they are Destroy()ing.

A motion tracker device and lathe recipe have been added. When mobs pick up the device they will be subscribed to the subsystem, as if they have the vibration sense trait. When the item is put in a bag, dropped, destroyed, or otherwise leaves their hotbar. It will unsubscribe them from the subsystem.

A verb has been added in abilities to toggle the visibility of echos if players find them annoying.
## Changelog
:cl: Will
add: Adds a new motion tracker subsystem to the game to allow for tracking entities through walls!
add: Adds a motion tracker item to R&D
add: Adds a vibration sense trait that allows users to hear entities moving through walls
qol: Teshari's and Xenochimera's 'listen in' verb has been upgraded to the new listen-in trait

/:cl:


Skysteak:
This is likely going to come down to play honour, but you may wish to have an admin only item or something else which disables this when used/worn, for the  purposes of stuff like Hide and Seek events.

Willburd:
> This is likely going to come down to player honour, but you may wish to have an admin only item or something else which disables this when used/worn, for the purposes of stuff like Hide and Seek events.

A proc to disable pings from the subsystem would also work, and allow admins to entirely disable all echo mechanics until re-enabled.

Kashargul:
> This is likely going to come down to player honour, but you may wish to have an admin only item or something else which disables this when used/worn, for the purposes of stuff like Hide and Seek events.

Will follow the next days.

PR #10295, opened 2025-03-04
[MIRROR] Simultaneous map definitions

CHOMPStation2StaffMirrorBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/16924
--------------------
## About The Pull Request

This PR is a follow up to #16861 making it so tether, ground base and stellar delight can have their map paths all simultaneously defined so all can be rendered in actions like nanomaps, MDB, and can be worked on in strongdmm without use of the map switch. This should bring the option of a map vote closer. However, the map switch is still needed to handle overrides and actually determines what dmm files get included.

This ought to be TM'd before just outright merged. But I cannot find anything obviously broken by the change.

## Changelog
:cl: Drathek
refactor: Reworked map includes for ground base, stellar delight, and tether to have all paths included simultaneously
/:cl:


Kashargul:
code\controllers\subsystems\planets.dm

needs adjustments with this too, see comments in the file

PR #10410, opened 2025-03-14
Hotfix - Ends the giant cave spider migration (and makes them small like they should be)

SpadesNeil:
## About The Pull Request

Cave spiders got increased inside instead of halved in size.

I would sure love to run a local test server to actually test this stuff better, but for some reason whenever I do that the game is basically broken.

## Changelog
:cl:
fix: Makes giant cave spiders no longer giant
/:cl:


modular_chomp/code/modules/mob/living/simple_mob/subtypes/animal/spider.dm
```diff
@@ -232,4 +232,4 @@
 
 /mob/living/simple_mob/animal/giant_spider/tunneler/cave/Initialize(mapload)
 	. = ..()
-	resize(50)
+	resize(0.50)
```
SpadesNeil, line 235: That IS how this proc works, right? >_>

PR #10470, opened 2025-03-19
Removes all weird windows, and also gives Noodle a cure little terrarium!

SpadesNeil:
## About The Pull Request

What started as a simple PR to give Noodle a terrarium opened up a can of worms when I noticed that the reinforced windows on the maps were absolutely and utterly fucked. Some had randomly high or low health, some had _null_ health for some reason, and one set of windows had an access requirement on it. Like, what the fuck? How? Why? Anyway, I went through all the main maps to remove those weird windows. The only one I didn't touch was the non-playable levels like CentCom and the shuttle space since I don't know if there are any changes on purpose there.

Oh yeah and Noodle's pen came out great.
image

## Changelog
:cl:
maptweak: Removed weird windows with weird variables they shouldn't have.
maptweak: Noodle has a terrarium now!
/:cl:

PR #10472, opened 2025-03-20
[Refactor] Reorganizing the variables on map files and the files in vorestation.dme

SpadesNeil:
## About The Pull Request

This changes literally nothing about the game, it just makes future PRs to the map not look like dogshit when people edit with DreamMaker.

All it does is shuffle around all the variables to the order they're supposed to be in. I don't know what program was used to make these maps, but everything is all out of order.

Also, are you editing the .dme file manually or something? This is supposed to be in alphabetical order, so DreamMaker keeps changing it every time I open this program. If SS13 code writing has evolved to like a DreamMaker 2.0 you guys need to tell me where to download it so I stop having to put up with this nonsense.

Maybe this problem doesn't happen with strongdmm or whatever but why not just fix it so people using DreamMaker stop running into issues? Hence this PR.

This PR will conflict with https://github.com/CHOMPStation2/CHOMPStation2/pull/10470 so if this gets merged close https://github.com/CHOMPStation2/CHOMPStation2/pull/10470 and I'll have to do it over.

## Changelog

:cl: Ace
refactor: SouthernCross map files
refactor: Vorestation.dme
/:cl:


SpadesNeil:
To be clear,

I get if you don't "support" people using DreamMaker. But this is all stuff that should be in order anyway and it's just not. Whatever program you are using must not actually care about it.

SpadesNeil:
image
As usual, I am blameless.

SpadesNeil:
After further investigation, I think what is happening is StrongDMM just adds lines of variables when you want to change something about an object on the map, whereas DreamMaker puts all the changed variables in the **same order** as they would have normally been defined in the original object path.

One of these methods (StrongDMM) is random, while the other (DreamMaker) is sane.

StrongDMM really should be updated to do it the same way, but until that happens, this kind of refactor pass should be done every now and then.

SpadesNeil:
Okay so I know 9000 lines of changes looks intimidating but it's really not a complicated or very time consuming PR to review.

This literally just moves lines around into the order they're supposed to be in. StrongDMM doesn't care what order the map is set up in but DreamMaker does. This is a quality of life change for anyone who chooses (or has to) use DreamMaker because StrongDMM does not support load-on-runtime maps or whatever the official name is for those.

Kashargul:
massively in conflict. Any redoing would be better if at all

PR #10487, opened 2025-03-21
Teshari Radiation Suit Fix

crabcube:
## Changelog

:cl:

qol: made something easier to use

Makes teshari radsuit slowdown match the normal radsuit slowdown. As teshari are doubly affected by the slowdown modifier. This item needs a reason to exist. 

Not done - change the helmet item icon to match the normal radsuit helmet in hand and on floor. - Would like help with this!
/:cl:


Kashargul:
Upstream merged

PR #10491, opened 2025-03-21
Ignore me

Cameron-The-Raven:
Using this to compare the changes.

This will be good to merge into the attached branch after the vrpod unbreaking part 2 gets merged upstream


code/modules/mob/living/living.dm
```diff
@@ -1482,3 +1482,141 @@
 			drop.invisibility = 0
 	//else
 		// come up with drips for other mobs someday
+
```
Kashargul, line 1485: mob tf?

PR #10528, opened 2025-03-25
Funny Event Things

FluffMedic:
## About The Pull Request

Just some funny creatures for event. So spoilers if looking at code.
And possibly gateway once the event is over, depending if it goes well or not.
Any issues that arise, will poke when awake.
## Changelog
:cl:
add: Something for event coming up.
/:cl:


FluffMedic:
Curse you relic base

PR #10544, opened 2025-03-28
we'll see what happens

Kashargul:
TM ONLY, DO NOT MERGE

Might break the server or might work.


Kashargul:
Redcudes the lag issues, but needs a proper implementation

PR #10571, opened 2025-04-01
Event Stuff Part 2

FluffMedic:
## About The Pull Request
Did some balancing, mostly nerfing/reworking the oranges and yellows
Fixed the shockwave attack
Added more things for event
## Changelog
:cl:
add: more event things
/:cl:


modular_chomp/code/modules/mob/living/simple_mob/subtypes/mechanical/nanoweave.dm
```diff
@@ -0,0 +1,768 @@
...
+
+/obj/effect/callstrike_nanoweave
+	anchored = TRUE
+	density = FALSE
+	mouse_opacity = 0
+	icon ='modular_chomp/icons/mob/hivebot.dmi'
+
+/obj/effect/callstrike_nanoweave/Initialize(mapload)
+	.=..()
+	icon_state = "arti"
```
Kashargul, line 293: we don't need to set this on init, we can set this right above

FluffMedic: right
Fixed it

PR #10648, opened 2025-04-13
Adds some more cosmetic turfs

Cameron-The-Raven:
Grass that look like sif turf
## About The Pull Request
Adds two grass flooring that looks like Sif turf. For use on stationside.

I haven't touched turfpacks before but I just followed what all the others were doing. 

Needed due to the fact that the /obj/machinery/shield_gen/external puts shields on tiles that are /turf/simulated/floor/outdoors even if it's a /turf/simulated/floor/outdoors/turfpack/station tile (as it's a /turf/simulated/floor/outdoors
## Changelog
Nothing player facing


Kazzc:
Sweet, really appreciate this~ <3

PR #10728, opened 2025-04-22
Adds Pilot access to Explorer

Aphast:
## About The Pull Request
Adds Pilot access to the standard explorer role. This is by request of Asmuth per the #suggestions channel, though my own bias is obviously here being the one to add the access. (Medic has not been touched, but 'edit by maintainers' is on for such a reason, or just yap at me to edit it)

This appears to build fine, and private testing shows the access works, even if I do not actually know how to pilot all that well.
I am told 'internal reception' is well, but This is supposed to be for an already set or mostly received well idea per the staff topic (I would link if I could), and [this thread.](https://discord.com/channels/180538463655821312/1301373266384125963)
image
## Changelog
:cl:
add: Pilot access added the standard explorer role
/:cl:

### DIFF
--- a/code/game/jobs/job/exploration.dm
+++ b/code/game/jobs/job/exploration.dm
@@ -103,8 +103,8 @@
 	selection_color = "#999440"
 	economic_modifier = 6
 	pto_type = PTO_EXPLORATION
-	access = list(access_explorer, access_external_airlocks, access_eva)
-	minimal_access = list(access_explorer, access_external_airlocks, access_eva)
+	access = list(access_explorer, access_external_airlocks, access_eva, access_pilot)
+	minimal_access = list(access_explorer, access_external_airlocks, access_eva, access_pilot)
 	outfit_type = /decl/hierarchy/outfit/job/explorer2
 	job_description = "An " + JOB_EXPLORER + " searches for interesting things, and returns them to the station."
 	alt_titles = list(JOB_ALT_SURVEYOR = /datum/alt_title/surveyor, JOB_ALT_OFFSITE_SCOUT = /datum/alt_title/offsite_scout)
--- a/code/game/jobs/job/exploration.dm
+++ b/code/game/jobs/job/exploration.dm
@@ -103,8 +103,8 @@
 	selection_color = "#999440"
 	economic_modifier = 6
 	pto_type = PTO_EXPLORATION
-	access = list(access_explorer, access_external_airlocks, access_eva, access_pilot)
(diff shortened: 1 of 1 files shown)


code/game/jobs/job/exploration.dm
```diff
@@ -103,8 +103,8 @@
 	selection_color = "#999440"
 	economic_modifier = 6
 	pto_type = PTO_EXPLORATION
-	access = list(access_explorer, access_external_airlocks, access_eva)
-	minimal_access = list(access_explorer, access_external_airlocks, access_eva)
+	access = list(access_explorer, access_external_airlocks, access_eva, access_pilot)
```
Kashargul, line 106: Edit of upstream files without comments

Kashargul (changes requested):

code/game/jobs/job/exploration.dm
Aphast: Should be sufficient, lest I misread soemthing

SpadesNeil:
Then why even have a pilot role?

Pilot should just be an alt of explorer role at this point.

Aphast:
> Then why even have a pilot role?
> 
> Pilot should just be an alt of explorer role at this point.

Some have suggested that per the thread, but I do not have the full picture myself, this is simply legwork done so the decision can be made. The typical statement is that 'pilot is just better explorer'

SpadesNeil:
> Some have suggested that per the thread, but I do not have the full picture myself, this is simply legwork done so the decision can be made. The typical statement is that 'pilot is just better explorer'

Well either we have to leave it as it was or we remove pilot as a role and make it an alt-role of explorer, because making pilot a duplicate of explorer or leaving it as "explorer but better" is rather dumb.

Not saying _you're_ the one who's dumb. Clearly that dumb choice was made long before either of us showed up. You're just trying to fix something that is, in fact, clearly very dumb.

Aphast:
> > Some have suggested that per the thread, but I do not have the full picture myself, this is simply legwork done so the decision can be made. The typical statement is that 'pilot is just better explorer'
> 
> Well either we have to leave it as it was or we remove pilot as a role and make it an alt-role of explorer, because making pilot a duplicate of explorer or leaving it as "explorer but better" is rather dumb.
> 
> Not saying _you're_ the one who's dumb. Clearly that dumb choice was made long before either of us showed up. You're just trying to fix something that is, in fact, clearly very dumb.

Many past decisions with explo are quite controversial that much is true. No offense taken either, I am just throwing this up due to it being requested, even if I do have some bias towards allowing any solution that lets piloting ships become more accessible.

PR #10732, opened 2025-04-24
Removing all hard access locks.

tacoguy7765093:
## About The Pull Request
Putting hard access locks that are unable to be bypassed except via access was a relic of a old past that 'babyproofed' a little bit too hard in case of job stealing, something that should be handled by moderation. And I feel like making the game 'anti player' for the sake of 'people following the rules' isn't good practice anymore. 

Plus, I think it would be great to have more conflict brewing if someone decides to step out of line.

Plus, if you managed to go as far as breaking into the department, hacking multiple doors after ordering things via cargo
Any computer console that had hard access (excluding the command console's login) had their hard access lock lifted.
closes #10728

## Changelog
:cl:
qol: Supply console, Robotics Fabs, and Pilot consoles no longer have hard access locks.
/:cl:

### DIFF
--- a/code/game/machinery/computer/supply.dm
+++ b/code/game/machinery/computer/supply.dm
@@ -23,7 +23,7 @@
 	icon_keyboard = "tech_key"
 	icon_screen = "supply"
 	light_color = "#b88b2e"
-	req_access = list(access_cargo)
+	//req_access = list(access_cargo) //CHOMPedit, removing hard access locks.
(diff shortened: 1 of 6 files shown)


SpadesNeil:
If I am reading this correctly, basically, if you're in the area, you're able to access the machine.

The access control is controlling access to the area. Or hacking in. But once you're in, you can use the things, because computers can't be hacked to allow access.

Is that correct?

PR #10733, opened 2025-04-24
Firefighter Alt title 

crabcube:
## About The Pull Request

Does what it says on the tin

We're fighting the fires now.

## Changelog

:cl:

add: Added more things

/:cl:


Kashargul (changes requested):
Move upstream, DMI files from upstream are NOT to be edited in any way.

added with #10782

PR #10879, opened 2025-05-15
[MIRROR] Hallucination Refactor

CHOMPStation2StaffMirrorBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/17661
--------------------
## About The Pull Request
Hallucination code is very old, and full of old memory leaks due to how much less was understood during its creation. As well, most of the code uses sleep and spawn extensively. This is a complete rework of the code to bring it up to modern standards while retaining most of the original features.

## Changelog
A new standardized object for sending images to specific clients safely has been added /image/client_only, motion tracker echos have been updated to use this, (they were a prototype for it.) Using << to send images to clients is unsafe and results in hard-dels. 

Xenochimera hallucination code was entirely copypasted and extended. It has been rolled into hallucination code, and given a subtype. It is still handled slightly differently, but uses the new component system as well. 

All events the hallucination component can cause have been moved to procs for ease of reading and future extension of the system. New effects can be added trivially with a new proc, and an extension to the switch that activates them. This also isolates event logic from the timer logic that controls the hallucination component itself. No more confusion while adding new stuff!

Hallucination attackers have been entirely rebuilt from the ground up and spawn rarely again. Recoded to allow for future additions to be done much easier. Including an example of a fleeing hallucination that is not spawned by standard hallucination code. Some minor features have been lost, such as the attacker hitting you with specific weapons, but these we not spawned at all before, and teleported to move, making them feel very unfinished and old.

:cl:
refactor: complete rebuild of hallucination code to use components and more safely handled client images
/:cl:


Cameron-The-Raven:
Decon'd in #10891

PR #10942, opened 2025-05-22
Golden-Egg-Ref

Kazzc:
Adds a new ship to the reference map, "1SP_Pods". The Golden Egg from Golden Goose Casino!


Kazzc:
Sure! I'll make a folder for the ref maps!

PR #11073, opened 2025-06-14
makes the switch tail layer verb let you choose from three layers

timothyteakettle:
## About The Pull Request
gives the choice of upper/standard/lower layers
upper is the old alt layer
standard is the default layer
lower is a new layer so that your tail goes UNDER clothing (lamias rejoice, you can wear clothing without looking stupid)

i also fixed changing direction calling update_tail_showing twice because legacy code is dumb and stupid

## Changelog

:cl:
add: makes the switch tail layer verb let you choose from three layers
/:cl:


timothyteakettle:
it feels like you didn't even bother looking at the changes

is your role just commenting on non-mirror PRs "do this upstream" - it feels like this is counterproductive to trying to get contributions for the codebase

timothyteakettle:
removed my prior comment (it was not at all productive!)

i'll make this upstream

PR #11400, opened 2025-08-12
Absorbed Prey Names

Fracshun:
## About The Pull Request
Adds a belly-specific toggle and text field for pred's to modify absorbed prey's visible name in Me, Subtle, Say, Whisper, Pme and Psay commands. This toggle is disabled by default, and the default name for absorbed prey is "%pred's %belly". Once enabled, it will show this name in place of the absorbed prey's name when they send messages. It does not rename their character. This means it will not update for the manifest, transcore messages, admin logs, or things like LOOC. This should enable those who enjoy name changes to experience it without needing to rename their whole slot.
## Changelog
:cl:
add: Added belly option to change the visual name on absorbed prey's messages.
/:cl:

### DIFF
--- a/code/modules/emotes/custom_emote.dm
+++ b/code/modules/emotes/custom_emote.dm
@@ -64,7 +64,17 @@
 
 /mob/proc/build_the_emote(m_type, message, input, range, runemessage)
 	if(client)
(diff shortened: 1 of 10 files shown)


Kashargul (changes requested):
Tgui changes. Upstream

PR #11430, opened 2025-08-15
[MIRROR] Explosion subsystem

CHOMPStation2StaffMirrorBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/17499
--------------------
## About The Pull Request
This change moves explosions into a subsystem so that they may be quickly processed over several ticks, reducing lag, and preparing for a future MC that is not tolerant of full lockups. Has been tested for multiple sessions on outpost21 without any issues, we have multiple large explosions regularly through the round due to liberal use of landmines (coded to actually explode as well.)

Explosions now behave as the old "recursive explosions" except now without the recursion and possibility of exploding the MC with a stack overflow. Over a certain size the "recursive" logic stops checking explosion resistance at the center of the explosion, and only toward the edges, as a way to boost the speed of explosions beyond max cap. Something only possible with unique explosions or admemes. There should be minimal player facing changes for servers with phoronics/toxins tank transfer bombs.

closes #11423

## Changelog
Explosions were originally giant MC blocking loops that were not chunked in any way. Resulting in explosions needing to fully resolve before any other processing could happen. While this does not matter much on MOST explosions. Anything larger than max-cap could easily lock the server up for up to 10 minutes.

:cl:Will
refactor: Explosions moved to a subsystem to substantially reduce lag of massive explosions.
/:cl:


Willburd:
God speed, time for my first MC crash

Kashargul:
> God speed, time for my first MC crash

https://www.youtube.com/watch?v=NZN_Hk4fbS8

PR #11515, opened 2025-08-27
[MIRROR] Atmospheric Device Changes, Added Volume Pump

CHOMPStation2StaffMirrorBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/18261
--------------------
## About The Pull Request

This PR introduces the ability to link devices to the atmos controllers. This is for the General Air Control and Large Tank Control. Alongside the ability to build Gas Sensors. And even added the a new gas pump, "Volumetric Pump."

The Atmos Controller changes enables engineers to rebuild or create their own atmos tanks. And even in some cases give better control to certain custom engine setups. While it is fairly limited, it is a good start to giving the engineers more tools to work with.

This all can be done by clicking on the Atmos Air Control computer, the Gas Injector, the Unary Vent, Sensor, and Meters with a Multi-tool. Configure their Radio Frequencies and Tags. And then use the multitool to link them to one another.

7s55lii

Next up is the Volumetric Pump.

The Volumetric Pump is a new and fairly powerful pump to Chomp Station, though others from other servers will be familiar with it. Instead of moving gas to set pressures. This pump tries to move gas as quickly as possible, maintaining the requested flow-rate till the Pressure limit (9Mpa) is reached on the other side.

Because the pump is naievely moving gases, it doesn't try to reach accurate pressures on the output. Meaning it can easily overpressure. HOWEVER. The pump can be overclocked to bypass this limit. But it comes with some risks. Bipassing the pressure  limits means that gases will start leaking into the environment.

https://github.com/user-attachments/assets/4227f3c5-327c-426e-859a-abd576a40ead

## Changelog
:cl: Jordhin Xenshire
add: Volumetric Pump: This fast pump cares only about moving gasses quickly from one side to the other without slowing - or care for accuracy. It has a pressure limit lower than a standard pump, however this can be bipassed with a Multitool. What consequences could there be?
add: The ability to build Gas Sensors from the Rapid Pipe Dispenser.
add: The ability to link Gas Sensors, Meters, Injectors, Unary Vents to an Atmos Air Controller. This enables you to build custom tanks. Functionality is somewhat limited for now. This can be done with a multitool.
add: Ability to print Tank Control computer circuit board.
/:cl:


moonshineTheleocat:
Errors are caused by Merge meta data bleeding into the code base.

Is there something I need to do to prevent this from happening? As this is the second time

PR #11577, opened 2025-09-07
Obliterates the SM Shard from Sif

Aroliacue:
## About The Pull Request
Obliterates the SM Shards from the two Deathclaw Temple POIs.

:cl:
maptweak: Removes the exposed SM Shard that can occasionally spawn within a POI on Sif.
/:cl:


Cyrelius:
The whole reason that POI exists is a former cult of supermatter.  It's probably better to just remove the whole POI.
The other option is to block off the supermatter with some tough glass and make its turf unsimulated, so it won't cause any problem unless it's broken open.

PR #11927, opened 2025-11-07
[DO NOT MERGE] Cetus MS

Kashargul:
Just for TM


lbnesquik:
image
Floating light

Kashargul:
> image
> 
> Floating light

Should be opened as issue or reported on discord in the map thread. This is just a map swap TM which likely isn't monitored by mappers.

PR #12134, opened 2025-12-19
[MIRROR] Unstable metamorph ray can be fixed with multitool

CHOMPStation2StaffMirrorBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/18830
--------------------
## About The Pull Request
Allows the unstable metamorphosis ray to be repaired into an advanced one via multitool, so that people who want it for scenes don't have to gamble as much if they'd like. An advanced metamorphosis ray can now also be 'scrambled' into an unstable one if people want too.
## Changelog
:cl:
qol: Unstable metamorphosis ray can be repaired into an advanced one with a multitool.
/:cl:


Cameron-The-Raven:
I can't see any way to obtain these that doesn't already let you obtain the non-random variant.

PR #12199, opened 2026-01-01
Some new Vac-Pack stuff

Verkister:
## About The Pull Request
-Adds a new subtype "Scoop Hopper" that basically does the same as regular vac but more classic trash bag style with different sounds and verbs and no animation wait. (no custom sprites yet tho)
-Turf targeted vacuuming with loose items now has a new visual effect. (doesn't show up on floor dirt or individually targeted items)
-Suck sound and verb and animation toggle now have editable vars.
-New dangerous power level mode that pulls loose targets in from 3x3 radius.
## Changelog
:cl:
add: Added a new vac-pack subtype and a new power level and effects.
qol: Vac-pack custom sounds and verbs and anim toggle are now editable vars.
/:cl:


Verkister:
Anything else I should add?

PR #12440, opened 2026-02-22
Prevent Prometheans and Proteans from taking Rotting Genetics

Sharkmare:
## About The Pull Request
Prevents Proteans and Prometheans from taking the rotting  Genetic trait as they dont have Genetics and are unaffected by the downsides.
## Changelog
:cl:
fix:Rotting genetics can no longer be taken by Prometheans and Proteans in Character setup
/:cl:


Cameron-The-Raven:
Upported.

PR #12622, opened 2026-03-31
adds a shapeshifting trait

zeskorion:
## About The Pull Request
adds a true shapeshifter trait. This allows for freely editing character appearance on the fly, and costs three points.
The three points is a compromise- It's less restrictive than the one-point cocoon, and I don't want every single character around to just suddenly be a shapeshifter. I'm very willing to change the point cost, make the trait neutral, whatever, if desired.

Made modular, I do not expect this to be desirable to virgo

## Changelog
:cl:
add: shapeshifter trait
/:cl:


BlackMajor:
>It's less restrictive than the one-point cocoon

You fool, Weaver exists in neutral traits that does what cocoon does FOR FREE.
(Make it Neutral fr since everyone already COULD neutral shapeshifter via Weaver)

PR #12804, opened 2026-06-05
Object Vore Liberation

Verkister:
## About The Pull Request
Something fun a lot of people had been wishing for, including myself.
Adds the trash eater restriction toggle verb to the vore verbs on living mob login and an admin log alert for sussing out possible abuse when the verb is used.
## Changelog
:cl:
add: Added trash eater toggle to the vore verbs on living mob login.
add: Added an admin log alert for the trash eater toggle verb for sussing out possible abuse.
/:cl:

### DIFF
--- a/code/modules/mob/living/login.dm
+++ b/code/modules/mob/living/login.dm
@@ -23,6 +23,7 @@
 	add_verb(src, /mob/living/proc/mute_entry)
 	add_verb(src, /mob/living/proc/liquidbelly_visuals)
 	add_verb(src, /mob/living/proc/fix_vore_effects)
+	add_verb(src, /mob/living/proc/restrict_trasheater) //CHOMPAdd
 
 	if(!no_vore)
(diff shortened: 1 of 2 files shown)


zeskorion:
I don't think that allowing this without oversight is the right play

Cameron-The-Raven:
Personally I'm iffy on this PR.
The logging is fine, but I think it should still require admin intervention to toggle it on instead of having just EVERYTHING be able to be eaten with a single toggle.

At that point, if we can just toggle the 'safety to eating items' off, why even _have_ the safety in the first place?

If the issuee is that items aren't trasheatable, it'd seem like the better solution would to enable trash eater on those items.

Verkister:
Will need unfucking of the whole system. It appears someone took the unwisely named whitelist bypass var a little too literally during some refactors and removed *all* restrictions at some point.

PR #12836, opened 2026-06-23
Borgs can eat Borg-o's!

WolvesAndOwls:
## About The Pull Request
Borgs can now, as before, eat borg-o's.
## Changelog
:cl:
fix: Borgs can now eat borg-o's
/:cl:

### DIFF
--- a/code/game/objects/items/trash_vr.dm
+++ b/code/game/objects/items/trash_vr.dm
@@ -25,6 +25,27 @@
 	desc = "Contains every type of scrap material your robot puppy needs to grow big and strong."
 	icon = 'icons/mob/dogborg_vr.dmi'
 	icon_state = "kibble"
+	//CHOMPAdd
+	var/bitecount = 0
+	var/bites = 5
+	//CHOMPEnd
+
+//CHOMPAdd
+//Nibble is for borgs poking at borg-o's.
+/obj/item/trash/rkibble/attack_ai(mob/user as mob)
+	. = ..()
+	if(isrobot(user))
+		bitecount++
+		user.setClickCooldown(user.get_attack_speed(user))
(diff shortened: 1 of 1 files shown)


code/game/objects/items/trash_vr.dm
```diff
@@ -25,6 +25,26 @@
...
+
+//CHOMPAdd
+/obj/item/trash/rkibble/attack_ai(mob/user as mob)
+	. = ..()
+	if(isrobot(user))
+		bitecount++
+		user.setClickCooldown(user.get_attack_speed(user))
+		//TODO I want to give borgs charge here but I don't know how to do that yet.
+		if(bitecount>=5)
+			user.visible_message(span_infoplain(span_bold("[user]") + " finishes eating \the [src]."),span_info("You finish eating \the [src]."))
```
Kashargul, line 41: why are you mixing visible message and balloon alert? Either or.

Kashargul (changes requested):

code/game/objects/items/trash_vr.dm
WolvesAndOwls: This is a copy paste from everything else. I'll change it.

WolvesAndOwls: Actually, why not both? Wouldn't it be more responsive visually and flavorly? You can see the action then check the log if you missed it.

Kashargul: visible message implies the other 

Why should it take the performance twice

WolvesAndOwls: I see, I didn't know you can pass a runechat message through visible_message. I'm also pushing this upstream to virgo

PR #12904, opened 2026-07-13
Gets rid of unused interface window

ItsSelis:
## About The Pull Request

Most references to this window/element have been removed in PR #10930. These seem to have been some missed leftovers as the system was upgraded.

It is most noticeable on Linux playing the game through Wine, as there is a black/empty and non-closeable "Marking Selection Editing" window that spawns with the game itself.

## Changelog
:cl:
fix: Removes old/unused interface window/element
/:cl:


ItsSelis:
Will be fixed in https://github.com/VOREStation/VOREStation/pull/19616

PR #12987, opened 2026-08-30
Adds Food Synthesizers to all maps

Poojawa:
Also adds direction Open signs for the bar, because it looked better with how cluttered the whole ass map is.

As a reminder, the machines automatically offset on round start.

Practical change log
Cetus: Moved light from next to the door to on top of the jukebox

Southern Cross: Deleted extra camera next to bar door, external camera can see through windows and airlock to fulfill vision needs. Made the bar open sign directional and shifted it away from where the food synthesizer is now. 

Relic/Soluna: No real changes of note, Relic had a single chair removed near the stairwell. 

StrongDMM cleaned up map vars automatically

:cl: Poojawa
maptweak: Added Food synthesizers to Southern Cross, Cetus, and for consistency Relic/Soluna
/:cl:


zeskorion:
I'd prefer these changes not be done on cetus. I'm actively maintaining the map, and have an apparent fix pr and would prefer avoid merge conflicts

Additionally, i do not believe this is a good change.

Poojawa:
Map conflicts shouldn't happen because they aren't in the same area as some of the changed areas, unless you're fully overhauling the bar...?

I also made efforts to be as low impact as possible to the overall layout, it really is just that moved light to put the machine in.

Kashargul:
Staff discussion prefers it to remain buildable and not having it mapped in.

PR #13025, opened 2026-09-11
[MIRROR] Improvifys spring traps

CHOMPStation2StaffMirrorBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/19712
--------------------
spring pad code
## About The Pull Request
Moves spring traps from the protolathe to hand-crafting, reduces a lot of their tool action times, and also makes them affected by tool efficency.

Additionally makes normal spring traps require manual resetting, while self-resetting ones require an additional motor to craft.

## Changelog
:cl:
add: added a self-resetting spring trap. The normal one nolonger self-resets as a side effect.
add: Made springtraps craftable via gui hand-crafting menu
del: Removed spring traps from the protolathe. They're improvised now.
qol: Added a mini-grace period to disarming spring traps without them going off. This should help with misclicks.
/:cl:


Cameron-The-Raven:
DBD_KEYCHUP_FNAF_CERT_PR_SCREENSHOT_02_VF_91c0d2703b

PR #13045, opened 2026-09-19
Adds an invisible shoe state for mech shoes

Poojawa:
Because digitigrade shoes are uh. looking for an override icon that doesn't exist. Also ensures they'll be looking for proper icons for digitigrade overrides now. This wasn't an issue because these shoes don't exist on Virgo.

:cl: Poojawa
fix: mech shoes given proper invisible sprites so they don't throw missing icon errors
/:cl:


Kashargul:
This breaks mirror bot. DMIs are only compared by hash.

They need to be 1:1 with virgo. Or the sprites need to he in modular.

Poojawa:
Mech shoes don't exist on virgo, though...? Why just instantly close the Pr when the dry golashes have to specifically be in mob_digi to show up??

Kashargul:
> Mech shoes don't exist on virgo, though...? Why just instantly close the Pr when the dry golashes have to specifically be in mob_digi to show up??

Because you didn't read the https://github.com/CHOMPStation2/CHOMPStation2/blob/master/.github/CONTRIBUTING.md and just updated upstream dmis here. Which as stated above would break mirror bot.

